PR #38, opened 2025-01-02
Build blocks asynchronously

yacovm:
When it is a node's turn to propose a block, it calls the BlockBuilder's BuildBlock method.
However, this method may take a long time to complete, and during this time, the Epoch instance
cannot process messages (since proposing a new block may be called indirectly via HandleMessage).
    
This commit adds an asynchronous task scheduler which the Epoch uses to register a callback to make itself propose the block after building it asynchronously.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -425,21 +425,18 @@ func (e *Epoch) handleFinalizationMessage(message *Finalization, from NodeID) er
 	// If we have not received the proposal yet, we won't have a Round object in e.rounds,
 	// yet we may receive the corresponding finalization.
 	// This may happen if we're asynchronously verifying the proposal at the moment.
-	var pendingProposal bool
 	if _, exists := e.rounds[finalization.Round]; !exists && e.round == finalization.Round {
-		pendingProposal = true
+		e.Logger.Debug("Received a finalization for the current round",
+			zap.Uint64("round", finalization.Round), zap.Stringer("from", from))
+		e.storeFutureFinalization(message, from, finalization.Round)
+		return nil
 	}
 
 	// This finalization may correspond to a proposal from a future round, or to the proposal of the current round
 	// which we are still verifying.
-	if (e.round < finalization.Round && finalization.Round-e.round < e.maxRoundWindow) || pendingProposal {
+	if e.round < finalization.Round && finalization.Round-e.round < e.maxRoundWindow {
 		e.Logger.Debug("Got vote from round too far in the future", zap.Uint64("round", finalization.Round), zap.Uint64("my round", e.round))
-		msgsForRound, exists := e.futureMessages[string(from)][finalization.Round]
-		if !exists {
-			msgsForRound = &messagesForRound{}
-			e.futureMessages[string(from)][finalization.Round] = msgsForRound
-		}
-		msgsForRound.finalization = message
+		e.storeFutureFinalization(message, from, finalization.Round)
 		return nil
 	}
 
@@ -462,6 +459,15 @@ func (e *Epoch) handleFinalizationMessage(message *Finalization, from NodeID) er
 	round.finalizations[string(from)] = message
 
 	return e.maybeCollectFinalizationCertificate(round)
+}
+
+func (e *Epoch) storeFutureFinalization(message *Finalization, from NodeID, round uint64) {
+	msgsForRound, exists := e.futureMessages[string(from)][round]
+	if !exists {
+		msgsForRound = &messagesForRound{}
+		e.futureMessages[string(from)][round] = msgsForRound
+	}
+	msgsForRound.finalization = message
 }
 
 func (e *Epoch) handleVoteMessage(message *Vote, from NodeID) error {
--- a/epoch.go
+++ b/epoch.go
@@ -6,7 +6,6 @@ package simplex
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -876,7 +875,7 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
 	}
 
 	// Create a task that will verify the block in the future, after its predecessors have also been verified.
-	task := e.createBlockVerificationTask(block, md, from, vote)
+	task := e.createBlockVerificationTask(block, from, vote)
 
 	// isBlockReadyToBeScheduled checks if the block is known to us either from some previous round,
 	// or from storage. If so, then we have verified it in the past, since only verified blocks are saved in memory.
@@ -885,16 +884,17 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
 	// Schedule the block to be verified once its direct predecessor have been verified,
 	// or if it can be verified immediately.
 	e.Logger.Debug("Scheduling block verification", zap.Uint64("round", md.Round))
-	e.sched.Schedule(md.Digest, task, md.Prev, canBeImmediatelyVerified)
+	e.sched.Schedule(task, md.Prev, canBeImmediatelyVerified)
 
 	return nil
 }
 
-func (e *Epoch) createBlockVerificationTask(block Block, md BlockHeader, from NodeID, vote Vote) func() {
-	return func() {
+func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote) func() Digest {
+	return func() Digest {
+		md := block.BlockHeader()
 		if err := block.Verify(); err != nil {
 			e.Logger.Debug("Failed verifying block", zap.Error(err))
-			return
+			return md.Digest
 		}
 
 		e.lock.Lock()
@@ -905,7 +905,7 @@ func (e *Epoch) createBlockVerificationTask(block Block, md BlockHeader, from No
 
 		if !e.storeProposal(block) {
 			e.Logger.Warn("Unable to store proposed block for the round", zap.Stringer("NodeID", from), zap.Uint64("round", md.Round))
-			return
+			return md.Digest
 			// TODO: timeout
 		}
 
@@ -915,13 +915,15 @@ func (e *Epoch) createBlockVerificationTask(block Block, md BlockHeader, from No
 		if !exists {
 			// This shouldn't happen, but in case it does, return an error
 			e.Logger.Error("programming error: round not found", zap.Uint64("round", md.Round))
-			return
+			return md.Digest
 		}
 		round.votes[string(vote.Signature.Signer)] = &vote
 
 		if err := e.doProposed(block, vote, from); err != nil {
 			e.Logger.Warn("Failed voting on block", zap.Error(err))
 		}
+
+		return md.Digest
 	}
 }
 
@@ -1051,30 +1053,25 @@ func (e *Epoch) buildBlock() {
 
 	task := e.createBlockBuildingTask(metadata)
 
-	// We set the task ID to be the hash of the previous block, since we don't know the digest
-	// of the block before it is built.
-	// We know, however, that any block before or after the previous block won't
-	// have this digest, under the assumption that the hash function is collision resistant.
-	// TODO: Inject this hash as a dependency and don't use SHA256 as a hardcoded function
-	taskID := sha256.Sum256(metadata.Prev[:])
-
 	e.Logger.Debug("Scheduling block building", zap.Uint64("round", metadata.Round))
 	canBeImmediatelyVerified := e.isBlockReadyToBeScheduled(metadata.Seq, metadata.Prev)
-	e.sched.Schedule(taskID, task, metadata.Prev, canBeImmediatelyVerified)
+	e.sched.Schedule(task, metadata.Prev, canBeImmediatelyVerified)
 }
 
-func (e *Epoch) createBlockBuildingTask(metadata ProtocolMetadata) func() {
-	return func() {
+func (e *Epoch) createBlockBuildingTask(metadata ProtocolMetadata) func() Digest {
+	return func() Digest {
 		block, ok := e.BlockBuilder.BuildBlock(e.finishCtx, metadata)
 		if !ok {
 			e.Logger.Warn("Failed building block")
-			return
+			return Digest{}
 		}
 
 		e.lock.Lock()
 		defer e.lock.Unlock()
 
 		e.proposeBlock(block)
+
+		return block.BlockHeader().Digest
 	}
 }
 
--- a/epoch.go
+++ b/epoch.go
@@ -892,13 +892,14 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
 
 func (e *Epoch) createBlockVerificationTask(block Block, md BlockHeader, from NodeID, vote Vote) func() {
 	return func() {
-		e.lock.Lock()
-		defer e.lock.Unlock()
-
 		if err := block.Verify(); err != nil {
 			e.Logger.Debug("Failed verifying block", zap.Error(err))
 			return
 		}
+
+		e.lock.Lock()
+		defer e.lock.Unlock()
+
 		record := BlockRecord(md, block.Bytes())
 		e.WAL.Append(record)
 
--- a/epoch.go
+++ b/epoch.go
@@ -793,6 +793,14 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
 		return nil
 	}
 
+	// The block is for a too high round, we shouldn't handle it as
+	// we have only so much memory.
+	if md.Round-e.round >= e.maxRoundWindow {
+		e.Logger.Debug("Received a block message for a too high round",
+			zap.Uint64("round", md.Round), zap.Uint64("our round", e.round))
+		return nil
+	}
+
 	// Ignore block messages sent by us
 	if e.ID.Equals(from) {
 		e.Logger.Debug("Got a BlockMessage from ourselves or created by us")
--- a/epoch.go
+++ b/epoch.go
@@ -919,7 +919,7 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
 		}
 		round.votes[string(vote.Signature.Signer)] = &vote
 
-		if err := e.doProposed(block, vote, from); err != nil {
+		if err := e.doProposed(block, vote); err != nil {
 			e.Logger.Warn("Failed voting on block", zap.Error(err))
 		}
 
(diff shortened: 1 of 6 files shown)


sched.go
```diff
@@ -0,0 +1,143 @@
...
+func (as *scheduler) run() {
+	as.lock.Lock()
+	defer as.lock.Unlock()
+
+	for !as.close {
+		as.maybeExecuteTask()
+		if as.close {
+			return
+		}
+		as.signal.Wait()
```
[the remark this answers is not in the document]
yacovm: When we execute a task that is ready to run, it will add itself to the ready queue. 

Then, it will perform another iteration of the loop in `maybeExecuteTask` which will not exit, as the ready queue is still not empty, and will execute the next task.

epoch.go
```diff
@@ -762,31 +845,70 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
 		return nil
 	}
 
-	if !e.storeProposal(block) {
-		e.Logger.Warn("Unable to store proposed block for the round", zap.Stringer("NodeID", from), zap.Uint64("round", md.Round))
-		// TODO: timeout
-	}
+	// Create a task that will verify the block in the future, after its predecessors have also been verified.
+	task := e.scheduleBlockVerification(block, md, from, vote)
```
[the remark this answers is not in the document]
yacovm: Since we couple votes from block proposers with the blocks they propose, we don't want some nefarious actor to take a vote from the real proposer and send us its own block + the vote from the real block proposer. 

We therefore check the vote actually votes on the block, and is not voting for an alternative block.

epoch.go
```diff
@@ -400,16 +413,40 @@ func (e *Epoch) validateFinalizationQC(fCert *FinalizationCertificate) (bool, er
...
+	// yet we may receive the corresponding finalization.
+	// This may happen if we're asynchronously verifying the proposal at the moment.
+	var pendingProposal bool
+	if _, exists := e.rounds[finalization.Round]; !exists && e.round == finalization.Round {
+		pendingProposal = true
+	}
+
+	// This finalization may correspond to a proposal from a future round, or to the proposal of the current round
+	// which we are still verifying.
+	if (e.round < finalization.Round && finalization.Round-e.round < e.maxRoundWindow) || pendingProposal {
```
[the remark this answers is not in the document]
yacovm: Address [here](https://github.com/ava-labs/Simplex/pull/38/commits/758812c0ad48beefd60a74c81cd298215a10e72e).

epoch.go
```diff
@@ -400,16 +413,40 @@ func (e *Epoch) validateFinalizationQC(fCert *FinalizationCertificate) (bool, er
...
+		if !exists {
+			msgsForRound = &messagesForRound{}
+			e.futureMessages[string(from)][finalization.Round] = msgsForRound
+		}
+		msgsForRound.finalization = message
+		return nil
+	}
+
 	// Have we already finalized this round?
 	round, exists := e.rounds[finalization.Round]
```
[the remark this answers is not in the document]
yacovm: Addressed [here](https://github.com/ava-labs/Simplex/pull/38/commits/e443792bbea6f7f397f3a6106591efff5ad8b3ac).

epoch.go
```diff
@@ -427,13 +464,46 @@ func (e *Epoch) handleFinalizationMessage(message *Finalization, from NodeID) er
...
+		msgsForRound, exists := e.futureMessages[string(from)][vote.Round]
+		if !exists {
+			msgsForRound = &messagesForRound{}
+			e.futureMessages[string(from)][vote.Round] = msgsForRound
+		}
+		msgsForRound.vote = message
+		return nil
+	}
+
 	// TODO: what if we've received a vote for a round we didn't instantiate yet?
```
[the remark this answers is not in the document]
yacovm: Addressed [here](https://github.com/ava-labs/Simplex/pull/38/commits/61ffd2e2998c58274938ca27b663f24d21a3fff8).

epoch.go
```diff
@@ -762,31 +845,70 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
-		return err
+	// Schedule the block to be verified once its direct predecessor have been verified,
+	// or if it can be verified immediately.
+	e.Logger.Debug("Scheduling block verification", zap.Uint64("round", md.Round))
+	e.sched.Schedule(md.Digest, task, md.Prev, canBeImmediatelyVerified)
+
+	return nil
+}
+
+func (e *Epoch) scheduleBlockVerification(block Block, md BlockHeader, from NodeID, vote Vote) func() {
```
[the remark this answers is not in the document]
yacovm: Addressed [here](https://github.com/ava-labs/Simplex/pull/38/commits/61ffd2e2998c58274938ca27b663f24d21a3fff8).

epoch.go
```diff
@@ -762,31 +845,70 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
+		round, exists := e.rounds[md.Round]
+		if !exists {
+			// This shouldn't happen, but in case it does, return an error
+			e.Logger.Error("programming error: round not found", zap.Uint64("round", md.Round))
+			return
+		}
+		round.votes[string(vote.Signature.Signer)] = &vote
+
+		if err := e.doProposed(block, vote, from); err != nil {
+			e.Logger.Debug("Failed voting on block", zap.Error(err))
```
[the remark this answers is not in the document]
yacovm: Addressed [here](https://github.com/ava-labs/Simplex/pull/38/commits/61ffd2e2998c58274938ca27b663f24d21a3fff8).

epoch.go
```diff
@@ -762,31 +845,70 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
+		// or there exists a round object for it.
+		// Since we only create a round object after we verify the block,
+		// it means we have verified this block in the past.
+		_, ok := e.locateBlock(seq-1, prev[:])
+		ready = ok
+	} else {
+		// The first block is always ready to be scheduled
+		ready = true
+	}
+	return ready
```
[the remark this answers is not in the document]
yacovm: Addressed [here](https://github.com/ava-labs/Simplex/pull/38/commits/61ffd2e2998c58274938ca27b663f24d21a3fff8), but keep in mind that your suggestion isn't semantically the same as the original code, because if seq == 0 then we should return true, not false.

yacovm:
Please don't review the latest two commits, I will extract them from this PR and try to simplify their code tomorrow.

epoch.go
```diff
@@ -980,7 +1163,7 @@ func (e *Epoch) startRound() error {
 	return e.handleBlockMessage(msgsForRound.proposal, leaderForCurrentRound)
 }
 
-func (e *Epoch) doProposed(block Block, voteFromLeader Vote) error {
+func (e *Epoch) doProposed(block Block, voteFromLeader Vote, from NodeID) error {
```
samliok, line 1166: nit: unused parameter from

epoch.go
```diff
@@ -762,31 +875,66 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
+
+		// Once we have stored the proposal, we have a Round object for the round.
+		// We store the vote to prevent verifying its signature again.
+		round, exists := e.rounds[md.Round]
+		if !exists {
+			// This shouldn't happen, but in case it does, return an error
+			e.Logger.Error("programming error: round not found", zap.Uint64("round", md.Round))
+			return
+		}
+		round.votes[string(vote.Signature.Signer)] = &vote
```
samliok, line 919: do we need to re-add this vote into the epochs memory when we sync a block record from the wal?

epoch_test.go
```diff
@@ -81,11 +82,169 @@ func TestEpochSimpleFlow(t *testing.T) {
 			injectTestFinalization(t, e, block, nodes[i], conf.Signer)
 		}
 
-		committedData := storage.data[i].Block.Bytes()
+		storage.waitForBlockCommit(uint64(i))
+
+		committedData := storage.data[uint64(i)].Block.Bytes()
```
samliok, line 87: should we have `waitForBlockCommit` to also return the `commitedData`?

epoch_test.go
```diff
@@ -81,11 +82,169 @@ func TestEpochSimpleFlow(t *testing.T) {
...
 		require.Equal(t, block.Bytes(), committedData)
 	}
 }
 
+func TestEpochBlockSentTwice(t *testing.T) {
+	l := makeLogger(t, 1)
+
+	var tooFarMsg, alreadyReceivedMsg bool
+
+	l.intercept(func(entry zapcore.Entry) error {
```
samliok, line 97: didn't know you could do this with logger, pretty cool

samliok (approved):

epoch.go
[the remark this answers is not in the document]
samliok: yep, meant to suggest we return true

epoch.go
yacovm: No, because if we have a block record in the WAL it's either one of the following cases: 

1. We have written it ourselves as a block proposer, so `createBlockVerificationTask` wasn't called.
2. We have written it ourselves when we received it from the block proposer. It's true that in `createBlockVerificationTask` we do just that, however - when we crash, we lose all vote messages and we never recover them when we recover. Therefore, it is perfectly fine to also lose this vote message from the leader. We add this vote just because the leader sends its vote as part of the block proposal, but in an alternative implementation it would've sent it separately and we'd lose it, and it's perfectly fine.

epoch_test.go
yacovm: we can do it in a separate PR, I added this `waitForBlockCommit` just because we now verify blocks asynchronously.

epoch.go
yacovm: fixed

epoch_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/54

sched.go
```diff
@@ -0,0 +1,186 @@
...
+		digest: id,
+	}
+
+	if ready {
+		as.ready = append(as.ready, task)
+	} else {
+		as.pending.Insert(task)
+	}
+
+	as.signal.Signal()
```
StephenButtolph, line 132: I think we only need to `Signal` if `ready == true` right?

epoch.go
```diff
@@ -980,7 +1163,7 @@ func (e *Epoch) startRound() error {
 	return e.handleBlockMessage(msgsForRound.proposal, leaderForCurrentRound)
 }
 
-func (e *Epoch) doProposed(block Block, voteFromLeader Vote) error {
+func (e *Epoch) doProposed(block Block, voteFromLeader Vote, _ NodeID) error {
```
StephenButtolph, line 1166: Why is this argument added if it is never used?

epoch.go
```diff
@@ -897,12 +1045,39 @@ func (e *Epoch) locateBlock(seq uint64, digest []byte) (Block, bool) {
 	return nil, false
 }
 
-func (e *Epoch) proposeBlock() error {
-	block, ok := e.BlockBuilder.BuildBlock(e.finishCtx, e.Metadata())
-	if !ok {
-		return errors.New("failed to build block")
+func (e *Epoch) buildBlock() {
```
StephenButtolph, line 1048: Is it ever possible for `canBeImmediatelyVerified` to be `false` here?

epoch.go
```diff
@@ -897,12 +1045,39 @@ func (e *Epoch) locateBlock(seq uint64, digest []byte) (Block, bool) {
...
+	metadata := e.metadata()
+
+	task := e.createBlockBuildingTask(metadata)
+
+	// We set the task ID to be the hash of the previous block, since we don't know the digest
+	// of the block before it is built.
+	// We know, however, that any block before or after the previous block won't
+	// have this digest, under the assumption that the hash function is collision resistant.
+	// TODO: Inject this hash as a dependency and don't use SHA256 as a hardcoded function
+	taskID := sha256.Sum256(metadata.Prev[:])
```
StephenButtolph, line 1058: I find the introduction of a "fake" `taskID` pretty unusual tbh. I think we could defer the expectation of a taskID until _after_ the task has been executed. Then we could correctly use the blockID after it was built as the taskID.

epoch.go
```diff
@@ -762,31 +875,66 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
+
+func (e *Epoch) createBlockVerificationTask(block Block, md BlockHeader, from NodeID, vote Vote) func() {
+	return func() {
+		e.lock.Lock()
+		defer e.lock.Unlock()
+
+		if err := block.Verify(); err != nil {
+			e.Logger.Debug("Failed verifying block", zap.Error(err))
+			return
+		}
```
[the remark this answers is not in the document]
yacovm: Ouch, I forgot to move the lock :-( 

Fixed [here](https://github.com/ava-labs/Simplex/pull/38/commits/fd77d0e1ebeaf6d22f099615f2a91824b4f8c82c).

epoch.go
```diff
@@ -762,31 +875,67 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
-		return err
+	// Schedule the block to be verified once its direct predecessor have been verified,
+	// or if it can be verified immediately.
+	e.Logger.Debug("Scheduling block verification", zap.Uint64("round", md.Round))
+	e.sched.Schedule(md.Digest, task, md.Prev, canBeImmediatelyVerified)
+
+	return nil
+}
+
+func (e *Epoch) createBlockVerificationTask(block Block, md BlockHeader, from NodeID, vote Vote) func() {
```
samliok, line 893: since blockheader is a field in `Block` can we remove passing in `md`

sched.go
```diff
@@ -0,0 +1,186 @@
...
+			return
+		}
+		if len(as.ready) > 0 {
+			continue
+		}
+		as.signal.Wait()
+	}
+}
+
+func (as *scheduler) maybeExecuteTask() {
```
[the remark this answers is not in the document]
yacovm: Correct, renamed [here](https://github.com/ava-labs/Simplex/pull/38/commits/6b94e1c576b694d583db9ee017614a9c2f37a08c).

sched.go
yacovm: Correct

epoch.go
yacovm: No, I probably should change it to true.

epoch.go
yacovm: Changed it [here](https://github.com/ava-labs/Simplex/pull/38/commits/6763e9787581dd72c49fa62a2e00c0f75fe4ba14).

sched.go
```diff
@@ -0,0 +1,186 @@
...
+func (t *dependencies[C, D]) Insert(v D) {
+	dependency := v.dependsOn()
+	t.dependsOn[dependency] = append(t.dependsOn[dependency], v)
+}
+
+func (t *dependencies[C, D]) Remove(id C) []D {
+	dependents := t.dependsOn[id]
+	delete(t.dependsOn, id)
+	return dependents
+}
```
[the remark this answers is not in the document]
yacovm: I [made](https://github.com/ava-labs/Simplex/pull/38/commits/6763e9787581dd72c49fa62a2e00c0f75fe4ba14) them non-generic. 

I did it generic for the test but after some thinking it's over-engineering from my side.

sched.go
```diff
@@ -0,0 +1,186 @@
...
+	return as.pending.Size()
+}
+
+func (as *scheduler) Close() {
+	as.lock.Lock()
+	defer as.lock.Unlock()
+
+	as.close = true
+
+	as.signal.Signal()
```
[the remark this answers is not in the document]
yacovm: I made it Signal because there is only a single goroutine consuming it. 

From an implementation perspective it does the same thing, as there can only be one goroutine parked on wait.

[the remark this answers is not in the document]
yacovm: I can change to Broadcast if you want, I'm not sure how it conveys anything different though, it's just related to how many goroutines can wake up from wait.

sched.go
```diff
@@ -0,0 +1,186 @@
...
+	go as.run()
+
+	return &as
+}
+
+func (as *scheduler) Size() int {
+	as.lock.Lock()
+	defer as.lock.Unlock()
+	return as.pending.Size()
+}
```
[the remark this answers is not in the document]
yacovm: Yes good point, I made it take into account the ready as well

sched.go
```diff
@@ -0,0 +1,186 @@
...
+		as.maybeExecuteTask()
+		if as.close {
+			return
+		}
+		if len(as.ready) > 0 {
+			continue
+		}
+		as.signal.Wait()
+	}
+}
```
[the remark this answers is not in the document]
yacovm: Yes I like this version much more, adopted it, thanks for the refactoring advice 👍

sched.go
```diff
@@ -0,0 +1,186 @@
...
+
+	as.signal.Signal()
+}
+
+func (as *scheduler) dispatchTaskAndScheduleDependingTasks(id Digest, task func()) func() {
+	return func() {
+		task()
+		as.lock.Lock()
+		newlyReadyTasks := as.pending.Remove(id)
+		as.ready = append(as.ready, newlyReadyTasks...)
```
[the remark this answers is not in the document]
yacovm: because initially the main loop of the scheduler was too big, so I refactored some of the code into this callback.

Your suggestion considerably shrunk the main loop so it makes sense to lose the callback and move it back to where it was.

epoch.go
yacovm: Removed it.

sched.go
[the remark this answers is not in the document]
yacovm: changed to Broadcast

epoch.go
yacovm: changed it to always be [true](https://github.com/ava-labs/Simplex/pull/38/commits/fc64e39e358b24873f65f10db14f39127f3475ce).

epoch.go
yacovm: changed it accordingly.

sched.go
```diff
@@ -0,0 +1,155 @@
...
+
+	for !as.close {
+
+		if len(as.ready) == 0 { // (1)
+			as.signal.Wait() // (2)
+			continue         // (3)
+		}
+
+		var task task
+		task, as.ready = as.ready[0], as.ready[1:] // (4)
```
StephenButtolph, line 93: This is probably not critical, but currently the tasks will only get GCed after the `ready` list is re-allocated during `append`. So typically this explicit zeroing out is recommended when using slices as a queue like this.
```suggestion
		task := as.ready[0]
		as.ready[0] = task{} // Allow the GC to clear the task after we are done executing it
		as.ready = as.ready[1:] // (4)
```

sched_test.go
```diff
@@ -0,0 +1,145 @@
...
+type testDependsOn []int
+
+func (m testDependsOn) dependsOn() int {
+	return m[1]
+}
+
+func (m testDependsOn) id() int {
+	return m[0]
+}
+
```
StephenButtolph, line 25:
```suggestion
```

StephenButtolph (approved):
2 small nits then lgtm

sched_test.go
yacovm: yeah I forgot to delete this...

sched.go
yacovm: Good idea, fixed!

PR #52, opened 2025-01-16
Out of order message

yacovm:


epoch.go
```diff
@@ -762,31 +949,85 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
+		record := BlockRecord(md, block.Bytes())
+		e.WAL.Append(record)
+
+		if !e.storeProposal(block) {
+			e.Logger.Warn("Unable to store proposed block for the round", zap.Stringer("NodeID", from), zap.Uint64("round", md.Round))
+			return
+			// TODO: timeout
+		}
+
+		e.deleteFutureProposal(from, md.Round)
```
samliok, line 1002: i think it would be more clear to call `deleteFutureX` directly after we perform the required action in `maybeLoadFutureMessages` .

epoch.go
```diff
@@ -762,31 +949,85 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, _ NodeID) error {
...
-	}
+	defer func() {
+		messagesFromNode, exists := e.futureMessages[string(from)]
+		if !exists {
+			return
+		}
+		if messagesForRound, exists := messagesFromNode[md.Round]; exists {
+			messagesForRound.proposal = nil
+		}
+	}()
```
samliok, line 960: can use the helper methods you already created for this

epoch.go
```diff
@@ -532,23 +660,38 @@ func (e *Epoch) persistFinalizationCertificate(fCert FinalizationCertificate) er
...
+			// Check if we can commit the next round
+			r++
+			round, exists := e.rounds[r]
+			if !exists {
+				break
+			}
+			if round.fCert == nil {
+				break
+			}
+			fCert = *round.fCert
```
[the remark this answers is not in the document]
yacovm: I don't think it does. If we are in this loop it means we have saved the finalization certificate to the WAL from earlier rounds, so we have broadcasted it then, earlier. 

If we do it here inside this for loop again, we'd be sending it the second time. 
While it doesn't hurt to do that, I don't think there is a point to broadcasting a message twice.

epoch.go
yacovm: done

epoch.go
```diff
@@ -532,23 +660,38 @@ func (e *Epoch) persistFinalizationCertificate(fCert FinalizationCertificate) er
...
-		e.Storage.Index(block, fCert)
-		e.Logger.Info("Committed block",
-			zap.Uint64("round", fCert.Finalization.Round),
-			zap.Uint64("sequence", fCert.Finalization.Seq),
-			zap.Stringer("digest", fCert.Finalization.BlockHeader.Digest))
-		e.lastBlock = block
+		for {
+			r := fCert.Finalization.Round
+			block := e.rounds[r].block
+			// TODO: also index future finalization certificates
```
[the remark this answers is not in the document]
yacovm: correct, removed

epoch.go
```diff
@@ -659,8 +713,14 @@ func (e *Epoch) maybeCollectNotarization() error {
 	votesForCurrentRound := e.rounds[e.round].votes
 	voteCount := len(votesForCurrentRound)
 
+	from := make([]NodeID, 0, voteCount)
+	for _, vote := range votesForCurrentRound {
+		from = append(from, vote.Signature.Signer)
+	}
+
```
samliok, line 720: we only use this array if `voteCount < e.quorumSize`. should we just move the creation of it to that condition so we don't need to generate it everytime

epoch_test.go
```diff
@@ -89,13 +89,135 @@ func TestEpochSimpleFlow(t *testing.T) {
...
+		Storage:             storage,
+		Comm:                noopComm(nodes),
+		BlockBuilder:        bb,
+		SignatureAggregator: &testSignatureAggregator{},
+	}
+
+	e, err := NewEpoch(conf)
+	require.NoError(t, err)
+
+	rounds := 10
```
samliok, line 130: nit: we could initialize this variable before `bb` to eliminate hardcoding the rounds twice

samliok (approved):
small nits, can merge if you'd like

epoch.go
yacovm: Addressed here: https://github.com/ava-labs/Simplex/pull/57

epoch_test.go
yacovm: Addressed here: https://github.com/ava-labs/Simplex/pull/57

PR #56, opened 2025-01-27
Notarization absence detection for leader failover

yacovm:


epoch.go
```diff
@@ -1220,6 +1231,65 @@ func (e *Epoch) metadata() ProtocolMetadata {
 	return md
 }
 
+func (e *Epoch) triggerProposalWaitTimeExpired(round uint64) {
+	leader := LeaderForRound(e.nodes, round)
+	e.Logger.Info("Timed out on block agreement", zap.Uint64("round", round), zap.Stringer("leader", leader))
+	// TODO: Actually start the empty block agreement
+}
+
+func (e *Epoch) monitorProgress(round uint64) {
```
samliok, line 1240: this function monitors the progress of just our block proposal stage right? I see that we also have a `cancelWaitForBlockNotarization` but we need set out a `waitUntil` call for the notarization phase. I'm assuming this is a todo for a future PR?

epoch_test.go
```diff
@@ -375,6 +377,155 @@ func TestEpochBlockTooHighRound(t *testing.T) {
 	})
 }
 
+func TestEpochLeaderFailover(t *testing.T) {
```
samliok, line 380: Would it make sense to separate these tests into their own file called `epoch_failover_test` or something similar? I think it would keep our helper functions from being lost, and organize the testing code a bit as I can imagine we will have a lot more failover tests.

epoch_test.go
```diff
@@ -375,6 +377,155 @@ func TestEpochBlockTooHighRound(t *testing.T) {
...
+
+	bb.blockShouldBeBuilt <- struct{}{}
+
+	waitForEvent(start, e, timeoutDetected)
+}
+
+func waitForEvent(start time.Time, e *Epoch, events chan struct{}) {
+	now := start
+	for {
+		now = now.Add(DefaultMaxProposalWaitTime / 5)
```
samliok, line 439: should we use the runtime variable `maxProposalWait` here and throughout the tests to avoid discrepancies when this deviates from the default? Or we could remove the runtime variable completely since we always set it to `DefaultMaxProposalWaitTime`.

monitor.go
```diff
@@ -0,0 +1,126 @@
...
+	deadline time.Time
+	f        func()
+}
+
+func NewMonitor(startTime time.Time, logger Logger) *Monitor {
+	m := &Monitor{
+		logger: logger,
+		close:  make(chan struct{}),
+		tasks:  make(chan func(), 1),
+		ticks:  make(chan time.Time, 1),
```
samliok, line 32: since tasks and ticks only have a capacity of 1, its possible that we could drop some `tasks/ticks` sent to the channel if they are sent directly after another.

I wrote this test which fails. Maybe i'm misunderstanding how the monitor should be used?
```go
func TestMonitorManyTick(t *testing.T) {
	start := time.Now()
	mon := NewMonitor(start, makeLogger(t))
	require.True(t, mon.shouldRun())
	for i := 0; i < 1000; i++ {
		mon.AdvanceTime(start.Add(time.Hour))
	}
	expectedTime := start.Add(1000 * time.Hour) 
	require.Equal(t, expectedTime, mon.time.Load().(time.Time))
}

```

epoch.go
yacovm: The `cancelWaitForBlockNotarization` is called when the block has been [notarized](https://github.com/ava-labs/Simplex/pull/56/commits/ecedb1f11a2a7254babd284b0b99b66d4100f491#diff-fa0c45dc9da1c22a9375093b2a1f051742d153a41647943a8dc0e6c49ca78740R779). 

we don't care if the block was proposed, but that it has been notarized.

monitor.go
yacovm: This is on purpose. Notice that the monitor goroutine does two things:

1. Waits for the VM to tell it "hey i think it's time a block should be proposed by someone!" - via the `WaitFor` call.
2. Waits for the timeout for a block to be notarized to expire, via the `WaitUntil` call.

During (1) we may get ticks, and we don't care, because when we do (1) we don't do (2) and we only need the ticks for (2). 

We are also guaranteed to only have up to a single `WaitFor` task and a up time a single `WaitUntil` task at the same time so we don't care about any other tasks not being invoked.

epoch_test.go
yacovm: done

epoch_test.go
yacovm: done

monitor.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	m.time.Store(startTime)
+
+	go m.run()
```
samliok, line 37: it might make sense to call this in a separate function to be a little more explicit

samliok (approved):

PR #58, opened 2025-01-27
Sched logging

yacovm:


yacovm:
Cannot pass CI until https://github.com/ava-labs/Simplex/pull/59 is merged

monitor.go
```diff
@@ -0,0 +1,123 @@
...
+	m.logger.Debug("Executing f", zap.Uint64("taskID", taskID), zap.Time("deadline", task.deadline))
+	task.f()
+	m.logger.Debug("Executed f with deadline", zap.Uint64("taskID", taskID))
+
+	// clean up future task to mark we have already executed it and to release memory
+	m.futureTask.Store(&futureTask{})
+}
+
+func (m *Monitor) run() {
+	var taskID uint64
```
samliok, line 74: we should increment the task id every time the task gets executed?

monitor.go
```diff
@@ -0,0 +1,123 @@
...
+func (m *Monitor) WaitUntil(timeout time.Duration, f func()) context.CancelFunc {
+	t := m.time.Load()
+	time := t.(time.Time)
+
+	m.futureTask.Store(&futureTask{
+		f:        f,
+		deadline: time.Add(timeout),
+	})
+
+	return nil
```
samliok, line 122: should we just set the function header to not return anything?

samliok (commented):
could you please add some background info for when we would use the `Monitor` and what sort of tasks we should expect the monitor to perform?

yacovm:
> could you please add some background info for when we would use the `Monitor` and what sort of tasks we should expect the monitor to perform?

This should be a question in the https://github.com/ava-labs/Simplex/pull/56 PR, here I am only adding some logs :)

monitor.go
yacovm: oops!

Fixed in the [original PR](https://github.com/ava-labs/Simplex/pull/56)

monitor.go
yacovm: Oops my bad, I have updated the [original PR](https://github.com/ava-labs/Simplex/pull/56) to address this, and also added a test case to test this

samliok:
ah ok! ill review that one soon

sched.go
```diff
@@ -85,20 +90,25 @@ func (as *scheduler) run() {
...
+			continue // (3)
 		}
 
 		taskToRun := as.ready[0]
 		as.ready[0] = task{}    // Cleanup any object references reachable from the closure of the task
 		as.ready = as.ready[1:] // (4)
 
-		as.lock.Unlock()    // (5)
+		as.lock.Unlock() // (5)
+		as.logger.Debug("Running task", zap.Int("remaining ready tasks", len(as.ready)))
```
StephenButtolph, line 104: This access of `as.ready` is racy.

yacovm: yeah it is, thanks for spotting this

yacovm: Fixed [here](https://github.com/ava-labs/Simplex/pull/58/commits/85a08af4f05dabb90bcf78bb988fcbc9c23812f4).

StephenButtolph (approved):

samliok (approved):

PR #59, opened 2025-01-28
Fix data race in test - logging interception

yacovm:


epoch_test.go
```diff
@@ -286,6 +286,15 @@ func TestEpochBlockSentTwice(t *testing.T) {
 func TestEpochBlockTooHighRound(t *testing.T) {
 	l := makeLogger(t, 1)
 
+	var rejectedBlock bool
+
+	l.intercept(func(entry zapcore.Entry) error {
```
yacovm, line 291: Needs to be right after the creation, otherwise we override the reference when intercepting, while goroutines try to read the reference

samliok (approved):

PR #63, opened 2025-01-28
Log leader ID when advancing rounds

yacovm:


epoch.go
```diff
@@ -1282,7 +1282,9 @@ func (e *Epoch) voteOnBlock(block Block) (Vote, error) {
 }
 
 func (e *Epoch) increaseRound() {
-	e.Logger.Info(fmt.Sprintf("Moving to a new round (%d --> %d", e.round, e.round+1), zap.Uint64("round", e.round+1))
+	leader := LeaderForRound(e.nodes, e.round)
+	e.Logger.Info(fmt.Sprintf("Moving to a new round (%d --> %d", e.round, e.round+1),
```
StephenButtolph, line 1286: fwiw - I think it's probably an anti-pattern to include formatted strings into a structured log output (as the log won't be able to be easily matched by a consumer)

StephenButtolph (approved):

epoch.go
yacovm: ok I removed the formatting

PR #70, opened 2025-01-31
Add replication support for lagging nodes

samliok:
# Add replication support for lagging nodes

This PR adds support for nodes to catch up when they fall behind in consensus. The changes introduce a request/response system that allows nodes to request missing finalization certificates and latest round information from other nodes in the network.

The replication process works as follows:

1. When a node receives a finalization certificate from the future, it begins broadcasting requests for the missing `fcerts` and `blocks`(up to `maxRoundWindow` ahead). It also sends a `LatestRoundRequest` in order to fully catch up to the latest round.
2. As responses arrive, certificates are either persisted to storage, or added to a future messages map where they will wait until `fCerts` from previous rounds arrive. 
4. Once we catch up with fCerts, we can load the `block`, `notarization` and potentially `fCert` from the latest round request earlier(TODO)

## TODO
- [x]  Add support for receiving `fCerts` from > `maxRoundWindow` away. This essentially requires adding a callback to restart requesting nodes for new `fCerts`. 
 - [x] Finish writing tests for the replication process. I've started writing an initial test for replication but this will require more work since I need to set up a custom network/communication struct(or fit the ones we have for `MultiNodeTest` in). The other tests I'd like to write are outlined in `replication_test.go`
 - [ ] We never do anything with the latest round information yet. But when we reach that round we should process the latest round and insert them into our `rounds` map


replication.go
```diff
@@ -0,0 +1,201 @@
...
+	for seq := start; seq <= end; seq++ {
+		// also request latest round in case this fCert is also behind
+		roundRequest := &Request{
+			LatestRoundRequest: &LatestRoundRequest{},
+			FinalizationCertificateRequest: &FinalizationCertificateRequest{
+				Seq: seq,
+			},
+		}
+		msg := &Message{Request: roundRequest}
+		comm.Broadcast(msg)
```
yacovm, line 198: By broadcasting the message, won't we cause a data duplication in the network and make all nodes send us duplicate information?

yacovm: Why not request each block from only a single node?

epoch.go
```diff
@@ -378,11 +376,35 @@ func (e *Epoch) handleFinalizationCertificateMessage(message *FinalizationCertif
...
 
+func (e *Epoch) collectFutureFinalizationCertificates(fCert *FinalizationCertificate) {
+	fCertRound := fCert.Finalization.Round
+	endSeq := math.Min(float64(fCertRound), float64(e.maxRoundWindow+e.round))
+	if e.latestRoundKnown.num >= uint64(endSeq) {
+		// we have already sent out messages collecting future finalization certificates
+		return
+	}
+	e.setLastReceivedFCertSeq(fCertRound)
+	e.sendFutureCertficatesRequests(e.round, uint64(endSeq), e.Comm)
```
yacovm, line 405: `e.round` and `endSeq` are different units. Since you may progress the round without progressing the sequence, it looks incorrect to me to mix them together. 

e.round should probably be the next sequence to what we're currently at.

yacovm, line 405: why pass `e.Comm` as a parameter if `sendFutureCertficatesRequests` is an instance method?

replication.go
```diff
@@ -0,0 +1,201 @@
...
+	LatestRoundRequest             *LatestRoundRequest
+}
+
+type Response struct {
+	FinalizationCertificateResponse *FinalizationCertificateResponse
+	LatestRoundResponse             *LatestRoundResponse
+}
+
+// request a finalization certificate for the given sequence number
+type FinalizationCertificateRequest struct {
```
yacovm, line 23: I think this should be:

```
type FinalizationCertificateRequest struct {
	Sequences []uint64
}
```

Because what if we want to ask from a single node, 100 sequences? Will we send it 100 different messages? If the remote node has a buffer that restricts message number, it will overflown.

replication.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (e *Epoch) sendFutureCertficatesRequests(start uint64, end uint64, comm Communication) {
+	if e.lastSequenceRequested >= end {
+		// no need to resend
+		return
+	}
+
+	// we want to collect
+	for seq := start; seq <= end; seq++ {
+		// also request latest round in case this fCert is also behind
```
yacovm, line 190: if we always ask for the latest round, then why not just always include that in the response?

replication.go
```diff
@@ -0,0 +1,201 @@
...
+type FinalizationCertificateResponse struct {
+	FCert FinalizationCertificate
+	Block Block
+}
+
+type LatestRoundRequest struct {
+}
+
+// if block has not been proposed yet, we should return the last round's block, notarization, and fCert
+type LatestRoundResponse struct {
```
yacovm, line 36: Maybe it would be more succinct to just send the notarization and fCert without the block? And then the requester would have the freedom to request the block if it needs it, just to reduce the network bandwidth.

replication.go
```diff
@@ -0,0 +1,201 @@
...
+	Block        Block
+	Notarization *Notarization
+	FCert        *FinalizationCertificate
+}
+
+func (e *Epoch) HandleRequest(req Request) *Response {
```
yacovm, line 42: > // TODO: should I update requests to be async? and have the epoch respond with e.Comm.Send(msg, node)

Yeah for now I think it makes sense for the comm to handle this request, and later on we could delegate this to a separate construct.

replication.go
```diff
@@ -0,0 +1,201 @@
...
+		response.FinalizationCertificateResponse = e.handleFinalizationCertificateRequest(req.FinalizationCertificateRequest)
+	}
+	if req.LatestRoundRequest != nil {
+		response.LatestRoundResponse = e.handleLatestBlockRequest()
+	}
+
+	return response
+}
+
+func (e *Epoch) handleFinalizationCertificateRequest(req *FinalizationCertificateRequest) *FinalizationCertificateResponse {
```
yacovm, line 56: I think we should first find this in the memory before going to storage. We have a method `locateBlock(seq uint64, digest []byte)` which does just that only it expects the block to have the specified hash. We could replace the `digest []byte` with a predicate function and re-use it here.

replication.go
```diff
@@ -0,0 +1,201 @@
...
+			return &LatestRoundResponse{}
+		}
+		return &LatestRoundResponse{
+			Block:        block,
+			FCert:        &fCert,
+			Notarization: nil,
+		}
+	}
+
+	latest := &LatestRoundResponse{
```
yacovm, line 96: It's not possible to have the same `e.round` both notarized and finalized at the same time.

If we have notarized the round, we'd move to `e.round+1` immediately and may or may not finalize it. 

Therefore, `e.round` always represents the current round we're waiting to notarize at the moment.

I think that the simplest thing to do, is to just always grab the latest finalization we have, and the latest notarization we have, regardless of the round we're at.

replication.go
samliok: I thought we discussed that sending to one node would overwhelm that node. Also if that node doesn't respond then our node would have wasted time waiting. Would it be better to randomly select a few nodes in the network to send?

yacovm: > I thought we discussed that sending to one node would overwhelm that node. Also if that node doesn't respond then our node would have wasted time waiting. Would it be better to randomly select a few nodes in the network to send?

I think that if a node doesn't respond, we can then just ask the block from a different one later. 
However, sending to everyone is needless data amplification. We wouldn't want to receive xN times the data we want to receive. This is expensive to the other nodes.

replication.go
samliok: since we delete the rounds entry after persisting a `finalizationCertificate` we would never have a fCert and block in memory.

util.go
```diff
@@ -13,10 +13,22 @@ func RetrieveLastBlockFromStorage(s Storage) (Block, error) {
...
 	}
-
 	lastBlock, _, retrieved := s.Retrieve(height - 1)
 	if !retrieved {
 		return nil, fmt.Errorf("failed retrieving last block from storage with seq %d", height-1)
 	}
 	return lastBlock, nil
 }
+
+func RetrieveLastFromStorage(s Storage) (Block, FinalizationCertificate, error) {
```
yacovm, line 23: There is not much use in making a new method almost identical as the one above. 

We can just change the function above to return an fCert and just use it. 

I'd say alternatively delete the method above and use this one, but the above method has a unit test and this doesn't.

replication_test.go
```diff
@@ -0,0 +1,256 @@
...
+		},
+	}, from)
+	require.NoError(t, err)
+
+	// wait for the block to be processed
+	time.Sleep(50 * time.Millisecond)
+	return block
+}
+
+func newStorage(t *testing.T, nodes []simplex.NodeID, bb simplex.BlockBuilder, seqs uint64) []simplex.SequenceData {
```
yacovm, line 237:
```suggestion
func createBlocks(t *testing.T, nodes []simplex.NodeID, bb simplex.BlockBuilder, seqCount uint64) []simplex.SequenceData {
	logger := testutil.MakeLogger(t, int(0))
	ctx := context.Background()
	data := make([]simplex.SequenceData, 0, seqCount)
	var prev simplex.Digest
	for i := uint64(0); i < seqCount; i++ {
		protocolMetadata := simplex.ProtocolMetadata{
			Seq: i,
			Round: i,
			Prev: prev,
		}
		
		block, ok := bb.BuildBlock(ctx, protocolMetadata)
		require.True(t, ok)
		prev = block.BlockHeader().Digest
		fCert, _ := newFinalizationRecord(t, logger, &testSignatureAggregator{}, block, nodes)
		data = append(data, simplex.SequenceData{
			Block: block,
			FCert: fCert,
		})
	}

	return data
}
```

replication_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+// }
+
+// // We replicate to the latest round(which doesn't have a fCert)
+// // ensuring that we have the latest block and notarization
+// func TestReplicationLatestRoundNotarized() {
+
+// }
+
+// // We get an FCert from the future, but it is actually not the global latest
```
yacovm, line 211: what are these commented out tests for? are they blueprints for future tests?

epoch_multinode_test.go
```diff
@@ -223,10 +235,41 @@ func (c *testComm) Broadcast(msg *Message) {
...
+func (n *inMemNetwork) addInstances(t *testing.T, node ...*testNode) {
+	allowed := false
+	for _, n := range node {
+		if n.e.ID.Equals(n.e.ID) {
+			allowed = true
+			break
+		}
+	}
+	require.True(t, allowed, "node not allowed to join network")
+	n.instances = append(n.instances, node...)
```
yacovm, line 260: I'm not sure why we moved this from `newSimplexNode`, can you explain this refactoring?

samliok: i moved it so that adding instances to the network would be more explicit. I'll undo this change because i think this was unnecessary

epoch.go
```diff
@@ -121,6 +128,15 @@ func (e *Epoch) HandleMessage(msg *Message, from NodeID) error {
 		return e.handleFinalizationMessage(msg.Finalization, from)
 	case msg.FinalizationCertificate != nil:
 		return e.handleFinalizationCertificateMessage(msg.FinalizationCertificate, from)
+	case msg.Response != nil:
+		return e.handleResponse(msg.Response, from)
+	case msg.Request != nil:
+		resp := e.HandleRequest(msg.Request, from)
+		msg := &Message{Response: resp}
+		if resp != nil {
```
yacovm, line 136: can we please fold these lines into `HandleRequest` ?

epoch.go
```diff
@@ -353,21 +367,13 @@ func (e *Epoch) Stop() {
...
-		e.Logger.Debug("Received finalization certificate for a non existent round", zap.Int("round", int(message.Finalization.Round)))
-		return nil
-	}
-
-	if round.fCert != nil {
-		e.Logger.Debug("Received finalization for an already finalized round", zap.Uint64("round", message.Finalization.Round))
+		zap.Stringer("from", from), zap.Uint64("round", message.Finalization.Round), zap.Uint64("seq", message.Finalization.Seq))
+	nextSeqToCommit := e.Storage.Height()
+	if nextSeqToCommit > message.Finalization.Seq {
+		// this may happen since we delete rounds after finalizing them
```
yacovm, line 374: > 		// this may happen since we delete rounds after finalizing them

I don't understand how this is related. Can you explain?

We may receive a finalization certificate from a node that is behind us, why is this related to deleting rounds?

epoch.go
```diff
@@ -378,11 +384,57 @@ func (e *Epoch) handleFinalizationCertificateMessage(message *FinalizationCertif
 			zap.Stringer("NodeID", from))
 		return nil
 	}
+
+	round, exists := e.rounds[message.Finalization.Round]
+	if !exists {
+		e.Logger.Debug("Received finalization certificate for a future round", zap.Uint64("round", message.Finalization.Round))
+		e.collectFutureFinalizationCertificates(message)
```
yacovm, line 391: While it make sense to collect a finalization certificate for a round we don't have, is this the only trigger we should be acting upon? 

What if we have a round instantiated but we don't have the round finalized? 

We only delete rounds if we have finalized all blocks from genesis until that round. 

However we may have the following blocks in the WAL: `b, b'` where we have finalized `b'` but not `b`. In that case we'll have a round for both `b` and `b'` but once we get a finalization for `b'` we won't ask for a finalization for `b` because the round for `b'` exists.

epoch.go
samliok: ah this comment is old but the check is still relevant.  updated the comment

epoch.go
```diff
@@ -378,11 +384,57 @@ func (e *Epoch) handleFinalizationCertificateMessage(message *FinalizationCertif
...
+		return nil
```
yacovm, line 389: Another issue I see with sending a request for synchronization immediately upon seeing a finalization that we don't have a round for, is that we may just be slow and therefore receive blocks slowly. 

Consider a node with a low downstream bandwidth - 10Mbps, or 2.5MBps.  It takes 400ms to transmit a 1MB block, but ~ 50ms to transmit a finalization certificate. 

Such a node may observe a finalization certificate before it has observed the proposal of the block and request it needlessly, yet if it would've waited for an extra 350ms, it would've received that block. 

I think it might make sense to only send the request if we haven't progressed within a configurable time frame (maybe default it to 2 seconds?)

epoch.go
samliok: >What if we have a round instantiated but we don't have the round finalized?

if we have the round instantiated but not finalized, we will still request `fCerts` for those rounds. This is because we request fCerts starting at the `nextSequenceToCommit` rather than `e.round`.

> but once we get a finalization for b' we won't ask for a finalization for b because the round for b' exists.

good point. I think a better trigger would be if we receive a `finalizationCert` for a round greater than `nextSequenceToCommit`.

epoch.go
samliok: i see what you are saying, but this would require us storing future finalization certificates even if we didn't get a block proposal yet. Before this PR we would reject fCerts if we didn't receive block proposals beforehand. Instead of adding a timeout & callback a simpler approach could be to only request `fCerts` if we are 2 blocks behind?

yacovm: > i see what you are saying, but this would require us storing future finalization certificates even if we didn't get a block proposal yet. 

Which is what we do for practically any message, so far. 

> Before this PR we would reject fCerts if we didn't receive block proposals beforehand. 

Right but we do store finalizations in the future messages, so it makes sense to do the same for finalization certificates.  The reason the code is as it is, is because we just didn't get to it yet.

> Instead of adding a timeout & callback a simpler approach could be to only request fCerts if we are 2 blocks behind?

Then we have a different problem - we may be one block behind because we have missed the proposal, and not replicate this block for a prolong time duration due to no new transactions being sent.

epoch.go
```diff
@@ -378,11 +381,57 @@ func (e *Epoch) handleFinalizationCertificateMessage(message *FinalizationCertif
...
+
 	round.fCert = message
 
 	return e.persistFinalizationCertificate(*message)
 }
 
+func (e *Epoch) collectFutureFinalizationCertificates(fCert *FinalizationCertificate) {
+	fCertRound := fCert.Finalization.Round
+	// Don't exceed the max round window
+	endSeq := math.Min(float64(fCertRound), float64(e.maxRoundWindow+e.round))
```
yacovm, line 405: we are interested in replicating blocks and finalizations, right? 

Then why are we looking at `e.round` and not at the next seq to commit?

yacovm: we may be in a high round but missing finalizations before it.

samliok: i chose `e.round`  because the `maxRoundWindow` relates to the window of storing future messages(starting from e.round->e.round + maxRoundWindow).  Therefore it should be fine to request sequences up to that round since we allow other messages from those round.

yacovm (commented):
I made a pass and gave some comments. While the review is far from being done, I thought I'd release some comments in the meanwhile.

My general high level comment is that adding more code paths of an orthogonal procedure (replication) that touch the same state makes testing harder, as we can test Simplex block agreement on its own, and test replication on its own, but testing the two intertwined is extremely complex.

This is why I suggest we think of a way of having the block agreement and replication avoid touching the same state as much as possible, to make testing simpler and minimize risk.

message_handler.go
```diff
@@ -0,0 +1,139 @@
...
+			e.storeFutureFinalizationResponse(data.FCert, data.Block, from)
+			continue
+		}
+		if round.fCert != nil {
+			// we should never be here because the round would have been deleted
+			e.Logger.Error("Received finalization certificate for a round that already has a finalization certificate", zap.Uint64("round", data.FCert.Finalization.Seq))
+			continue
+		}
+		err := e.persistFinalizationCertificate(data.FCert)
+		if err != nil {
```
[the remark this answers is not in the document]
samliok: > so there is little point in persisting the fCerts.

This point is making me very confused. I was under the impression that replication should sync the lagging node up to the most recent state to eventually become fully functional(and fault tolerant again). If we do not care about persisting `fCerts` then this node will never be fully functional. 

For example, there is so much logic in our consensus code that relies on the height of the storage. If we don't care about persisting to storage then any logic requiring the most recent storage height and data will always be out of sync once a node gets behind. Secondly, if we don't persist fCerts this node will never be able to help other nodes in the case they fall behind(since we wont have fCerts to send help replicate)?

samliok:
> This is why I suggest we think of a way of having the block agreement and replication avoid touching the same state as much as possible, to make testing simpler and minimize risk.

I think we should maybe sync up offline to talk about how to better structure this logic. I actually wrote the PR with the aim of trying to re-use the same consensus logic and state, as I was trying to emulate the standard path during the execution of rounds. Otherwise if the logic ever changes, we'd need to remember to update the logic in multiple code paths. That is why i re-used the consensus message handling functions, since I was under the impression that a lagging node should perform the same steps.

For example I wanted to re-use the logic of `persistFinalizationCertificate` and `createBlockVerificationTask` since storing `fCerts` and verifying blocks should be the same regardless of if we are in `replication` or during normal consensus. However, `persistFinalizationCertificate` doesn't just store the fCert in storage, it also broadcasts the finalization and starts the new round. Similarly, our current logic for blockVerification in `createBlockVerificationTask` verifies the block but also stores the vote by the proposer and then proceeds to vote on the block. Would it make sense to decouple these methods a little so they can be re-used in both paths?

Maybe my main confusion and second guessing is coming from this [comment](https://github.com/ava-labs/Simplex/pull/70#discussion_r1947496949), as I don't understand why we shouldn't persist the fCert. If we don't need to persist the `fCert` then I guess the easiest solution would be to verify the blocks and fCerts as we receive them, without worrying about order. A big reason why I used the current logic for `futureMessages` struct and calling the `handleX` functions is because finalizations need to be indexed in-order. Without this constraint we could definitely simplify the logic but I'm not sure if this is what you're referring to. 

> state makes testing harder, as we can test Simplex block agreement on its own, and test replication on its own, but testing the two intertwined is extremely complex.

I tried to make the tests quite simple, and they came out pretty clean.  The `replication` tests are less than 30 lines and pretty straightforward. I feel like this is pretty manageable but maybe they aren't testing the right things?

replication_test.go
```diff
@@ -0,0 +1,187 @@
...
+	normalNode3 := newSimplexNodeWithStorage(t, nodes[2], net, bb, storageData)
+	laggingNode := newSimplexNode(t, nodes[3], net, bb)
+
+	firstBlock := storageData[0].Block
+	record := simplex.BlockRecord(firstBlock.BlockHeader(), firstBlock.Bytes())
+	laggingNode.wal.Append(record)
+
+	firstNotarizationRecord, err := newNotarizationRecord(laggingNode.e.Logger, laggingNode.e.SignatureAggregator, firstBlock, nodes[0:quorum])
+	require.NoError(t, err)
+	laggingNode.wal.Append(firstNotarizationRecord)
```
yacovm, line 139: I think it's better to inject votes / notarizations to the node directly and not rely on recovery logic. We're should aim to replicate (not pun intended) conditions that are as close to what would happen in a real scenario. 

In a real scenario we most likely wouldn't be replicating right after a crash, but while we're up and running.

epoch.go
```diff
@@ -78,6 +78,8 @@ type Epoch struct {
 	maxPendingBlocks               int
 	monitor                        *Monitor
 	cancelWaitForBlockNotarization context.CancelFunc
+
+	replicationState *ReplicationState
```
yacovm, line 82: would it be possible to abstract out the replication from Epoch? This way if we want to test something without replication getting in the way, we can just provide an empty implementation of it.

epoch.go
```diff
@@ -359,18 +365,12 @@ func (e *Epoch) Stop() {
...
-		zap.Stringer("from", from), zap.Uint64("round", message.Finalization.Round))
+		zap.Stringer("from", from), zap.Uint64("round", message.Finalization.Round), zap.Uint64("seq", message.Finalization.Seq))
 
-	round, exists := e.rounds[message.Finalization.Round]
-	if !exists {
-		e.Logger.Debug("Received finalization certificate for a non existent round", zap.Int("round", int(message.Finalization.Round)))
-		return nil
-	}
-
-	if round.fCert != nil {
```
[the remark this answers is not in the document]
samliok: because I wanted to make sure the fCert was valid before sending a request to collect other certificates. Otherwise a node can just send us invalid finalizations for rounds way in advance which would cause unnecessary network traffic.

epoch.go
```diff
@@ -1158,7 +1199,6 @@ func (e *Epoch) buildBlock() {
 	metadata := e.metadata()
 
 	task := e.createBlockBuildingTask(metadata)
-
```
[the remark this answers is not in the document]
samliok: no, i think I accidentally delete/add blank lines when I'm editing the code and forget to reset them back

epoch.go
samliok: assuming this is an old comment?

msg.go
```diff
@@ -11,6 +11,8 @@ type Message struct {
 	Notarization            *Notarization
 	Finalization            *Finalization
 	FinalizationCertificate *FinalizationCertificate
+	Response                *Response
```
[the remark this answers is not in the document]
samliok: i was hoping these would be generic, and within them we would multiplex them to the correct request/response handler.  Right now they are only used for replication, but in the future we may have other response/request handling unrelated to replication.

yacovm (commented):
Good job on decoupling the replication code path from the standard consensus path. 

I am wondering if it's possible to even further decouple it though, and for example - not use the rounds map at all?

replication.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	valid, err := IsFinalizationCertificateValid(&data.FCert, r.quorumSize, r.logger)
+	// verify the finalization certificate
+	if err != nil || !valid {
+		return fmt.Errorf("Finalization certificate failed verification")
+	}
+
+	// we should never receive two valid fCerts for the same round, check anyways
+	if _, ok := r.receivedFinalizationCertificates[data.FCert.Finalization.Seq]; ok {
```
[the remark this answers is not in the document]
samliok: not disallowed, i'll make it so we return early with no error though

epoch.go
```diff
@@ -623,48 +644,13 @@ func (e *Epoch) progressRoundsDueToCommit(round uint64) {
...
-			if fCert.Finalization.Round+e.maxRoundWindow < e.round {
-				delete(e.rounds, fCert.Finalization.Round)
-			}
-			// Clean up the future messages - Remove all messages we may have stored for the round
-			// the finalization is about.
-			for _, messagesFromNode := range e.futureMessages {
-				delete(messagesFromNode, fCert.Finalization.Round)
-			}
-
-			// Check if we can commit the next round
```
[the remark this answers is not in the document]
samliok: hmmm, i think all this logic is necessary for replication. Say we haven't indexed a block yet and our rounds map has blocks and notarizations for rounds `1-3`. It's possible we have finalizations for round `2` and `3`, but not 1. In this case we would be considered a lagging node and ask for the fcert of round 1. When we process the fcert received for round `1` from`ReplicationResponse`, we also want to commit rounds `2` & `3` as well as remove them from the rounds cache(similar to how we do before).

replication.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (r *ReplicationState) StoreFinalizedBlock(data FinalizedBlock) error {
+	// ensure the finalization certificate we get relates to the block
+	blockDigest := data.Block.BlockHeader().Digest
+	if !bytes.Equal(blockDigest[:], data.FCert.Finalization.BlockHeader.Digest[:]) {
+		return fmt.Errorf("Finalization certificate does not match the block")
+	}
+
+	valid, err := IsFinalizationCertificateValid(&data.FCert, r.quorumSize, r.logger)
```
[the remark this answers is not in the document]
samliok: i think it does?
```
func IsFinalizationCertificateValid(fCert *FinalizationCertificate, quorumSize int, logger Logger) (bool, error) {
	valid, err := validateFinalizationQC(fCert, quorumSize, logger)
	if err != nil {
		return false, err
	}
```

epoch.go
```diff
@@ -974,6 +1048,52 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
...
+		e.lock.Lock()
+		defer e.lock.Unlock()
+
+		if !e.storeProposal(block) {
+			e.Logger.Warn("Unable to store proposed block for the round", zap.Uint64("round", md.Round))
+			return md.Digest
+			// TODO: timeout
+		}
+
+		// add the fCert to the round
```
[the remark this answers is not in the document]
samliok: good point! removed using the rounds map in the verification task. However I still think we will still need to touch it in this specific case https://github.com/ava-labs/Simplex/pull/70/files#r1953899419

replication.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+	return NodeID{}
+}
+
+func (r *ReplicationState) collectFutureFinalizationCertificates(fCert *FinalizationCertificate, currentRound uint64, nextSeqToCommit uint64) {
+	fCertRound := fCert.Finalization.Round
+	// Don't exceed the max round window
+	endSeq := math.Min(float64(fCertRound), float64(r.maxRoundWindow+currentRound))
+	if r.highestFCertReceived == nil || fCertRound > r.highestFCertReceived.Finalization.Seq {
```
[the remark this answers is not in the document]
samliok: no this is intended. the check for double sends is done with `lastSequenceRequested`. The reason why we shouldn't exit early here is because we may need to send many requests in the case that `highestFCertReceived` is from a round much further than `MaxRoundWindow`. 

we do call this with the same `highestFcertReceived` in 
```go
func (r *ReplicationState) maybeCollectFutureFinalizationCertificates(round uint64, nextSequenceToCommit uint64) {
```

replication.go
samliok: done

epoch.go
```diff
@@ -926,6 +954,52 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, from NodeID) error {
...
+		seqDigest := finalizedBlock.FCert.Finalization.BlockHeader.Digest
+		if !bytes.Equal(roundDigest[:], seqDigest[:]) {
+			e.Logger.Warn("Received finalized block that is different from the one we have in the rounds map",
+				zap.Stringer("roundDigest", roundDigest), zap.Stringer("seqDigest", seqDigest))
+			return nil
+		}
+		round.fCert = &finalizedBlock.FCert
+		e.indexFinalizationCertificates(round.num)
+		e.processReplicationState()
+		return e.maybeLoadFutureMessages()
```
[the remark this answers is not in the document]
samliok: yep, i think you're right. Removed the call to futureMessages

replication.go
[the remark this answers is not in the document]
yacovm: but `validateFinalizationQC` itself doesn't return an error:

```
func validateFinalizationQC(fCert *FinalizationCertificate, quorumSize int, logger Logger) (bool, error) {
	if fCert.QC == nil {
		return false, nil
	}

	// Check enough signers signed the finalization certificate
	if quorumSize > len(fCert.QC.Signers()) {
		logger.Debug("ToBeSignedFinalization certificate signed by insufficient nodes",
			zap.Int("count", len(fCert.QC.Signers())),
			zap.Int("Quorum", quorumSize))
		return false, nil
	}

	signedTwice := hasSomeNodeSignedTwice(fCert.QC.Signers(), logger)

	if signedTwice {
		return false, nil
	}

	if err := fCert.Verify(); err != nil {
		return false, nil
	}

	return true, nil
}
```

epoch.go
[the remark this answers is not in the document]
yacovm: fair enough.

epoch.go
```diff
@@ -766,7 +794,7 @@ func (e *Epoch) persistNotarization(notarization Notarization) error {
 
 	e.increaseRound()
 
-	return errors.Join(e.doNotarized(notarization.Vote.Round), e.maybeLoadFutureMessages(e.round))
```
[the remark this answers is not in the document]
yacovm: yes that's what i meant.

epoch.go
```diff
@@ -820,6 +799,9 @@ func (e *Epoch) persistFinalizationCertificate(fCert FinalizationCertificate) er
 			zap.Int("size", len(recordBytes)),
 			zap.Stringer("digest", fCert.Finalization.BlockHeader.Digest))
 
+		// we receive a finalization certificate for a future round
+		e.Logger.Debug("Received a finalization certificate for a future sequence", zap.Uint64("seq", fCert.Finalization.Seq), zap.Uint64("nextSeqToCommit", nextSeqToCommit))
+		e.replicationState.collectFutureFinalizationCertificates(&fCert, e.round, nextSeqToCommit)
```
yacovm, line 804: what is the reason we call `collectFutureFinalizationCertificates` here? Is calling it in `handleFinalizationCertificateMessage` not sufficient?

epoch.go
```diff
@@ -1696,12 +1804,108 @@ func (e *Epoch) storeProposal(block Block) bool {
...
+	return nil
+}
+
+func (e *Epoch) processReplicationState() {
+	nextSeqToCommit := e.Storage.Height()
+	finalizedBlock, ok := e.replicationState.receivedFinalizationCertificates[nextSeqToCommit]
+	if !ok {
+		return
+	}
+
```
yacovm, line 1904: isn't this where we probably should delete `finalizedBlock` from `receivedFinalizationCertificates` ? I don't see where we delete it.

epoch.go
```diff
@@ -1696,12 +1804,108 @@ func (e *Epoch) storeProposal(block Block) bool {
...
+	e.replicationState.maybeCollectFutureFinalizationCertificates(e.round, e.Storage.Height())
```
yacovm, line 1905: I might be missing something but why are we sending requests for blocks here ?

epoch.go
samliok: we call it here in case we receive a finalization for a future round that _is_ in our rounds map. The call above only collects if the round doesn't exist, however we still may be behind if it does.

epoch.go
samliok: in case we have progressed enough rounds to send out more certificates. It might make more sense to call it when we progress the round due to a commit, but I think its also fine calling here as well

epoch_test.go
```diff
@@ -766,6 +766,7 @@ func (mem *InMemStorage) Index(block Block, certificate FinalizationCertificate)
 
 	_, ok := mem.data[seq]
 	if ok {
+		fmt.Println("already indexed1!!")
```
yacovm, line 769: can you remove this?

replication.go
```diff
@@ -0,0 +1,137 @@
...
+
+	startSeq := math.Max(float64(nextSeqToCommit), float64(r.lastSequenceRequested))
+	r.logger.Debug("Node is behind, requesting missing finalization certificates", zap.Uint64("round", fCertRound), zap.Uint64("startSeq", uint64(startSeq)), zap.Uint64("endSeq", uint64(endSeq)))
+	r.sendFutureCertficatesRequests(uint64(startSeq), uint64(endSeq))
+}
+
+// sendFutureCertficatesRequests sends requests for future finalization certificates for the
+// range of sequences [start, end] <- inclusive
+func (r *ReplicationState) sendFutureCertficatesRequests(start uint64, end uint64) {
+	if r.lastSequenceRequested >= end {
```
yacovm, line 67: this condition is checked already in line 55

replication.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	roundRequest := &ReplicationRequest{
+		FinalizationCertificateRequest: &FinalizationCertificateRequest{
+			Sequences: seqs,
+		},
+	}
+	msg := &Message{ReplicationRequest: roundRequest}
+
+	requestFrom := r.requestFrom()
```
yacovm, line 83: would it make sense (as a follow-up PR) to request the sequences from the nodes that helped construct the fCert? We can get the list from `fCert.QC.Signers()`.

yacovm (approved):
LGTM and thanks for bearing through several review iterations!

epoch.go
```diff
@@ -1198,6 +1273,45 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
...
+		// possible we received a fCert for this block in the meantime. This check ensures we commit
+		// the block only if it is still the next sequence to commit.
+		if e.Storage.Height() != md.Seq {
+			e.Logger.Debug("Received finalized block that is not the next sequence to commit",
+				zap.Uint64("seq", md.Seq), zap.Uint64("height", e.Storage.Height()))
+			return md.Digest
+		}
+		e.indexFinalizationCertificate(block, finalizedBlock.FCert)
+		e.processReplicationState()
+		err := e.maybeLoadFutureMessages()
```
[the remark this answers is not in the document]
samliok: yep, also we may have received more messages during replication(such as blocks & notarizations) that we may need to start processing.

replication.go
samliok: noted in this issue
https://github.com/ava-labs/Simplex/issues/82

PR #72, opened 2025-02-05
Advance rounds when finalizing blocks without notarizing them

yacovm:


epoch.go
```diff
@@ -654,9 +654,17 @@ func (e *Epoch) assembleFinalizationCertificate(round *Round) error {
 	return e.persistFinalizationCertificate(fCert)
 }
 
+func (e *Epoch) progressRoundsDueToCommit(round uint64) {
```
samliok, line 657: in what scenario would increment round be called more than once here?

epoch_test.go
```diff
@@ -80,19 +145,25 @@ func notarizeAndFinalizeRound(t *testing.T, nodes []NodeID, round uint64, e *Epo
...
+			injectTestVote(t, e, block, nodes[i])
 		}
-		injectTestVote(t, e, block, nodes[i])
 	}
 
 	for i := 1; i < quorum; i++ {
 		injectTestFinalization(t, e, block, nodes[i])
 	}
 
+	if skipNotarization {
```
samliok, line 163: side not: when reviewing i realized we allow messages to be sent to ourselves. Should we add a check in `handleMessage` ensuring `e.ID` != `from`? Only thought of this because we could set the above for loop to start at 0, and remove this if statement.

samliok (approved):

epoch.go
yacovm: For now, it's really a check for *not* incrementing it, instead of incrementing it. The *for* acts both as an *if* and also as a *for* here. 

However in the future we may encounter a case where we are several empty blocks behind, and we have a block that we received, along with its finalization certificate. As we enter the block commit path, we have to increment our round according to the finalization certificate, which will be several rounds and not just one.

epoch_test.go
yacovm: yeah we can definitely do that.

PR #73, opened 2025-02-05
Use correct sequence and metadata when building block

yacovm:
When building a block, we look at the latest round and build the round on top of it, only if we have already committed a block already.

However, since we notarize rounds asynchronously, we may advance to a new round before we commit a block.

We therefore need to use the rounds themselves as an indicator.


samliok (approved):
approved, but have a follow up question: 

Why should we enforce `seq` = `round` when building a block? it's possible for the round and sequence to deviate(and expected in the case of failover).

PR #75, opened 2025-02-05
Decouple vote from handleBlockMessage

samliok:
When handling `blockMessages` we also process a vote attached to the message. During replication, we wont have this vote attached to blocks. This PR decouples the logic for handling votes and verifying blocks, allowing us to use the same logic during replication.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -103,12 +103,18 @@ func (e *Epoch) HandleMessage(msg *Message, from NodeID) error {
 		return nil
 	}
 
+	if from.Equals(e.ID) {
+		e.Logger.Warn("Received message from self")
+		return nil
+	}
+
 	// Guard against receiving messages from unknown nodes
 	_, known := e.eligibleNodeIDs[string(from)]
 	if !known {
 		e.Logger.Warn("Received message from an unknown node", zap.Stringer("nodeID", from))
 		return nil
 	}
+
 
 	switch {
 	case msg.BlockMessage != nil:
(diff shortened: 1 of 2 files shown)


epoch.go
```diff
@@ -877,41 +877,15 @@ func (e *Epoch) hasSomeNodeSignedTwice(nodeIDs []NodeID) bool {
...
-func (e *Epoch) handleBlockMessage(message *BlockMessage, from NodeID) error {
-	block := message.Block
-	if block == nil {
-		e.Logger.Debug("Got empty block in a BlockMessage")
-		return nil
-	}
-
+// handleBlockMessageFromLeader expects the block message
+// to come from the leader of the round, otherwise the block will not be processed.
+func (e *Epoch) handleBlockMessageFromLeader(message *BlockMessage, from NodeID) error {
```
yacovm, line 882: probably this method should be called `handleBlockMessage` while the other one should be called `processBlock` or something like that, and have its arguments changed to not accept a block message but just the block.

epoch.go
```diff
@@ -971,7 +977,7 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, from NodeID) error {
 	defer e.deleteFutureProposal(from, md.Round)
 
 	// Create a task that will verify the block in the future, after its predecessors have also been verified.
-	task := e.createBlockVerificationTask(block, from, vote)
+	task := e.createBlockVerificationTask(block, from, message.Vote)
```
yacovm, line 980: where are we going to get a vote from, if we call this method in the context of replication? We don't have a vote from anywhere, yet we only assign the vote once we finished verifying the block.... isn't that a problem?

samliok: in the context of replication we would just pass the vote struct, but never use it. In my replication PR blockVerification will terminate before ever needing the vote since theres no point in handling processing votes when replicating.

samliok:
addind to replication pr

PR #76, opened 2025-02-10
Cleanup recovery from fCert in WAL

samliok:
Couple small improvements
- No need to call `persistFinalizationCertifcate` when recovering from the WAL
- Fixed tests to reflect WAL(we would never add a fCert to the wal if its the next sequence to commit).

### DIFF
--- a/epoch_test.go
+++ b/epoch_test.go
@@ -11,8 +11,6 @@ import (
 	"encoding/asn1"
 	"encoding/binary"
 	"fmt"
-	"github.com/stretchr/testify/require"
-	"go.uber.org/zap/zapcore"
 	"math"
 	rand2 "math/rand"
 	. "simplex"
@@ -20,6 +18,9 @@ import (
 	"simplex/wal"
 	"sync"
 	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zapcore"
 )
 
 func TestEpochFinalizeThenNotarize(t *testing.T) {
--- a/recovery_test.go
+++ b/recovery_test.go
@@ -60,7 +60,7 @@ func TestRecoverFromWALProposed(t *testing.T) {
 	err = e.Start()
 	require.NoError(t, err)
 
-	rounds := uint64(100)
+	rounds := uint64(1)
 	for i := uint64(0); i < rounds; i++ {
 		leader := LeaderForRound(nodes, uint64(i))
 		isEpochNode := leader.Equals(e.ID)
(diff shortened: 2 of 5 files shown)


recovery_test.go
```diff
@@ -477,59 +477,7 @@ func TestWalWritesFinalizationCert(t *testing.T) {
 }
 
 // TestRecoverFromMultipleRounds tests that the epoch can recover from a wal with multiple rounds written to it.
-func TestRecoverFromMultipleRounds(t *testing.T) {
```
samliok, line 480: we can remove this tests since we will never append a fCert that is the next sequence to commit. Without that check, this test becomes the same test as `TestRecoverFromNotarization`

recovery_test.go
```diff
@@ -476,60 +476,7 @@ func TestWalWritesFinalizationCert(t *testing.T) {
 	require.Equal(t, uint64(0), e.Storage.Height())
 }
 
-// TestRecoverFromMultipleRounds tests that the epoch can recover from a wal with multiple rounds written to it.
```
yacovm, line 479: why is this deleted?

yacovm (commented):
production code changes LGTM but not sure why was the test deleted, is it redundant at this point?

recovery_test.go
```diff
@@ -580,15 +528,25 @@ func TestRecoverFromMultipleNotarizations(t *testing.T) {
 	wal.Append(secondNotarizationRecord)
 
 	// Create finalization record for first block
```
yacovm, line 530: > // Create finalization record for first block

Isn't this for the second block?

yacovm (dismissed):

recovery_test.go
```diff
@@ -568,27 +586,33 @@ func TestRecoverFromMultipleNotarizations(t *testing.T) {
...
+	conf.Storage.Index(secondBlock, fCert2)
+	conf.Storage.Index(thirdBlock, fCer3)
+
+	e, err := NewEpoch(conf)
 	require.NoError(t, err)
+	require.Equal(t, uint64(3), e.Storage.Height())
+	require.NoError(t, e.Start())
 
+	// ensure the round is properly set to 3
 	require.Equal(t, uint64(3), e.Metadata().Round)
```
yacovm, line 617: can we add that the sequence is equal to 3 also? We had a bug with that, want to make sure when we recover we also compute it correctly.

epoch_test.go
```diff
@@ -11,15 +11,16 @@ import (
...
 	"math"
 	rand2 "math/rand"
 	. "simplex"
 	"simplex/testutil"
 	"simplex/wal"
 	"sync"
 	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zapcore"
```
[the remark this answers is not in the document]
yacovm: There is a reason I don't add golangci linter to our CI. It's a waste of time so early on 😉

yacovm (commented):
LGTM modulo comments

yacovm (approved):

PR #78, opened 2025-02-11
Move FCert Validation out of Epoch

samliok:


yacovm:
can you squash please?

util_test.go
```diff
@@ -44,12 +52,82 @@ func TestRetrieveFromStorage(t *testing.T) {
...
+	nodes := []NodeID{{1}, {2}, {3}, {4}, {5}}
+	quorumSize := Quorum(len(nodes))
+	signatureAggregator := &testSignatureAggregator{}
+	// Test
+	tests := []struct {
+		name        string
+		fCert       FinalizationCertificate
+		quorumSize  int
+		valid       bool
+		expectedErr error
```
yacovm, line 78: isn't expected error nil everywhere?

samliok: yep, updated

util_test.go
```diff
@@ -44,12 +52,78 @@ func TestRetrieveFromStorage(t *testing.T) {
...
+				return fCert
+			}(),
+			quorumSize: quorumSize,
+			valid: false,
+		},
+		{
+			name: "signer signed twice",
+			fCert: func () FinalizationCertificate{
+				block := newTestBlock(ProtocolMetadata{})
+				doubleNodes := append(nodes[:quorumSize], nodes[0])
```
yacovm, line 102: > doubleNodes := append(nodes[:quorumSize], nodes[0])

I think it's easier to just allocate the slice of the double nodes inline, no?  `[]NodeID{nodes[0], nodes[1], ..., nodes[0]}`

Also - I know that `nodes` doesn't have enough capacity to append, so a new slice is allocated. But had it enough capacity - it would change the original `nodes` and influence other tests.

util_test.go
```diff
@@ -44,12 +52,78 @@ func TestRetrieveFromStorage(t *testing.T) {
...
+				return fCert
+			}(),
+			quorumSize: quorumSize,
+			valid: false,
+		},
+		{
+			name: "quorum certificate not in finalization certificate",
+			fCert: func () FinalizationCertificate{
+				block := newTestBlock(ProtocolMetadata{})
+				fCert, _ := newFinalizationRecord(t, l, signatureAggregator, block, nodes[:quorumSize])
```
yacovm, line 113: shouldn't we just instantiate the fCert explicitly? Why initialize it "the right way" and then wipe out its QC? Most of what it has, is the QC. 

You can just initialize it explicitly by only giving it the block.Metadata()

samliok: updated

util_test.go
samliok: updated

yacovm (approved):

PR #79, opened 2025-02-11
Add Test For Blocks Sent by Non Leaders

samliok:


epoch_test.go
```diff
@@ -365,6 +365,56 @@ func TestEpochBlockSentTwice(t *testing.T) {
...
+	bb := &testBlockBuilder{out: make(chan *testBlock, 1)}
+	storage := newInMemStorage()
+
+	nodes := []NodeID{{1}, {2}, {3}, {4}}
+	conf := EpochConfig{
+		MaxProposalWait:     DefaultMaxProposalWaitTime,
+		Logger:              l,
+		ID:                  nodes[1],
+		Signer:              &testSigner{},
+		WAL:                 wal.NewMemWAL(t),
```
yacovm, line 388: shall we check that the WAL is empty?

samliok: updated

PR #85, opened 2025-02-14
    Prevent re-execution of a block in the same round

yacovm:
This commit prevents re-execution of a block of the same round, twice or more.
    
This prevents two things:
    
- A leader sending a block twice to the same node
- A node executing a block twice by receiving one from the leader via the consensus agreement path, and also receiving one from the replication code path.
    
It works by introducing a wrapper around the existing scheduler, which tracks the round number of the block being executed.

This doesn't prevent the same block from being proposed by different nodes in different rounds.

### DIFF
--- a/sched.go
+++ b/sched.go
@@ -5,6 +5,7 @@ package simplex
 
 import (
 	"go.uber.org/zap"
+	"math"
 	"sync"
 )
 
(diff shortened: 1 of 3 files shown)


sched.go
```diff
@@ -168,3 +169,34 @@ func (t *dependencies) Remove(id Digest) []task {
 	delete(t.dependsOn, id)
 	return dependents
 }
+
+// oneTimeBlockScheduler ensures each block is only scheduled once by forcing that blocks
+// would be scheduled in ascending order.
+type oneTimeBlockScheduler struct {
+	lock               sync.Mutex
+	scheduler          *scheduler
+	lastRoundScheduled uint64
```
samliok, line 178: i think we can just use an `atomic.uint64` rather than the variable and lock. especially since we don't need the lock when calling `scheduler.Schedule`

samliok (approved):

sched.go
yacovm: Good point, since we schedule under a lock we don't need any kind of synchronization.

The lock was there from a previous incarnation of the PR where I had a closure that was passed to the original scheduler and that closure was executed asynchronously.

PR #87, opened 2025-02-17
Allow Disconnects/Reconnects from TestComm

samliok:


epoch_multinode_test.go
```diff
@@ -246,15 +263,22 @@ type inMemNetwork struct {
 	t         *testing.T
 	nodes     []NodeID
 	instances []*testNode
+	connected map[string]bool
```
yacovm, line 266: 1. I think it's easier to store the complement - if the node is *not* connected, and let it be connected by default. 
2. Since the test goroutine is asynchronous to the test's communication goroutines, there needs to be some synchronization here. 
3. Let's store empty structs instead of booleans?

epoch_multinode_test.go
```diff
@@ -271,6 +295,23 @@ func (n *inMemNetwork) addNode(node *testNode) {
 	n.instances = append(n.instances, node)
 }
 
+func (n *inMemNetwork) IsConnected(node NodeID) bool {
```
yacovm, line 298: It will be faster if we look at if the node is *not* connected. This way we don't need to iterate over the instances each time and they can be an empty list by default.

epoch_multinode_test.go
```diff
@@ -271,6 +295,23 @@ func (n *inMemNetwork) addNode(node *testNode) {
...
+	for _, instance := range n.instances {
```
yacovm, line 299: Not sure I follow - why we're iterating over all instances, can't we just do a map lookup?

epoch.go
```diff
@@ -1205,10 +1205,10 @@ func (e *Epoch) processFinalizedBlock(finalizedBlock *FinalizedBlock) error {
 		return nil
 	}
 	md := finalizedBlock.Block.BlockHeader()
-	if !e.verifyProposalIsPartOfOurChain(finalizedBlock.Block) {
-		e.Logger.Debug("Got invalid block in a BlockMessage")
-		return nil
-	}
+	// if !e.verifyProposalIsPartOfOurChain(finalizedBlock.Block) {
```
samliok, line 1208: I need to hit up this issue first #92. will uncomment after.

epoch.go
```diff
@@ -1616,6 +1618,8 @@ func (e *Epoch) monitorProgress(round uint64) {
 		}
 	}
 
+	e.cancelWaitForBlockNotarization()
```
samliok, line 1621: temporary fix for the issue we talked about offline. Will remove after the issue is resolved

samliok, line 1620: temporary fix for the issue we talked about offline. Will remove after the issue is resolved

replication_test.go
```diff
@@ -172,6 +174,98 @@ func TestReplicationStartsBeforeCurrentRound(t *testing.T) {
...
+	
+	for _, n := range net.instances {
+		require.Equal(t, endDisconnect-missedSeqs+1, n.storage.Height()-1)
+		require.Equal(t, endDisconnect+2, n.e.Metadata().Round)
+	}
+}
+
+func advanceWithoutLeader(t *testing.T, net *inMemNetwork, bb *testControlledBlockBuilder, epochTimes []time.Time) {
+	for range net.instances {
+		bb.blockShouldBeBuilt <- struct{}{}
```
samliok, line 262: would it be better to have block builders for each individual node?

epoch.go
samliok, line 1208: I need to hit up this issue first #92. will uncomment after.

epoch.go
```diff
@@ -1215,10 +1215,6 @@ func (e *Epoch) processFinalizedBlock(finalizedBlock *FinalizedBlock) error {
 		return nil
 	}
 	md := finalizedBlock.Block.BlockHeader()
-	if !e.verifyProposalIsPartOfOurChain(finalizedBlock.Block) {
```
yacovm, line 1218: i don't think this is related to the PR

epoch.go
```diff
@@ -1636,6 +1632,8 @@ func (e *Epoch) monitorProgress(round uint64) {
 		}
 	}
 
+	e.cancelWaitForBlockNotarization()
```
yacovm, line 1635: Is this PR ready for review?

replication_test.go
```diff
@@ -230,7 +232,106 @@ func TestReplicationFutureFinalizationCertificate(t *testing.T) {
...
+//
+// All nodes make progress for `startDisconnect` blocks. The lagging node disconnects
+// and the rest of the nodes continue to make progress for another `endDisconnect - startDisconnect` blocks.
+// The lagging node reconnects and the after the next `fCert` is sent, the lagging node catches up to the latest height.
+func TestReplicationAfterNodeDisconnects(t *testing.T) {
+	bb := newTestControlledBlockBuilder(t)
+	nodes := []simplex.NodeID{{1}, {2}, {3}, []byte("lagging")}
+	net := newInMemNetwork(t, nodes)
+	startDisconnect := uint64(5)
+	endDisconnect := uint64(17) // TODO: create test where the lagging node is the leader after replication
```
yacovm, line 248: alternatively we can parametrize the startDisconnect and endDisconnect indices, and run through them in parallel. We don't need that much, I think a startDisconnect that starts from 0 and can go up to 5, and an endDisconnect that starts from, say, 10, and goes up to 20 should be sufficient. Overall 50 tests that can run in parallel.

replication_test.go
```diff
@@ -230,7 +232,106 @@ func TestReplicationFutureFinalizationCertificate(t *testing.T) {
...
+		require.Equal(t, startDisconnect, n.storage.Height())
+	}
+
+	// lagging node disconnects
+	net.Disconnect(nodes[3])
+
+	missedSeqs := uint64(0)
+	// normal nodes continue to make progress
+	for i := startDisconnect; i < endDisconnect; i++ {
+		bb.triggerNewBlock()
```
yacovm, line 284: why do we trigger new block when we have a empty round? I don't think we should do that. I know it's a no-op if the control buffer is full, but we're needlessly filling up the control buffer.

replication_test.go
yacovm: why?

replication_test.go
```diff
@@ -230,7 +232,106 @@ func TestReplicationFutureFinalizationCertificate(t *testing.T) {
...
+	for _, n := range net.instances {
+		require.Equal(t, endDisconnect-missedSeqs+1, n.storage.Height()-1)
+		require.Equal(t, endDisconnect+2, n.e.Metadata().Round)
+	}
+
+	isLaggingNodeLeader = bytes.Equal(simplex.LeaderForRound(nodes, endDisconnect+2), nodes[3])
+	require.True(t, isLaggingNodeLeader, "the lagging node should be the leader")
+
+	bb.triggerNewBlock() // the lagging node should build a block when triggered
+	for _, n := range net.instances {
```
yacovm, line 322: great test

epoch.go
```diff
@@ -142,8 +142,8 @@ func (e *Epoch) HandleMessage(msg *Message, from NodeID) error {
 		return e.handleEmptyVoteMessage(msg.EmptyVoteMessage, from)
 	case msg.Notarization != nil:
 		return e.handleNotarizationMessage(msg.Notarization, from)
-	//case msg.EmptyNotarization != nil:
-	//	return e.handleEmptyNotarizationMessage(msg.EmptyNotarization, from)
+	// case msg.EmptyNotarization != nil:
```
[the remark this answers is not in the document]
samliok: sure, i think running go fmt changed this for me

replication_test.go
samliok: so we can have finer grained control over the specific channel we are triggering. Just a thought, i think this is fine as well

replication_test.go
samliok: we need to trigger this so that the node that is disconnected doesn't hang

replication_test.go
samliok: not sure if this would work for certain cases. 

`	require.False(t, isLaggingNodeLeader, "the lagging node should not be the leader")`

We require this because after the node re-connects it needs to be sent an fcert to kick start replication. I can still create the tests but we'd need to skip  when `endDisconect+1` is the lagging node. Should I still do it?

replication_test.go
yacovm: Hang on what? We wait for a new block without the lock so we are open for new messages.

Can you elaborate?

samliok: because the lagging node waits on this channel so we need to send to it. Otherwise, since the block builder is shared across nodes, the next time we call `triggerNewBlock` the lagging node might receive it causing the other nodes to hang since they are waiting on the channel as well.

samliok: ah nevermind, fixed

yacovm (approved):

PR #88, opened 2025-02-17
Simplify Failover Test

samliok:
Found it hard to follow the failover tests due to the out of order checking of WAL appends.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -1245,7 +1245,14 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
 		defer e.lock.Unlock()
 
 		record := BlockRecord(md, block.Bytes())
-		e.WAL.Append(record)
+		err := e.WAL.Append(record)
+		if err != nil {
+			e.Logger.Error("Failed to append block record to WAL", zap.Error(err))
+			return md.Digest
+		}
+		e.Logger.Debug("Persisted block to WAL",
+			zap.Uint64("round", md.Round),
+			zap.Stringer("digest", md.Digest))
 
 		if !e.storeProposal(block) {
 			e.Logger.Warn("Unable to store proposed block for the round", zap.Stringer("NodeID", from), zap.Uint64("round", md.Round))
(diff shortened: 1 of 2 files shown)


epoch.go
```diff
@@ -1245,7 +1245,14 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
 		defer e.lock.Unlock()
 
 		record := BlockRecord(md, block.Bytes())
-		e.WAL.Append(record)
+		err := e.WAL.Append(record)
+		if err != nil {
+			e.Logger.Error("Failed to append block record to WAL", zap.Error(err))
```
yacovm, line 1250: why do we return `md.Digest` and not an error?

epoch_failover_test.go
```diff
@@ -61,30 +61,22 @@ func TestEpochLeaderFailover(t *testing.T) {
...
-
-	prev := lastBlock.BlockHeader().Digest
-
-	md := ProtocolMetadata{
-		Round: 3,
-		Seq:   2,
-		Prev:  prev,
-	}
+	
+	emptyBlockMd := e.Metadata()
```
yacovm, line 72: I would rather explicitly craft the block manually and not rely on the metadata because I want whoever reads the test to understand what's going on with the sequences and rounds, and also I don't want to rely on the correctness of the Metadata(), as this is essentially test input.

epoch.go
yacovm: ah right it's from that block verification closure, i didn't expand the entire diff so didn't notice it. 

I think what we should do is to:

1. Not continue further down that closure as you did
2. Asynchronously cause the node to halt via propagating an error from this callback (we're executing this under a lock so should be simple to just add an error variable somewhere)
3. Inside of`HandleMessage` we will check if the node is halted and if so, return that error.

What do you think?

samliok: sure that sounds good with me. should I make that into a new PR? id like to get this one merged first

yacovm: I think a new PR makes sense, but can you take out this WAL.Append error handling out of this PR?

samliok: done

yacovm (approved):

PR #89, opened 2025-02-17
Wait for block proposal

samliok:
making this change because in the case of multinode tests, its not sufficient to just intercept the logger messages.

### DIFF
--- a/epoch_failover_test.go
+++ b/epoch_failover_test.go
@@ -16,15 +16,7 @@ import (
 )
 
 func TestEpochLeaderFailover(t *testing.T) {
-	timeoutDetected := make(chan struct{})
-
 	l := testutil.MakeLogger(t, 1)
-	l.Intercept(func(entry zapcore.Entry) error {
-		if entry.Message == `Timed out on block agreement` {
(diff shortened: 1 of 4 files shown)


epoch_failover_test.go
```diff
@@ -387,17 +367,18 @@ func createEmptyVote(md ProtocolMetadata, signer NodeID) *EmptyVote {
 	return emptyVoteFrom2
 }
 
-func waitForEvent(t *testing.T, start time.Time, e *Epoch, events chan struct{}) {
-	now := start
+func waitForBlockProposerTimeout(t *testing.T, e *Epoch, startTime time.Time) {
```
yacovm, line 370: I think it's better to use the approach from `assertNotarization` instead of this one.

epoch_failover_test.go
```diff
@@ -387,17 +367,18 @@ func createEmptyVote(md ProtocolMetadata, signer NodeID) *EmptyVote {
...
+	startRound := e.Metadata().Round
```
yacovm, line 371: is it possible to pass the round we wish to wait for as a parameter?

Issue #81, opened 2025-02-14
Don't begin replication if block is being verified


samliok:
Currently we would send replication requests, which would use unnecessary bandwidth since we already have the block(it's just not in the rounds map yet). Rather, we should add the finalization certificate to future messages, and re-call `handleFinalizationCertificateMessage` after we receive a proposal.

PR #93, opened 2025-02-19
Don't Send FutureFCertRequests if currently verifying proposal. 

samliok:
Addresses #81

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -545,26 +545,6 @@ func (e *Epoch) storeFutureFinalization(message *Finalization, from NodeID, roun
 	msgsForRound.finalization = message
 }
 
-func (e *Epoch) storeFutureFinalizationCertificate(fCert *FinalizationCertificate, from NodeID, round uint64) {
-	msgsForRound, exists := e.futureMessages[string(from)][round]
-	if !exists {
-		e.Logger.Warn("Attempting to store a future finalization certificate for a round we have not seen yet",
-			zap.Uint64("round", round), zap.Stringer("from", from))
-		return
-	}
-	msgsForRound.finalizationCertificate = fCert
-}
-
-func (e *Epoch) futureProposalForRoundExists(round uint64) bool {
-	for _, msgs := range e.futureMessages {
-		msgForRound, exists := msgs[round]
-		if exists && msgForRound.proposal != nil {
-			return true
-		}
-	}
-	return false
-}
-
 func (e *Epoch) handleEmptyVoteMessage(message *EmptyVote, from NodeID) error {
 	vote := message.Vote
 
--- a/epoch.go
+++ b/epoch.go
@@ -442,6 +442,16 @@ func (e *Epoch) handleFinalizationCertificateMessage(message *FinalizationCertif
 
 	round, exists := e.rounds[message.Finalization.Round]
 	if !exists {
+		// delay collecting future finalization certificate if we are verifying the proposal for that round
+		// and the fCert is for the current round
+		msgsForRound, exists := e.futureMessages[string(from)][message.Finalization.Round]
+		if exists && msgsForRound.proposal != nil && message.Finalization.Round == e.round {
+			e.Logger.Debug("Received finalization certificate for this round, but we are still verifying the proposal",
+				zap.Uint64("round", message.Finalization.Round)) 
+			e.storeFutureFinalizationCertificate(message, from, message.Finalization.Round)
+			return nil
+		}
+
 		// TODO: delay requesting future fCerts and blocks, since blocks could be in transit
 		e.Logger.Debug("Received finalization certificate for a future round", zap.Uint64("round", message.Finalization.Round))
 		e.replicationState.collectFutureFinalizationCertificates(message, e.round, nextSeqToCommit)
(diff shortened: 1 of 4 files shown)


epoch_test.go
```diff
@@ -743,6 +743,7 @@ type testBlock struct {
 }
 
 func (tb *testBlock) Verify() error {
+	time.Sleep(1 * time.Second)
```
samliok, line 746: this is temporary. need to either make a separate block struct(`LongVerifyBlock`) or include a `verifyTime` field in `testBlock` that toggles the time it takes to verify.

epoch.go
```diff
@@ -442,6 +442,16 @@ func (e *Epoch) handleFinalizationCertificateMessage(message *FinalizationCertif
 
 	round, exists := e.rounds[message.Finalization.Round]
 	if !exists {
+		// delay collecting future finalization certificate if we are verifying the proposal for that round
+		// and the fCert is for the current round
+		msgsForRound, exists := e.futureMessages[string(from)][message.Finalization.Round]
```
yacovm, line 447: How are we sure we should get the proposal from the same node that sent us the finalization certificate?

epoch.go
```diff
@@ -1167,7 +1187,13 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, from NodeID) error {
 		return nil
 	}
 
-	defer e.deleteFutureProposal(from, md.Round)
+	// save in future messages while we are verifying the block
+	_, exists := e.futureMessages[string(from)][md.Round]
+	if !exists {
```
yacovm, line 1192: what if it exists? don't we need to store the proposal in that case?

epoch.go
samliok: updated, added a helper function

epoch.go
samliok: updated

epoch.go
```diff
@@ -529,6 +538,26 @@ func (e *Epoch) storeFutureFinalization(message *Finalization, from NodeID, roun
 	msgsForRound.finalization = message
 }
 
+func (e *Epoch) storeFutureFinalizationCertificate(fCert *FinalizationCertificate, from NodeID, round uint64) {
+	msgsForRound, exists := e.futureMessages[string(from)][round]
+	if !exists {
```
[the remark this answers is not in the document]
yacovm: Don't forget that when we load the future finalization we do it via:

```
		for from, messagesFromNode := range e.futureMessages {
			if msgs, exists := messagesFromNode[round]; exists {
```

so it doesn't really matter to which `nodeID` we associate it to.

replication_test.go
```diff
@@ -172,6 +173,65 @@ func TestReplicationStartsBeforeCurrentRound(t *testing.T) {
...
+	fCert, _ := newFinalizationRecord(t, l, signatureAggregator, block, nodes[0:quorum])
+	// send fcert
+	err = e.HandleMessage(&simplex.Message{
+		FinalizationCertificate: &fCert,
+	}, nodes[0])
+	require.NoError(t, err)
+
+	storedBlock := storage.waitForBlockCommit(0)
+	require.Equal(t, uint64(1), storage.Height())
+	// make sure it was indexed and no replication requests where sent
```
yacovm, line 230: we don't check that no replication requests were sent though

replication_test.go
```diff
@@ -172,6 +173,65 @@ func TestReplicationStartsBeforeCurrentRound(t *testing.T) {
 	}
 }
 
+func TestReplicationFutureFinalizationCertificate(t *testing.T) {
+	// send a block, then simultaneously send a finalization certificate for the block
+	l := testutil.MakeLogger(t, 1)
+	bb := newTestDelayedVerificationBlockBuilder(50 * time.Millisecond)
```
yacovm, line 179: the test passes even when i replace this with `newTestControlledBlockBuilder`. 

The 50ms sleep is not something we should rely on... can't we make something more controlled? 

How about doing this and then you can ditch the entire `testDelayedVerificationBlockBuilder` and its friends?

```
func (tb *testBlock) Verify() error {
	if tb.verificationDelay == nil {
		return nil
	}
	
	<- tb.verificationDelay
	
	return nil
}
```

Then all we need is to add:

```
	err = e.HandleMessage(&simplex.Message{
		FinalizationCertificate: &fCert,
	}, nodes[0])
	require.NoError(t, err)

	block.verificationDelay <- struct{}{} // <----

	storedBlock := storage.waitForBlockCommit(0)
```

replication_test.go
samliok: yea, i think checking for sent requests is actually unnecessary so i'm going to delete the comment

epoch.go
```diff
@@ -529,6 +545,26 @@ func (e *Epoch) storeFutureFinalization(message *Finalization, from NodeID, roun
 	msgsForRound.finalization = message
 }
 
+func (e *Epoch) storeFutureFinalizationCertificate(fCert *FinalizationCertificate, from NodeID, round uint64) {
```
yacovm, line 548: Isn't this redundant now?

epoch.go
```diff
@@ -529,6 +545,26 @@ func (e *Epoch) storeFutureFinalization(message *Finalization, from NodeID, roun
...
+	msgsForRound, exists := e.futureMessages[string(from)][round]
+	if !exists {
+		e.Logger.Warn("Attempting to store a future finalization certificate for a round we have not seen yet",
+			zap.Uint64("round", round), zap.Stringer("from", from))
+		return
+	}
+	msgsForRound.finalizationCertificate = fCert
+}
+
+func (e *Epoch) futureProposalForRoundExists(round uint64) bool {
```
yacovm, line 558: so is this, no?

epoch.go
```diff
@@ -1786,6 +1810,11 @@ func (e *Epoch) maybeLoadFutureMessages() error {
 						return err
 					}
 				}
+				if msgs.finalizationCertificate != nil {
+					if err := e.handleFinalizationCertificateMessage(msgs.finalizationCertificate, NodeID(from)); err != nil {
```
yacovm, line 1814: I thought about something - when we save the finalization certificate in the future messages - will we always find it here and start replicating? What will prevent us from this happening? 

Is this something you plan on addressing later?

samliok: why would we begin replication? we only add the finalizationCertificate to future messages for the current round(if there is no proposal). Then when we process the proposal and later call `handleFinalizationCertificate` we will be handling it for the current round with a proposal, so there would be no need for replication.

yacovm: oh right right, it's for the current round only.

yacovm (approved):

PR #94, opened 2025-02-19
Try to collect an empty notarization after sending an empty vote

yacovm:
A node might be the last node among a quorum that times out. It therefore receives the empty votes before sending the empty vote itself, so it should check if it is possible to assemble an empty notarization right after sending the empty vote, because otherwise it might never get a trigger to check if an empty notarization can be assembled.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -1582,6 +1582,11 @@ func (e *Epoch) triggerProposalWaitTimeExpired(round uint64) {
 	emptyVotes.votes[string(e.ID)] = &signedEV
 
 	e.Comm.Broadcast(&Message{EmptyVoteMessage: &signedEV})
+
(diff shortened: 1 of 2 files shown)


epoch.go
```diff
@@ -1582,6 +1582,8 @@ func (e *Epoch) triggerProposalWaitTimeExpired(round uint64) {
 	emptyVotes.votes[string(e.ID)] = &signedEV
 
 	e.Comm.Broadcast(&Message{EmptyVoteMessage: &signedEV})
+
+	e.maybeAssembleEmptyNotarization()
```
samliok, line 1586: I think we should catch the error returned from `maybeAssembleEmptyNotarization` and propagate it up. Potentially we can re-use the `haltedError` variable?

samliok (approved):

PR #96, opened 2025-02-20
Error on double start

samliok:


yacovm:
But this is already [merged](https://github.com/ava-labs/Simplex/pull/95), no?

PR #98, opened 2025-02-21
Fine tune logs

yacovm:


epoch.go
```diff
@@ -1000,8 +1000,10 @@ func (e *Epoch) maybeCollectNotarization() error {
 	}
 
 	if voteCountForOurDigest < e.quorumSize {
-		e.Logger.Verbo("Counting votes for the digest we received from the leader",
-			zap.Uint64("round", e.round), zap.Int("votes", voteCount))
+		e.Logger.Warn("Counting votes for the digest we received from the leader",
```
samliok, line 1003: not sure this makes sense as a warning?

samliok (approved):

epoch.go
yacovm: This means the leader has equivocated or we have lying nodes that vote for digests that were never proposed, or we have a bug somewhere.
 
Because in line 980 we check if the total vote count is smaller than a quorum. 
If we passed the check and reaches this line (1003) it means we have a quorum of votes for this round, but not all of them are for the same block.

PR #99, opened 2025-02-22
Refuse finalizing a timed out round despite being notarized

yacovm:
This commit makes it that:
    
1) Nodes can accept a notarization of a block that they timed out on
2) Nodes will not vote on a block that they timed out on.
3) Nodes will not broadcast a finalization of a block that they accepted its notarization.

### DIFF
--- a/epoch_test.go
+++ b/epoch_test.go
@@ -763,9 +763,9 @@ func (t *testBlockBuilder) IncomingBlock(ctx context.Context) {
 }
 
 type testBlock struct {
-	data     []byte
-	metadata ProtocolMetadata
-	digest   [32]byte
+	data              []byte
+	metadata          ProtocolMetadata
+	digest            [32]byte
 	verificationDelay chan struct{}
 }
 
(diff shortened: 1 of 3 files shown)


epoch_test.go
```diff
@@ -858,6 +858,16 @@ func (mem *InMemStorage) waitForBlockCommit(seq uint64) Block {
 	}
 }
 
+func (mem *InMemStorage) ensureNoBlockCommit(t *testing.T, seq uint64) {
+	require.Never(t, func() bool {
```
samliok, line 862: interesting, haven't seen `Never` used before. For readability though, i think it would make sense to create the helper function and pass it in, rather than creating it as a parameter.

samliok (approved):

epoch_test.go
yacovm: Generally yes, but this is a utility method of the InMemStorage that is re-usable. 
If I extract the passed callback to a function, the method would be a one-liner. 
I'm not sure it's worth it, as the function extracted then will not be re-usable for anything other than the InMemStorage.

what do you think?

PR #100, opened 2025-02-24
Store future notarization message

yacovm:
We may receive a notarization message for a future round, and in that case - when we move to that round and finish verifying the proposal, we can just use it to advance to the next round.

### DIFF
--- a/epoch_test.go
+++ b/epoch_test.go
@@ -24,6 +24,82 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+func TestEpochHandleNotarizationFutureRound(t *testing.T) {
+	l := testutil.MakeLogger(t, 1)
+	bb := &testBlockBuilder{}
+	nodes := []NodeID{{1}, {2}, {3}, {4}}
+	// Create the two blocks ahead of time
+	blocks := createBlocks(t, nodes, bb, 2)
+	firstBlock := blocks[0].Block.(*testBlock)
+	secondBlock := blocks[1].Block.(*testBlock)
+	bb.out = make(chan *testBlock, 1)
+	bb.in = make(chan *testBlock, 1)
+
+	storage := newInMemStorage()
+
+	wal := newTestWAL(t)
+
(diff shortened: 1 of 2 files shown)


epoch_test.go
```diff
@@ -24,6 +24,85 @@ import (
...
+	nodes := []NodeID{{1}, {2}, {3}, {4}}
+
+	// Create the two blocks ahead of time
+	blocks := createBlocks(t, nodes, bb, 2)
+
+	firstBlock := blocks[0].Block.(*testBlock)
+	secondBlock := blocks[1].Block.(*testBlock)
+
+	// re-create the block builder
+	bb = &testBlockBuilder{out: make(chan *testBlock, 1), in: make(chan *testBlock, 1)}
```
samliok, line 40: would it be simpler to do something like this? 
```suggestion
	bb := &testBlockBuilder{}
	nodes := []NodeID{{1}, {2}, {3}, {4}}

	// Create the two blocks ahead of time
	blocks := createBlocks(t, nodes, bb, 2)

	firstBlock := blocks[0].Block.(*testBlock)
	secondBlock := blocks[1].Block.(*testBlock)

	bb.out = make(chan *testBlock, 1)
	bb.in = make(chan *testBlock, 1)
```

epoch_test.go
```diff
@@ -24,6 +24,85 @@ import (
...
+	require.NoError(t, err)
+
+	// The node should store the notarization of the second block once it gets the block.
+	wal.assertNotarization(1)
+
+	for i := 1; i < quorum; i++ {
+		injectTestFinalization(t, e, secondBlock, nodes[i])
+	}
+
+	storage.waitForBlockCommit(1)
```
samliok, line 103: might as well assert the `block` returned here is as expected

samliok (approved):

epoch_test.go
yacovm: Fixed

epoch_test.go
yacovm: Fixed

PR #104, opened 2025-02-26
 Add tests for nodes crashing right after a timeout

yacovm:


epoch_failover_test.go
```diff
@@ -117,9 +118,20 @@ func TestEpochLeaderFailoverWithEmptyNotarization(t *testing.T) {
 	nextBlockSeqToCommit := uint64(3)
 	nextRoundToCommit := uint64(4)
 
-	// Ensure our node proposes block with sequence 3 for round 4
-	notarizeAndFinalizeRound(t, nodes, nextRoundToCommit, nextBlockSeqToCommit, e, bb, quorum, storage, false)
-	require.Equal(t, uint64(4), storage.Height())
+	bbAfterCrash := &testBlockBuilder{
+		out:                make(chan *testBlock, 1),
+		in:                 make(chan *testBlock, 1),
+		blockShouldBeBuilt: make(chan struct{}, 1),
```
samliok, line 124: i think we can delete this since it unused in this test

epoch_failover_test.go
```diff
@@ -746,6 +769,68 @@ func TestEpochLeaderFailoverNotNeeded(t *testing.T) {
 	require.False(t, timedOut.Load())
 }
 
+func runCrashAndRestartExecution(t *testing.T, e *Epoch, nodes []NodeID, bb *testBlockBuilder, wal *testWAL, storage *InMemStorage, f func(t *testing.T, e *Epoch, bb *testBlockBuilder, storage *InMemStorage, wal *testWAL)) {
```
samliok, line 772: could we simplify the function signature? Potentially we can remove the `wal` and `storage` variables from the signature and just grab them from `e`? Also i think it would be nice to create a type alias for the function header `f`.

epoch_failover_test.go
```diff
@@ -746,6 +769,68 @@ func TestEpochLeaderFailoverNotNeeded(t *testing.T) {
...
+	// Split the test into two scenarios:
+	// 1) The node proceeds as usual.
+	// 2) The node crashes and restarts.
+	cloneWAL := wal.Clone()
+	cloneStorage := storage.Clone()
```
samliok, line 777: can we rename these variables to be slightly more descriptive? or even remove them completely and clone them in line

```go
	// Case 1:
	t.Run(fmt.Sprintf("%s-no-crash", t.Name()), func(t *testing.T) {
		f(t, e, bb, storage.clone(), wal.clone())
	})
```

epoch_failover_test.go
yacovm: I removed the `nodes` and grabbed them from `e`. 

I don't like too much the idea of getting `wal` and `storage` from `e` because that would need casting since `e` holds them as interfaces, and it's a bit ugly... but can do it if you feel strongly about it

epoch_failover_test.go
yacovm: oops, forgot to clean that

epoch.go
```diff
@@ -304,6 +320,24 @@ func (e *Epoch) resumeFromWal(records [][]byte) error {
 		lastMessage := Message{Notarization: &notarization}
 		e.Comm.Broadcast(&lastMessage)
 		return e.doNotarized(notarization.Vote.Round)
+	case record.EmptyVoteRecordType:
+		tbsEmptyVote, err := ParseEmptyVoteRecord(lastRecord)
```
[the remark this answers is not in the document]
yacovm: yep indeed.

epoch_failover_test.go
samliok: ah no its fine as is. i'm just not a fan of big function headers if we can avoid them without much hassle

PR #106, opened 2025-02-26
Remove Extra Check For Valid Vote

samliok:
stumbled upon this double check when handling notarization requests

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -630,6 +630,10 @@ func (e *Epoch) handleVoteMessage(message *Vote, from NodeID) error {
 		return nil
 	}
 
+	if !e.isVoteValid(vote) {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


epoch.go
```diff
@@ -1099,20 +1099,6 @@ func (e *Epoch) handleNotarizationMessage(message *Notarization, from NodeID) er
 	e.Logger.Verbo("Received notarization message",
 		zap.Stringer("from", from), zap.Uint64("round", vote.Round))
 
-	// Ignore votes for previous rounds
```
yacovm, line 1102: why are we deleting this?

samliok: because `e.isVoteValid` performs these checks

yacovm (approved):

PR #109, opened 2025-02-28
Check if round exists before accessing fields in metadata

samliok:
when the e.rounds map is populated but has no blocks with `notarizations` or `finalizations` the `metadata` function would panic on a `nil` dereference.


samliok:
I would prefer having the logic stay in `metadata` rather hackily populating `e.rounds` on init. The second one would still fail though because you do `currMed := e.getHighestRound().block.BlockHeader()` before checking if nil. 

Whats wrong with my approach? I think to make it better, i would keep the function signature of `e.getHighestRound()` the same, and check if the return value is nil.

yacovm (approved):

PR #110, opened 2025-02-28
Remove Last Block Runtime Variable

samliok:
I think if we should avoid unnecessary runtime variables that can get out of sync from their original value. Saw this small change when I found the bug on #109

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -74,7 +74,6 @@ type Epoch struct {
 	// Runtime
 	sched                          *oneTimeBlockScheduler
 	lock                           sync.Mutex
-	lastBlock                      Block // latest block commited
 	canReceiveMessages             atomic.Bool
 	finishCtx                      context.Context
 	finishFn                       context.CancelFunc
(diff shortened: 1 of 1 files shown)


epoch.go
```diff
@@ -1641,8 +1624,13 @@ func (e *Epoch) metadata() ProtocolMetadata {
 		seq = currMed.Seq + 1
 	}
 
-	if e.lastBlock != nil {
-		currMed := e.lastBlock.BlockHeader()
+	lastBlockFromStorage, _, err := RetrieveLastIndexFromStorage(e.Storage)
+	if err != nil {
+		e.Logger.Error("Error retrieving block from storage", zap.Uint64("seq", e.Storage.Height() -1))
```
samliok, line 1629: should we `panic` here, or return the error?

epoch.go
```diff
@@ -1641,8 +1624,13 @@ func (e *Epoch) metadata() ProtocolMetadata {
 		seq = currMed.Seq + 1
 	}
 
-	if e.lastBlock != nil {
-		currMed := e.lastBlock.BlockHeader()
+	lastBlockFromStorage, _, err := RetrieveLastIndexFromStorage(e.Storage)
```
yacovm, line 1627: I don't think we should be doing I/O every time we decide which metadata to use when we build a block.

samliok: fair, although i thought it would be fairly fast since the last index would most likely be in our cache.

samliok: going to close this pr

PR #112, opened 2025-02-28
Delete Old Runtime Map Key/Values

samliok:
Rather than deleting from `emptyVotes` immediately after persisting to the wal, we delete once the round has advanced enough. This is needed for replication, since nodes need to be able to retrieve `emptyNotarziation` when requested.  

Tthis PR also adds the same logic to the `rounds` map, as before `rounds` could have grown indefinitely.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -250,7 +250,7 @@ func (e *Epoch) restoreEmptyVoteRecord(r []byte) error {
 	emptyVote := &EmptyVote{
 		Signature: Signature{
 			Signer: e.ID,
-			Value: signature,
+			Value:  signature,
 		},
 		Vote: vote,
 	}
@@ -326,11 +326,11 @@ func (e *Epoch) resumeFromWal(records [][]byte) error {
 			return err
 		}
 		round, exists := e.emptyVotes[ev.Round]
-		if ! exists {
+		if !exists {
 			return fmt.Errorf("round %d not found for empty vote", ev.Round)
 		}
 		emptyVote, exists := round.votes[string(e.ID)]
-		if ! exists {
+		if !exists {
 			return fmt.Errorf("could not find my own vote for round %d", ev.Round)
 		}
 		lastMessage := Message{EmptyVoteMessage: emptyVote}
@@ -1812,7 +1812,7 @@ func (e *Epoch) voteOnBlock(block Block) (Vote, error) {
 
 func (e *Epoch) deleteRoundsTooFarInThePast() {
 	for i, round := range e.rounds {
-		if round.block.BlockHeader().Round + e.maxRoundWindow < e.round {
+		if round.block.BlockHeader().Round+e.maxRoundWindow < e.round {
 			if round.fCert != nil {
 				delete(e.rounds, i)
 			}
@@ -1824,7 +1824,7 @@ func (e *Epoch) deleteEmptyVoteForPreviousRound() {
 	if e.round == 0 {
 		return
 	}
-	delete(e.emptyVotes, e.round - 1)
+	delete(e.emptyVotes, e.round-1)
 }
 
 func (e *Epoch) increaseRound() {
--- a/epoch.go
+++ b/epoch.go
@@ -1820,14 +1820,11 @@ func (e *Epoch) deleteRoundsTooFarInThePast() {
 	}
 }
 
-func (e *Epoch) deleteEmptyVotesTooFarInThePast() {
-	for i, emptyVote := range e.emptyVotes {
-		if i + e.maxRoundWindow < e.round {
-			if emptyVote.emptyNotarization != nil {
-				delete(e.emptyVotes, i)
-			}
-		}
+func (e *Epoch) deleteEmptyVoteForPreviousRound() {
+	if e.round == 0 {
+		return
 	}
+	delete(e.emptyVotes, e.round - 1)
 }
 
 func (e *Epoch) increaseRound() {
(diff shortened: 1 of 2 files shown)


epoch.go
```diff
@@ -1816,11 +1810,34 @@ func (e *Epoch) voteOnBlock(block Block) (Vote, error) {
...
+	}
+}
+
 func (e *Epoch) increaseRound() {
 	// In case we're waiting for a block to be notarized, cancel the wait because
 	// we advanced to the next round.
 	e.cancelWaitForBlockNotarization()
 
+	e.deleteRoundsTooFarInThePast()
+	e.deleteEmptyVotesTooFarInThePast()
```
samliok, line 1839: i think it's a bit inefficient to iterate over all the key/values of both maps. Maybe it's worth using/creating a sorted map data structure?

samliok: it would be nice to know the lowest key in the map, so we can just iterate over [lowestValue, e.round-e.maxRoundWindow]

yacovm: > i think it's a bit inefficient to iterate over all the key/values of both maps. Maybe it's worth using/creating a sorted map data structure?

I think we should benchmark how much time this takes before we invest into making this sorted.

epoch.go
```diff
@@ -1816,11 +1810,34 @@ func (e *Epoch) voteOnBlock(block Block) (Vote, error) {
...
+	for i, round := range e.rounds {
+		if round.block.BlockHeader().Round + e.maxRoundWindow < e.round {
+			if round.fCert != nil {
+				delete(e.rounds, i)
+			}
+		}
+	}
+}
+
+func (e *Epoch) deleteEmptyVotesTooFarInThePast() {
```
yacovm, line 1823: I think for this we can delete once we advance to the next round, no?

epoch.go
```diff
@@ -989,8 +985,6 @@ func (e *Epoch) persistEmptyNotarization(emptyNotarization *EmptyNotarization, s
 		zap.Int("size", len(emptyNotarizationRecord)),
 		zap.Uint64("round", emptyNotarization.Vote.Round))
 
-	delete(e.emptyVotes, e.round)
```
yacovm, line 992: instead of removing it here, I think we can just delete the empty vote of previous round once we advance to a next round. 

We should keep the empty vote of the current round in memory because that will let you replicate it to other nodes. 

Thoughts?

samliok: so for replication I actually need to keep more than just the current round. if a node is `5` rounds behind and there have been 5 empty notarizations they should replicate all 5.

yacovm: why do you need to replicate all 5? we have the sequence which proves that all other 4 are empty notarizations, no?

epoch.go
```diff
@@ -1816,11 +1810,34 @@ func (e *Epoch) voteOnBlock(block Block) (Vote, error) {
...
+		if round.block.BlockHeader().Round + e.maxRoundWindow < e.round {
+			if round.fCert != nil {
+				delete(e.rounds, i)
+			}
+		}
+	}
+}
+
+func (e *Epoch) deleteEmptyVotesTooFarInThePast() {
+	for i, emptyVote := range e.emptyVotes {
```
yacovm: I'm not sure we need to iterate over the rounds. Why not pass the round number we have advanced to, and just delete the previous round?

```
func (e *Epoch) deleteEmptyVotesTooFarInThePast(round uint64) {
    delete(e.emptyVotes, round - 1)
}
```

Then call it via:

```
func (e *Epoch) increaseRound() {
        ...
	e.deleteEmptyVotesTooFarInThePast(e.round)
        ...
```

If we wish to keep `K` old empty votes we can just `delete(e.emptyVotes, round - K)` in `deleteEmptyVotesTooFarInThePast`. 

Since we only advance rounds via going through `increaseRound`, for every round `r` that has empty votes associated to it, there exists a round `r+K` that once we get to it, we will delete the empty votes of the round `r`.

epoch.go
```diff
@@ -1816,11 +1810,34 @@ func (e *Epoch) voteOnBlock(block Block) (Vote, error) {
...
+		}
+	}
+}
+
 func (e *Epoch) increaseRound() {
 	// In case we're waiting for a block to be notarized, cancel the wait because
 	// we advanced to the next round.
 	e.cancelWaitForBlockNotarization()
 
+	e.deleteRoundsTooFarInThePast()
```
yacovm, line 1839: I don't think we should garbage collect rounds from within `increaseRound`. 

In `deleteRoundsTooFarInThePast`, we iterate over all rounds and if it's too far in the past, we delete it if it has an fCert:

```
		if round.block.BlockHeader().Round + e.maxRoundWindow < e.round {
			if round.fCert != nil {
				delete(e.rounds, i)
			}
		}
```

However, if from some reason we don't finalize blocks but only notarize them, we may delete a round that hasn't been indexed into storage. We have rounds that have fCerts but are waiting for previous rounds to be indexed before they can be indexed, and this operation may delete them prematurely. 

I think that what we need is to only allow deletion of rounds older than the specific round that we index in storage:

```
func (e *Epoch) deleteRoundsTooFarInThePast(round uint64) {
	for i, round := range e.rounds {
		if round.block.BlockHeader().Round + e.maxRoundWindow < round {
                       delete(e.rounds, i)
		}
	}
}
```

and invoke it from the `indexFinalizationCertificates` as previously.

epoch.go
samliok: done. and i don't think we need to keep more than one previous `emptyVote` as mentioned in your comment above

yacovm (approved):

PR #114, opened 2025-02-28
Distinguish between verified and unverified blocks

yacovm:
When we receive a block from the network, it should be verified before it can be persisted into stable storage.

Similarly, when we load a block from stable storage, there is no more need to verify it, as we had verified it before.

This commit makes the compiler enforce that a block received from the network cannot be persisted into stable storage before it has been verified.

### DIFF
--- a/epoch_multinode_test.go
+++ b/epoch_multinode_test.go
@@ -47,7 +47,7 @@ func newSimplexNodeWithStorage(t *testing.T, nodeID NodeID, net *inMemNetwork, b
 	wal := newTestWAL(t)
 	conf := defaultTestNodeEpochConfig(t, nodeID, net, wal, bb, true)
 	for _, data := range storage {
-		conf.Storage.Index(data.Block, data.FCert)
+		conf.Storage.Index(data.VerifiedBlock, data.FCert)
 	}
 	e, err := NewEpoch(conf)
 	require.NoError(t, err)
@@ -264,6 +264,8 @@ func (c *testComm) SendMessage(msg *Message, destination NodeID) {
 		return
 	}
 
+	c.maybeTranslateOutoingToIncomingMessageTypes(msg)
+
 	for _, instance := range c.net.instances {
 		if bytes.Equal(instance.e.ID, destination) {
 			instance.ingress <- struct {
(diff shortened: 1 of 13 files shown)


epoch_multinode_test.go
```diff
@@ -264,6 +264,17 @@ func (c *testComm) SendMessage(msg *Message, destination NodeID) {
 		return
 	}
 
+	if msg.ReplicationResponse != nil {
```
yacovm, line 267: we convert the verified blocks into regular ones, so we can inject it inside the remote node's instance.

epoch_multinode_test.go
```diff
@@ -280,6 +291,10 @@ func (c *testComm) Broadcast(msg *Message) {
 		return
 	}
 
+	if msg.BlockMessage != nil {
```
yacovm, line 294: we convert the verified block to an unverified one so we could inject it as input to another instance.

epoch_multinode_test.go
samliok: hmmm, it feels a little wrong to do the type conversion in send message. Also when would `FinalizedBlock` have a have a just a `Block`? Would it make sense to remove the `Block` field from FinalizedBlock?

msg.go
```diff
@@ -184,8 +184,9 @@ func (n *Notarization) Verify() error {
 }
 
 type BlockMessage struct {
-	Block Block
-	Vote  Vote
+	Block         Block
```
samliok, line 187: why do we accept both a `Block` and `VerifiedBlock` in the block message? Where do we use verified block?

yacovm: When we receive a block as a non-leader, it should be a `Block`. When we send a block (as the leader) it should be a `VerifiedBlock`.

epoch_multinode_test.go
yacovm: We do it because it's a test. Since we only deal with message instances that live in memory in both ends of the sender and receiver, we don't have much of a choice, as the receiver node may only receive `Block`s and the sender can only send `VerifiedBlocks`.  
In production this won't be an issue, as avalanchego would take care of sending both types down the wire.

> Also when would FinalizedBlock have a have a just a Block? Would it make sense to remove the Block field from FinalizedBlock?

When a node is a sender, it sends a `VerifiedBlock`. When a node is a receive, it receives a `Block`.

samliok (approved):

PR #118, opened 2025-03-04
Make sure QC only contains authorized signers and is not signed twice

yacovm:
This commit makes sure that a QC only contains signers that are among the nodes the Epoch was initialized with. Similarly, it checks the QC is not signed twice or not signed by too few nodes.

This eases the implementation of a QC by the application, especially for naive implementations where the QC contains the public keys of the signers, and the Signers() method just returns a deterministic derivation of the aforementioned public keys.

### DIFF
--- a/util.go
+++ b/util.go
@@ -24,8 +24,8 @@ func RetrieveLastIndexFromStorage(s Storage) (Block, *FinalizationCertificate, e
 	return lastBlock, &fCert, nil
 }
 
-func IsFinalizationCertificateValid(fCert *FinalizationCertificate, quorumSize int, logger Logger) bool {
-	valid := validateFinalizationQC(fCert, quorumSize, logger)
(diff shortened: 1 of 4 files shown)


util.go
```diff
@@ -36,7 +36,7 @@ func IsFinalizationCertificateValid(fCert *FinalizationCertificate, quorumSize i
 	return true
 }
 
-func validateFinalizationQC(fCert *FinalizationCertificate, quorumSize int, logger Logger) bool {
+func validateFinalizationQC(eligibleSigners map[string]struct{}, fCert *FinalizationCertificate, quorumSize int, logger Logger) bool {
```
samliok, line 39: should we remove `quorumSize` and use our helper function Quorum(len(eligibleSigners)) instead?

samliok (approved):

util.go
yacovm: why? the quorum size cannot change within an epoch. It makes sense to compute it once and just reference it each time.

Issue #82, opened 2025-02-14
Distribute requests across all nodes


samliok:
Rather than sending requests to one node, distribute the load across the network. Also manage requests/responses that are unanswered.

samliok:
we can even improve on by sending requests from @yacovm's suggestion

> would it make sense (as a follow-up PR) to request the sequences from the nodes that helped construct the fCert? We can get the list from fCert.QC.Signers().

PR #123, opened 2025-03-12
Replication of Future Sequences & Rounds

samliok:
This PR updates the code for replication. Rather than having two separate requests for notarizations and finalizations, this consolidates them into a single `ReplicationRequest` struct. 

```go
type  ReplicationRequest  struct {
	Seqs []uint64  // sequences we are requesting
	LatestRound  uint64  // latest round that we are aware of
}

type  ReplicationResponse  struct {
	Data []QuorumRound
	LatestRound  *QuorumRound
}

type  QuorumRound  struct {
	Block  Block
	Notarization  *Notarization
	FCert  *FinalizationCertificate
	EmptyNotarization  *EmptyNotarization
}
```

The `LatestRound` field notifies the responding node, whether the requesting node is behind. This way it knows whether to send its most recent `QuorumRound`. Receiving a higher latest round, tells the requesting node that it is still behind and may need to send out more replication requests.


epoch.go
```diff
@@ -2047,119 +2188,246 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+			e.Logger.Debug("Received quorum round for a seq that is too far behind", zap.Uint64("seq", data.GetSequence()))
 			continue
 		}
 
-		valid := IsFinalizationCertificateValid(e.eligibleNodeIDs, &data.FCert, e.quorumSize, e.Logger)
-		// verify the finalization certificate
-		if !valid {
-			e.Logger.Debug("Received invalid finalization certificate", zap.Uint64("seq", data.FCert.Finalization.Seq), zap.String("from", from.String()))
+		// we pass the sequence rather than round, since we can have large gaps due to empty notarizations
+		if e.isRoundTooFarAhead(data.GetSequence()) {
```
yacovm, line 2281: What if our round is 1000000 but the sequence is 100? This call to `isRoundTooFarAhead` will return false while we might be at sequence 1000 and cannot process such a far sequence.

epoch.go
```diff
@@ -2047,119 +2188,246 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+	}
+
+	// make sure the latest round is well formed
+	if err := latestRound.isWellFormed(); err != nil {
+		e.Logger.Debug("Received invalid latest round", zap.Error(err))
+		return err
+	}
+
+	if latestRound.GetRound() > e.round {
+		e.replicationState.collectMissingSequences(latestRound.GetSequence(), e.round, e.Storage.Height())
```
yacovm, line 2324: I think we shouldn't send requests for more data until we fully finish processing what we have gotten via `processReplicationState`. 

So it might make sense to call `processLatestRoundReceived` later after `processReplicationState`.

yacovm (commented):
Did a first pass, will make another pass tomorrow.

epoch.go
```diff
@@ -2047,119 +2188,246 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+
+	if err := e.processLatestRoundReceived(resp.LatestRound); err != nil {
+		e.Logger.Debug("Failed processing latest round", zap.Error(err))
+		return nil
 	}
 
 	return e.processReplicationState()
 }
 
+func (e *Epoch) processLatestRoundReceived(latestRound *QuorumRound) error {
```
yacovm, line 2312: I might be wrong but we don't verify this latest round, and we should....

epoch.go
```diff
@@ -2172,6 +2440,61 @@ func (e *Epoch) getHighestRound() *Round {
...
+
+	eNoteRound, emptyNotarization := e.getHighestEmptyNotarization()
+	if emptyNotarization != nil && (eNoteRound > highestRound || !exists) {
+		verifiedQuorumRound = &VerifiedQuorumRound{
+			EmptyNotarization: emptyNotarization,
+		}
+		exists = true
+	}
+
+	lastBlock, fCert, err := RetrieveLastIndexFromStorage(e.Storage)
```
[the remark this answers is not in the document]
samliok: ah, we need to use the last block field because the finalization certificate is stored in storage, should the last finalization certificate also be stored in memory?

epoch.go
```diff
@@ -2047,119 +2188,246 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+				FCert:         round.fCert,
+			}
+		}
+	}
 
-			return round.block, *round.fCert, true
+	// check empty votes
+	emptyVotes, exists := e.emptyVotes[e.round]
+	if exists && emptyVotes.emptyNotarization != nil {
+		if emptyVotes.emptyNotarization.Vote.Round == seq {
```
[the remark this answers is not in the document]
samliok: oops, yea i think we can remove this altogether since we will only send back an empty notarization in the `LastRound` field

epoch.go
samliok: yep, we should be comparing it to `nextSeqToCommit`

epoch.go
samliok: i agree, i added a check in `StoreQuorumRound(latestRound)` 
```
if round.GetSequence() > r.highestSeqReceived {
		r.highestSeqReceived = round.GetSequence()
}
```
this way we can request new sequences using the normal path of `maybeCollectFutureSequences`

msg.go
```diff
@@ -221,33 +223,115 @@ type QuorumCertificate interface {
...
+	EmptyNotarization *EmptyNotarization
 }
 
-type VerifiedFinalizedBlock struct {
-	VerifiedBlock VerifiedBlock
-	FCert         FinalizationCertificate
+// isWellFormed returns an error if the QuorumRound has either
+// (block, notarization) or (block, finalization certificate) or
+// (empty notarization)
+func (q QuorumRound) isWellFormed() error {
```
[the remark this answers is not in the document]
samliok: fixed and added tests

msg.go
```diff
@@ -221,33 +223,115 @@ type QuorumCertificate interface {
...
+	}
+
+	return fmt.Errorf("no finalization certificate, empty notarization or notarization found")
+}
+
+// String returns a string representation of the QuorumRound.
+// It is meant as a debugging aid for logs.
+func (q *QuorumRound) String() string {
+	if q != nil {
+		err := q.isWellFormed()
```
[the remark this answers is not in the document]
samliok: yep because a nil dereference with `GetRound` & `GetSequence`. Should i add the `isWellFormed` check to these methods? In the case that it is _not_ well formed should we return 0? I'm assuming we wouldn't want to panic

epoch.go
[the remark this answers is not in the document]
yacovm: I think so... can we unify them in a single variable though, to be consistent?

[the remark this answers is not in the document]
samliok: done

epoch.go
```diff
@@ -2172,6 +2440,61 @@ func (e *Epoch) getHighestRound() *Round {
...
+		}
+	}
+
+	return max, emptyNotarization
+}
+
+// getLatestVerifiedQuorumRound returns the latest verified quorum round.
+// it returns either an empty notarization, or a block
+// with either a notarization or finalization certificate.
+func (e *Epoch) getLatestVerifiedQuorumRound() (*VerifiedQuorumRound, error) {
```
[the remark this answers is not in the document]
samliok: done

epoch.go
```diff
@@ -2047,119 +2187,244 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+	return nil
+}
+
 func (e *Epoch) processReplicationState() error {
 	nextSeqToCommit := e.Storage.Height()
 
 	// check if we are done replicating and should start a new round
-	if e.replicationState.isReplicationComplete(nextSeqToCommit) {
+	if e.replicationState.isReplicationComplete(nextSeqToCommit, e.round) {
+		// TODO: an adversarial node can send multiple empty replication responses, causing us
```
yacovm, line 2331: We can self assign the replicationEnabled to be false by default, and only when we send out a replication request we can then enable it, and we can then turn if off again at startRound.

epoch.go
```diff
@@ -2047,119 +2187,244 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+		// to call start round multiple times. This is potentially bad if we are the leader, since we will
+		// propose multiple blocks for the same round.
 		return e.startRound()
 	}
 
-	finalizedBlock, ok := e.replicationState.receivedFinalizationCertificates[nextSeqToCommit]
-	if !ok {
-		return nil
+	e.replicationState.maybeCollectFutureSequences(e.Storage.Height())
```
yacovm, line 2337: shouldn't we move this `e.replicationState.maybeCollectFutureSequences(e.Storage.Height())` to be after everything else in this function? `e.Storage.Height()` will be more up to date then, no?

epoch.go
```diff
@@ -2047,119 +2187,244 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+	block, fCert, exists := e.replicationState.GetFinalizedBlockForSequence(nextSeqToCommit)
+	if exists {
+		delete(e.replicationState.receivedQuorumRounds, block.BlockHeader().Round)
+		return e.processFinalizedBlock(block, fCert)
 	}
 
-	delete(e.replicationState.receivedFinalizationCertificates, nextSeqToCommit)
-	e.replicationState.maybeCollectFutureFinalizationCertificates(e.round, e.Storage.Height())
-	return e.processFinalizedBlock(&finalizedBlock)
+	// TODO: for this pr include a helper function to allow the node to deduce whether
```
yacovm, line 2346: TODO for me - better understand this piece in the next review round

epoch.go
```diff
@@ -2047,119 +2187,244 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+	// empty notarization will be sent by the request
+	roundAdvanced, err := e.maybeAdvanceRoundFromEmptyNotarizations()
+	if err != nil {
+		return err
+	}
+	if roundAdvanced {
+		return e.processReplicationState()
+	}
+
+	// TODO: we need to make sure that we do not forget about notarizations missing for rounds < e.round
```
yacovm, line 2357: I'm also a bit puzzled regarding why have `e.round` here? Are we assuming that `processReplicationState` gave us sequences that advanced the round to the point where we might be right before an empty notarization?

epoch_multinode_test.go
```diff
@@ -242,15 +248,52 @@ func (tw *testWAL) containsEmptyVote(round uint64) bool {
...
+// denyFinalizationMessages blocks any messages that would cause nodes in
+// a network to index a block in storage.
+func denyFinalizationMessages(msg *Message, destination NodeID) bool {
+	if msg.Finalization != nil {
+		return false
+	}
+	if msg.FinalizationCertificate != nil {
+		return false
+	}
+	if msg.Notarization != nil {
```
yacovm, line 269: why do we also need to block notarizations? isn't it enough to block the above two?

epoch_multinode_test.go
```diff
@@ -259,6 +302,12 @@ func (c *testComm) ListNodes() []NodeID {
 }
 
 func (c *testComm) SendMessage(msg *Message, destination NodeID) {
+	c.lock.RLock()
```
yacovm, line 305: I assume the lock is needed for the message filter. 

Is it possible to just lock around it? and not use a defer.

We have a channel send below the function and I generally don't like sending into channels while holding a lock because the thread that reads from the channel might need the lock to progress. Not speaking about this current code but in general, no reason to hold an RLock for long if there is no need for it.

epoch_multinode_test.go
```diff
@@ -316,13 +394,19 @@ func (c *testComm) maybeTranslateOutoingToIncomingMessageTypes(msg *Message) {
 }
 
 func (c *testComm) Broadcast(msg *Message) {
+	c.lock.RLock()
```
yacovm, line 397: I think it makes sense to just make a function `isMessagePermitted` that RLocks around the message filter and not explicitly RLock everywhere we use it.

yacovm (commented):
Made another pass, will make another final pass later.

msg.go
```diff
@@ -221,33 +223,133 @@ type QuorumCertificate interface {
...
+	}
+
+	if q.Notarization != nil {
+		if !bytes.Equal(blockDigest[:], q.Notarization.Vote.Digest[:]) {
+			return fmt.Errorf("notarization does not match the block")
+		}
+		return q.Notarization.Verify()
+	}
+
+	return nil
```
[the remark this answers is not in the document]
samliok: i dont think we need this. `IsWellFormed` would have returned early if so.

also one of your previous comments suggested that i shouldn't return early after verifying the `fCert`. i think it makes more sense to go with what you said before, and verify both the `fCert` and `notarization` if necessary
https://github.com/ava-labs/Simplex/pull/123#discussion_r1994316006

epoch.go
samliok: hmm, i think reusing `replicationEnabled` was initially intended to toggle replication as a whole. If we change the functionality of `replicationEnabled` to be adjustable we wouldn't be able to fully disable replication. 

At the same time, the `replicationEnabled` flag was supposed to be temporary, so we can definitely repurpose it here. Maybe in a separate pr though?

epoch.go
samliok: due to the recursive nature of the function, putting it at the end will mean we send the requests only after process all the received `QuorumRounds`. 

For example, if we are replicating blocks 0-20, we would first receive 0-10. Currently we would send another request replicating 10-15 when `nextSeqToCommit` is 5. However if we `maybeCollect` at the end, then we would sent the request once `nextSeqToCommit` reaches 10 for rounds 10-20.

epoch.go
samliok: i think this is an old comment. The point of it was to double check if we would properly replicate notarizations that we receive that are less than our current round. However, i don't think we can ever receive a notarization that is less than `e.round` since we only advance `e.round` after we notarize or finalize.

epoch.go
yacovm: yeah we should tackle this in a separate PR.

yacovm: This was related to [this](https://github.com/ava-labs/Simplex/pull/123#discussion_r1996275741) comment.

epoch.go
```diff
@@ -1380,18 +1397,18 @@ func (e *Epoch) handleBlockMessage(message *BlockMessage, from NodeID) error {
...
 // otherwise it will verify the block first.
-func (e *Epoch) processFinalizedBlock(finalizedBlock *FinalizedBlock) error {
-	round, exists := e.rounds[finalizedBlock.FCert.Finalization.Round]
+func (e *Epoch) processFinalizedBlock(block Block, fCert FinalizationCertificate) error {
+	round, exists := e.rounds[fCert.Finalization.Round]
 	// dont create a block verification task if the block is already in the rounds map
 	if exists {
 		roundDigest := round.block.BlockHeader().Digest
-		seqDigest := finalizedBlock.FCert.Finalization.BlockHeader.Digest
+		seqDigest := fCert.Finalization.BlockHeader.Digest
```
yacovm, line 1405: I think I missed this before: 

```
if !bytes.Equal(roundDigest[:], seqDigest[:]) {
			e.Logger.Warn("Received finalized block that is different from the one we have in the rounds map",
				zap.Stringer("roundDigest", roundDigest), zap.Stringer("seqDigest", seqDigest))
			return nil
		}
```

But this shouldn't happen. we should return an error here instead, and mark the halted error.

epoch.go
```diff
@@ -2099,119 +2240,244 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+		}
+		delete(e.replicationState.receivedQuorumRounds, qRound.GetRound())
+		return true, nil
+	}
+
+	round++
+	numEmptyNotarizedRounds := 1
+
+	// its possible we did not receive this round since it was an empty notarization round
+	// check to see if we a future round with the sequence that we expect
```
yacovm, line 2453: > // check to see if we a future round with the sequence that we expect

Can we fix this comment?

replication_test.go
```diff
@@ -118,33 +73,168 @@ func TestReplication(t *testing.T) {
...
+		comm := newTestComm(from, net, denyFinalizationMessages)
+		return &testNodeConfig{
+			comm:               comm,
+			replicationEnabled: true,
+		}
+	}
+
+	normalNode1 := newSimplexNode(t, nodes[0], net, bb, newNodeConfig(nodes[0]))
+	normalNode2 := newSimplexNode(t, nodes[1], net, bb, newNodeConfig(nodes[1]))
+	newSimplexNode(t, nodes[2], net, bb, newNodeConfig(nodes[2]))
```
yacovm, line 176: why is the third instance not given a variable?

replication_test.go
```diff
@@ -118,33 +73,168 @@ func TestReplication(t *testing.T) {
...
+	net.setAllNodesMessageFilter(onlyAllowEmptyRoundMessages)
+
+	// normal nodes continue to make progress
+	for i := uint64(1); i < uint64(numNotarizations); i++ {
+		bb.triggerNewBlock()
+		emptyNotarization := newEmptyNotarization(nodes[:3], i, 1)
+		msg := &simplex.Message{
+			EmptyNotarization: emptyNotarization,
+		}
+		normalNode2.e.Comm.SendMessage(msg, normalNode1.e.ID)
```
yacovm, line 203: is it not possible to make the nodes send the messages from their own will, as done in the failover tests via `waitForBlockProposerTimeout` or something?

replication_test.go
```diff
@@ -118,33 +73,168 @@ func TestReplication(t *testing.T) {
...
+		require.Equal(t, uint64(1), n.e.Metadata().Seq)
+		require.Equal(t, uint64(0), n.e.Storage.Height())
+	}
+
+	net.setAllNodesMessageFilter(allowAllMessages)
+	net.Connect(nodes[3])
+
+	fCert, _ := newFinalizationRecord(t, laggingNode.e.Logger, laggingNode.e.SignatureAggregator, block, nodes)
+
+	// we broadcast from the second node so that node 1 will be able to respond
```
yacovm, line 223: The point of network tests is to simulate a real scenario of how nodes behave under certain network conditions of crashes and message omissions.

Forcing them to send messages as we see fit isn't ideal for testing a real scenario.

yacovm:
production code LGTM, will make another pass on the tests tomorrow.

replication_test.go
samliok: there's no need to give it a variable name, it would be unused

replication_test.go
samliok: yes. Do you mind if I also do this in a follow up? I think it would be a more invasive change to the test, since i would need to change how `advanceRound` works as well as the number of nodes in the network for this test.

I did some changes locally and got the test passing with a flake, i'd like to look at and create a new PR with a clean diff.

replication_test.go
samliok: i agree this isn't ideal, but our current code for replication is hardcoded to be sent from the first `node`.  I think this test should be updated in the same pr as #82.

epoch.go
samliok: done

replication_test.go
```diff
@@ -10,95 +10,50 @@ import (
...
+// TestReplication tests the replication process of a node that
+// is behind the rest of the network by less than maxRoundWindow.
+func TestBasicReplication(t *testing.T) {
+	nodes := []simplex.NodeID{{1}, {2}, {3}, []byte("lagging")}
 
-func TestNilFinalizationCertificateResponse(t *testing.T) {
-	bb := newTestControlledBlockBuilder(t)
-	nodes := []simplex.NodeID{{1}, {2}, {3}, {4}}
-	net := newInMemNetwork(t, nodes)
+	// wihtih max round window for now
```
yacovm, line 24: within?

yacovm (approved):

PR #126, opened 2025-03-18
Replicate Empty Notarizations

samliok:
This PR updates the `TestReplicationEmptyNotarizations` to send messages by themselves. It also simplifies the logic for handling emptyNotarizations.


epoch.go
```diff
@@ -2411,65 +2433,39 @@ func (e *Epoch) processReplicationState() error {
...
 //
 //	NBlock1 { round 1, seq 1 }
-//	NBlock2 { round 3, seq 2 }
+//	NBlock2 { round 8, seq 1 }
 //
-// in this case we can infer there was an empty notarization for round 2.
+// in this case we can infer there was 8-1 empty notarizations during rounds [2, 8].
 func (e *Epoch) maybeAdvanceRoundFromEmptyNotarizations() (bool, error) {
 	round := e.round
 	expectedSeq := e.metadata().Seq
```
yacovm, line 2450: why do we use `expectedSeq := e.metadata().Seq` and not just `e.Storage.Height()` ?

epoch.go
```diff
@@ -2411,65 +2433,39 @@ func (e *Epoch) processReplicationState() error {
...
-			// we have an entry
-			for range numEmptyNotarizedRounds {
-				e.increaseRound()
-			}
-			return true, nil
-		} else {
-			// if we have an entry from a previous round but its not the expected sequence,
-			// we may be missing some notarizations for previous rounds
-			return false, nil
+	// if there is no sequence, then maybe there is one with the same sequence but an empty notarization
```
yacovm, line 2461: I don't understand this case. Can't `expectedSeq - 1` be our current latest sequence? What is the meaning of it?

replication.go
```diff
@@ -138,44 +138,22 @@ func (r *ReplicationState) StoreQuorumRound(round QuorumRound) {
...
+				continue
 			}
-
 			return round.Block, *round.FCert, true
 		}
 	}
 	return nil, FinalizationCertificate{}, false
 }
 
-type NotarizedBlock struct {
```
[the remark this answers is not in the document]
samliok: we don't use `NotarizedBlock` often, and when we did we could have just passed the two separate components. Moreso a cleanup & style thing

epoch_failover_test.go
```diff
@@ -668,17 +666,16 @@ func createEmptyVote(md ProtocolMetadata, signer NodeID) *EmptyVote {
...
 
-func waitForBlockProposerTimeout(t *testing.T, e *Epoch, startTime time.Time) {
-	startRound := e.Metadata().Round
+func waitForBlockProposerTimeout(t *testing.T, e *Epoch, startTime *time.Time, startRound uint64) {
 	timeout := time.NewTimer(time.Minute)
 	defer timeout.Stop()
 
 	for {
-		if e.WAL.(*testWAL).containsEmptyVote(startRound) {
+		if e.WAL.(*testWAL).containsEmptyVote(startRound) || e.WAL.(*testWAL).containsEmptyNotarization(startRound) {
```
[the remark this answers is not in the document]
samliok: when we call `waitForBlockProposerTimeout`, it's possible that our node has received an empty notarization before the `waitForBlockNotification` has been triggered. When we process an empty notarization we `cancelWaitForBlockNotarization`, meaning we will never send an empty vote for the round.

epoch.go
samliok: exactly. if we have a sequence that is our current latest sequence and that quorum round has an empty notarization then we should advance the rounds

epoch.go
samliok: we want to use the `metadata` seq because that takes into account the seq of notarized blocks

replication_test.go
```diff
@@ -171,67 +190,73 @@ func TestReplicationEmptyNotarizations(t *testing.T) {
...
 	net.setAllNodesMessageFilter(allowAllMessages)
-	net.Connect(nodes[3])
+	net.Connect(laggingNode.e.ID)
 
 	fCert, _ := newFinalizationRecord(t, laggingNode.e.Logger, laggingNode.e.SignatureAggregator, block, nodes)
-
 	// we broadcast from the second node so that node 1 will be able to respond
 	// to the lagging nodes request
-	normalNode2.e.Comm.Broadcast(&simplex.Message{
+	normalNode2.e.Comm.SendMessage(&simplex.Message{
```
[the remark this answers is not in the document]
samliok: this is to initiate the replication requests, which is for a different pr. this pr addresses

https://github.com/ava-labs/Simplex/pull/123#discussion_r1999849734
https://github.com/ava-labs/Simplex/pull/123#discussion_r1999689462

epoch.go
```diff
@@ -2393,16 +2404,27 @@ func (e *Epoch) processReplicationState() error {
 
 	e.replicationState.maybeCollectFutureSequences(e.Storage.Height())
 
-	// first we check if we can commit the next sequence
+	// first we check if we can commit the next sequence, it is ok to try and commit the next sequence
```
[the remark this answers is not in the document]
samliok: reworded this comment. The initial meaning was to mention we don't need to look at the round if we have a fCert.

epoch.go
yacovm: but shouldn't [this](https://github.com/ava-labs/Simplex/blob/fa3e4da75dc274380d10dc56e7a6d2cfdace2239/epoch.go#L2423) take care of it? 

```
	// the current round is an empty notarization
	if ok && qRound.EmptyNotarization != nil {
		delete(e.replicationState.receivedQuorumRounds, qRound.GetRound())
		return e.processEmptyNotarization(qRound.EmptyNotarization)
	}
```

samliok: no, we only enter this condition if their is an empty notarization for the current round(e.round). The about checks for empty notarizations for the current sequence past e.round.

yacovm (approved):

PR #128, opened 2025-03-20
Avoid verifying the same block, but allow verifying blocks in the same round

yacovm:


util.go
```diff
@@ -137,3 +139,58 @@ func SetRound(block VerifiedBlock, notarization *Notarization, fCert *Finalizati
 
 	return round
 }
+
+type oneTimeVerifier struct {
```
yacovm, line 143: I probably should add a UT only for this, and a UT that checks that we have verified two blocks of the same round but different digests

util.go
```diff
@@ -137,3 +139,58 @@ func SetRound(block VerifiedBlock, notarization *Notarization, fCert *Finalizati
...
+	err error
+}
+
+type oneTimeVerifiedBlock struct {
+	otv *oneTimeVerifier
+	Block
+}
+
+func (block *oneTimeVerifiedBlock) Verify(ctx context.Context) (VerifiedBlock, error) {
+	fmt.Println("#######", block.Block.BlockHeader().Digest)
```
yacovm, line 166: oops forgot to remove this

samliok (approved):

PR #132, opened 2025-03-26
Distribute Replication Requests Across Relevant Nodes

samliok:


replication.go
```diff
@@ -69,44 +113,63 @@ func (r *ReplicationState) collectMissingSequences(observedSeq uint64, nextSeqTo
...
+	reqPerNode := uint64(math.Ceil(float64(end+1-start) / float64(len(nodes))))
+
+	nodeIndex := r.requestIterator
+
+	// <= because we are inclusive
+	for curSeq := uint64(0); curSeq <= end+1-start; curSeq += reqPerNode {
+		endSeq := uint64(math.Min(float64(end), float64(start+curSeq+reqPerNode)))
+		index := nodeIndex % len(nodes)
+		// it's possible our node has signed [highestSequenceObserved].
+		// For example this may happen if our node has sent a finalization
```
yacovm, line 131: > // in this case we shouldn't send a request to ourselves.

We shouldn't send the request to ourselves in other cases as well. 
I think for simplicity instead of doing the `if` below we can just filter out our own ID above when we get the nodes to replicate from.

replication_test.go
```diff
@@ -571,7 +577,7 @@ func onlyAllowBlockProposalsAndNotarizations(msg *simplex.Message, to simplex.No
 }
 
 func TestReplicationNotarizationWithoutFinalizations(t *testing.T) {
-	nodes := []simplex.NodeID{{1}, {2}, {3}, []byte("lagging")}
```
yacovm, line 574: not that there is something wrong with this change or anything, but why the change here? is it for consistency?

samliok: yea so we use the same nodes for this check before calling the test
`isLaggingNodeLeader := bytes.Equal(simplex.LeaderForRound(nodes, numBlocks), nodes[3])`

epoch.go
```diff
@@ -1433,7 +1433,7 @@ func (e *Epoch) processFinalizedBlock(block Block, fCert FinalizationCertificate
 
 	// Schedule the block to be verified once its direct predecessor have been verified,
 	// or if it can be verified immediately.
-	e.Logger.Debug("Scheduling block verification", zap.Uint64("round", md.Round))
+	e.Logger.Debug("Scheduling block verification", zap.Uint64("round", md.Round), zap.Bool("canBeImmediatelyVerified", canBeImmediatelyVerified))
```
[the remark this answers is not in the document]
samliok: no, was just useful for debugging. ill undo this

yacovm (approved):

yacovm (commented):
LGTM

yacovm (approved):

Issue #133, opened 2025-03-27
Monitor dropping tasks we should wait for


samliok:
There is currently a flake in `TestEpochLeaderFailoverAfterProposal` that causes us to hang when calling `waitForBlockProposerTimeout`. 

`waitForBlockProposerTimeout`  can hang if we are unable to produce an empty vote or notarization. However this shouldn't happen since the test sends to the `bb.blockShouldBeBuilt` channel. Sending to this channel tells the node that a block should be built, which will then schedule a timeout with the monitor and cause our node to produce the empty vote. 

Interestingly, when the test flakes I noticed we never log this
```go
e.Logger.Info("It is time to build a block", zap.Uint64("round", round))`
```

After some further debugging, i noticed the entire `blockShouldBeBuiltNotification` function never gets called. This explained why the `blockShouldBeBuilt` channel didn't seem to be working, since we were never waiting on it. The reason why this function is never called, is because the monitor drops it from the `tasks` channel.  We add to the monitor `tasks` channel on this call(which is non-blocking)

```go
e.monitor.WaitFor(blockShouldBeBuiltNotification)
```

so the reason the test flakes is because occasionally `tasks` will be full when calling `WaitFor`, meaning we will drop the task `blockShouldBeBuiltNotification` and thus hang since we never produce an empty vote.

### How can the channel be full? 

Since the monitor runs on a different thread, sending to the `tasks` channel does not mean it will be executed right away. We could have sent a `blockShouldBeBuiltNotification` task from a previous round(that has not yet been read) by the time we send another `blockShouldBeBuiltNotification`. In this case the second `blockShouldBeBuiltNotification` will be dropped :(

yacovm:
But in the test `TestEpochLeaderFailoverAfterProposal` we only have a single failover, and the `blockShouldBeBuilt` channel is only written to once. We should only have a single `blockShouldBeBuiltNotification`, shouldn't we?

samliok:
we only trigger `blockShouldBeBuilt` once, but we attempt to put `blockShouldBeBuiltNotification` in the monitor every round

yacovm:
I see, so the invocation of `blockShouldBeBuiltNotification` for the last round may not be scheduled due to the rounds progressing too fast for the goroutine of the monitor to dequeue them, and therefore no one will invoke `IncomingBlock()`.

Nice find! 

What are your thoughts about how we should address this?

PR #135, opened 2025-03-27
Fix monitor dropping tasks

samliok:
flake described in more detail in #133 

the solution is to ensure the previous task was executed before scheduling a new one.

### DIFF
--- a/monitor.go
+++ b/monitor.go
@@ -106,7 +106,7 @@ func (m *Monitor) WaitFor(f func()) {
 	select {
 	case m.tasks <- f:
 	default:
-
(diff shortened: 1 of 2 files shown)


monitor.go
```diff
@@ -106,7 +106,7 @@ func (m *Monitor) WaitFor(f func()) {
 	select {
 	case m.tasks <- f:
 	default:
-
+		m.logger.Warn("Dropping task because the monitor tasks channel is full")
```
yacovm, line 109: When I put here `m.logger.Fatal` it didn't crash, good.

yacovm (approved):

PR #137, opened 2025-03-28
Ensure Monitor Tasks Don't Get Dropped

samliok:


monitor.go
```diff
@@ -76,9 +79,12 @@ func (m *Monitor) run() {
 		case tick := <-m.ticks:
 			m.tick(tick, taskID)
 		case f := <-m.tasks:
-			m.logger.Verbo("Executing f", zap.Uint64("taskID", taskID))
-			f()
-			m.logger.Verbo("Task executed", zap.Uint64("taskID", taskID))
+			id := taskID
+			go func() {
```
samliok, line 83: Since we handle ticks in the same goroutine as executing the tasks, we run the task in its own separate go routine so their is no deadlock

PR #140, opened 2025-03-31
Monitor

yacovm:


samliok (approved):
Overall I think this is a pretty clean solution. Couple things I would note though
Firstly I would say the term `monitor` feels a bit out of place. The struct acts more like a single task scheduler with a timeout. Not sure if `monitor` is the best name for it. 
Secondly, ideally we would reuse this struct in some way for the replication requests. Maybe by creating some sort of wrapper struct that handles multiple tasks/timeouts. 
I'll take on this task, and try to make the monitor more modular(if necessary) so we can reuse it for the replication request timeouts.

samliok (approved):

PR #143, opened 2025-04-14
Resend Replication Requests After a Timeout

samliok:
This PR introduces a `TimeoutHandler`, that tracks outgoing replication requests and resends them once a timeout has occurred. 

The logic for handling incomplete requests is as follows

- we send out replication requests, and not the timeouts in `TimeoutHandler`
- when we receive a `ReplicationResponse` we check if the response data contains all the expected sequences
  - if so, we can return early and remove the timeout task
  - if not, we create new requests by figuring out the missing sequences and distributing them across nodes

This approach is nice because it keeps the logic fairly simple. However, because we only remove `1 TimeoutTask` every time we receive a response, it's possible we could send duplicate replication requests over the network. For example say we have 2 TimeoutTasks from a node one expects seqs 5-10 and another with 11-15. If we receive seqs `5-13` from the node, we will only remove the timeout with `5-10` from the tasks maps, and resend a request for 11-15 after a timeout. To keep the logic simpler, I chose to not worry about this case since it's highly unlikely that a node would have two separate timeouts for consecutive sequences. 

- [x] Create TimeoutHandler Struct
- [x] Make tests for TimeoutHandler
- [x] Integrate TimeoutHandler into ReplicationState
- [x] Make tests for replication when nodes timeout
- [x] Ensure incomplete responses don't get forgotten

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -190,17 +190,7 @@ func (r *ReplicationState) receivedReplicationResponse(data []QuorumRound, node 
 
 	slices.Sort(seqs)
 
-	var task *TimeoutTask
-	// first find the id this request belongs too
-seq:
-	for _, seq := range seqs {
-		for _, t := range r.timeoutHandler.tasks[string(node)] {
-			if seq > t.Start && seq < t.End {
-				task = t
-				break seq
-			}
-		}
-	}
+	task := r.timeoutHandler.FindTask(node, seqs)
 
 	if task == nil {
 		return
--- a/replication.go
+++ b/replication.go
@@ -57,7 +57,6 @@ type ReplicationState struct {
 	// request iterator
 	requestIterator int
 
-	now     time.Time
 	handler TimeoutHandler
 }
 
@@ -69,13 +68,11 @@ func NewReplicationState(logger Logger, comm Communication, id NodeID, maxRoundW
 		id:                   id,
 		maxRoundWindow:       maxRoundWindow,
 		receivedQuorumRounds: make(map[uint64]QuorumRound),
-		now:                  start,
 		handler:              *NewTimeoutHandler(start),
 	}
 }
 
 func (r *ReplicationState) AdvanceTime(now time.Time) {
-	r.now = now
 	r.handler.Tick(now)
 }
 
@@ -136,10 +133,17 @@ func (r *ReplicationState) sendReplicationRequests(start uint64, end uint64) {
 		r.sendRequestToNode(nodeStart, nodeEnd, nodes, nodeIndex)
 	}
 
+	r.lastSequenceRequested = end
 	// next time we send requests, we start with a different permutation
 	r.requestIterator++
 }
 
+// sendRequestToNode requests the sequences [start, end] from nodes[index].
+// Incase the nodes[index] does not respond, we create a timeout that will
+// re-send the request.
+//
+// Note: This function can be called concurrently from the timeout handler. Therefore
+// we do not modify [r] to avoid race conditions and locking.
 func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, nodes []NodeID, index int) {
 	r.logger.Debug("Requesting missing finalization certificates ",
 		zap.Stringer("from", nodes[index]),
@@ -159,12 +163,11 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, nodes []N
 	timeoutTask := &TimeoutTask{
 		ID:      id,
 		Task:    task,
-		Timeout: r.now.Add(DefaultReplicationRequestTimeout),
+		Timeout: r.handler.GetTime().Add(DefaultReplicationRequestTimeout),
 	}
 
 	r.handler.AddTask(timeoutTask)
 
-	r.lastSequenceRequested = end
 	r.comm.SendMessage(msg, nodes[index])
 }
 
--- a/replication.go
+++ b/replication.go
@@ -57,7 +57,7 @@ type ReplicationState struct {
 	// request iterator
 	requestIterator int
 
-	now time.Time
+	now     time.Time
 	handler TimeoutHandler
 }
 
@@ -69,7 +69,7 @@ func NewReplicationState(logger Logger, comm Communication, id NodeID, maxRoundW
 		id:                   id,
 		maxRoundWindow:       maxRoundWindow,
 		receivedQuorumRounds: make(map[uint64]QuorumRound),
-		now: start,
+		now:                  start,
 	}
 }
 
@@ -156,18 +156,19 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, nodes []N
 
 	task, id := r.createReplicationTimeoutTask(start, end, nodes, index)
 	timeoutTask := &TimeoutTask{
-		ID: id,
-		Task: task,
+		ID:      id,
+		Task:    task,
 		Timeout: r.now.Add(DefaultReplicationRequestTimeout),
 	}
+
 	r.handler.AddTask(timeoutTask)
 	r.lastSequenceRequested = end
 	r.comm.SendMessage(msg, nodes[index])
 }
 
-func (r *ReplicationState) createReplicationTimeoutTask(start, end uint64, nodes []NodeID, index int) (func (), ID) {
+func (r *ReplicationState) createReplicationTimeoutTask(start, end uint64, nodes []NodeID, index int) (func(), ID) {
 	return func() {
-		r.sendRequestToNode(start, end, nodes, (index + 1) % len(nodes))
+		r.sendRequestToNode(start, end, nodes, (index+1)%len(nodes))
 	}, r.createUniqueTimeoutID(start, end, nodes, index)
 }
 
--- a/replication.go
+++ b/replication.go
@@ -159,9 +159,9 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, nodes []N
 
 	task := r.createReplicationTimeoutTask(start, end, nodes, index)
 	timeoutTask := &TimeoutTask{
-		ID:      r.getUniqueTimeoutID(start, end, nodes[index]),
-		Task:    task,
-		Timeout: r.timeoutHandler.GetTime().Add(DefaultReplicationRequestTimeout),
+		ID:       r.getUniqueTimeoutID(start, end, nodes[index]),
+		Task:     task,
+		Deadline: r.timeoutHandler.GetTime().Add(DefaultReplicationRequestTimeout),
 	}
 
 	r.timeoutHandler.AddTask(timeoutTask)
--- a/replication.go
+++ b/replication.go
@@ -73,9 +73,9 @@ func NewReplicationState(logger Logger, comm Communication, id NodeID, maxRoundW
 	}
 }
 
-func (r *ReplicationState) Tick(now time.Time) {
-	r.now = now
-	r.handler.Tick(now)
+func (r *ReplicationState) AdvanceTime(now time.Time) {
+	// r.now = now
+	// r.handler.Tick(now)
 }
 
 // isReplicationComplete returns true if we have finished the replication process.
@@ -162,6 +162,7 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, nodes []N
 	}
 
 	r.handler.AddTask(timeoutTask)
+
 	r.lastSequenceRequested = end
 	r.comm.SendMessage(msg, nodes[index])
 }
@@ -169,11 +170,29 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, nodes []N
 func (r *ReplicationState) createReplicationTimeoutTask(start, end uint64, nodes []NodeID, index int) (func(), ID) {
 	return func() {
 		r.sendRequestToNode(start, end, nodes, (index+1)%len(nodes))
-	}, r.createUniqueTimeoutID(start, end, nodes, index)
+	}, r.createUniqueTimeoutID(start, end, nodes[index])
 }
 
-func (r *ReplicationState) createUniqueTimeoutID(start, end uint64, nodes []NodeID, index int) ID {
-	return ID(fmt.Sprintf("timeout_%d_to_%d_nodes_%s_index_%d", start, end, NodeIDs(nodes), index))
+func (r *ReplicationState) createUniqueTimeoutID(start, end uint64, node NodeID) ID {
+	return ID(fmt.Sprintf("timeout_%d_to_%d_node_%s", start, end, node))
+}
+
+func (r *ReplicationState) receivedReplicationResponse(data []QuorumRound, node NodeID) {
+	var min, max uint64
+	min = math.MaxUint64
+
+	for _, qr := range data {
+		if qr.GetSequence() > max {
+			max = qr.GetSequence()
+		}
+
+		if qr.GetSequence() < min {
+			min = qr.GetSequence()
+		}
+	}
+
+	id := r.createUniqueTimeoutID(min, max, node)
+	r.handler.RemoveTask(id)
 }
 
 func (r *ReplicationState) replicateBlocks(fCert *FinalizationCertificate, nextSeqToCommit uint64) {
--- a/replication.go
+++ b/replication.go
@@ -120,7 +120,7 @@ func (r *ReplicationState) sendReplicationRequests(start uint64, end uint64) {
 	numSeqs := end + 1 - start
 	seqsPerNode := numSeqs / uint64(numNodes)
 
-	r.logger.Debug("Distributing replication requests", zap.Uint64("start", start), zap.Uint64("end", end), zap.Stringer("ndoes", NodeIDs(nodes)))
+	r.logger.Debug("Distributing replication requests", zap.Uint64("start", start), zap.Uint64("end", end), zap.Stringer("nodes", NodeIDs(nodes)))
 	// Distribute sequences evenly among nodes in round-robin fashion
 	for i := range numNodes {
 		nodeIndex := (r.requestIterator + i) % numNodes
@@ -195,6 +195,7 @@ func (r *ReplicationState) receivedReplicationResponse(data []QuorumRound, node 
 
 	task := r.timeoutHandler.FindTask(node, seqs)
 	if task == nil {
+		r.logger.Debug("Could not find a timeout task associated with the replication response", zap.Stringer("from", node))
 		return
 	}
 	r.timeoutHandler.RemoveTask(node, task.TaskID)
@@ -206,6 +207,7 @@ func (r *ReplicationState) receivedReplicationResponse(data []QuorumRound, node 
 	}
 
 	// if not all sequences were returned, create new timeouts
+	r.logger.Debug("Received missing sequences in the replication response", zap.Stringer("from", node), zap.Any("missing", missing))
 	nodes := r.highestSequenceObserved.signers.Remove(r.id)
 	lenNodes := len(nodes)
 
@@ -229,9 +231,9 @@ func (r *ReplicationState) receivedReplicationResponse(data []QuorumRound, node 
 // findMissingNumbersInRange finds numbers in an array constructed by [start...end] that are not in [nums]
 // ex. (3, 10, [1,2,3,4,5,6]) -> [7,8,9,10]
 func findMissingNumbersInRange(start, end uint64, nums []uint64) []uint64 {
-	numMap := make(map[uint64]bool)
+	numMap := make(map[uint64]struct{})
 	for _, num := range nums {
-		numMap[num] = true
+		numMap[num] = struct{}{}
 	}
 
 	var result []uint64
--- a/replication.go
+++ b/replication.go
@@ -226,19 +226,17 @@ func (r *ReplicationState) receivedReplicationResponse(data []QuorumRound, node 
 	r.timeoutHandler.AddTask(newTask)
 }
 
+// findMissingNumbersInRange finds numbers in an array constructed by [start...end] that are not in [nums]
+// ex. (3, 10, [1,2,3,4,5,6]) -> [7,8,9,10]
 func findMissingNumbersInRange(start, end uint64, nums []uint64) []uint64 {
-	// Create a map to track numbers in the input array for O(1) lookups
 	numMap := make(map[uint64]bool)
 	for _, num := range nums {
 		numMap[num] = true
 	}
 
-	// Create result array of missing numbers
 	var result []uint64
 
-	// Check each number in the range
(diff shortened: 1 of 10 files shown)


timeout_handler.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// find the task using the task map
+	// remove it from the heap using the index
+	heap.Remove(&t.heap, t.tasks[ID].index)
+	delete(t.tasks, ID)
+}
+
+// ----------------------------------------------------------------------
+type TaskHeap []*TimeoutTask
```
yacovm, line 99: If we assume that time moves only forward and not backwards, and we always add a task with the same timeout, then why do we need a heap and not a list? Is it ever possible that we add an element and it won't be the last one?

timeout_handler.go
```diff
@@ -0,0 +1,135 @@
...
+	// go through the heap executing relevant tasks
+	for t.heap.Len() > 0 {
+		next := t.heap[0]
+		if next.Timeout.After(t.now) {
+			break
+		}
+
+		heap.Pop(&t.heap)
+		delete(t.tasks, next.ID)
+		go next.Task()
```
yacovm, line 66: I would rather we not create goroutines in a non-controlled manner like that. Can we not run this in this current goroutine?

timeout_handler.go
samliok: was thinking this gives us more flexibility. if we decide to send timeout requests with different timeouts(for example exponential backoff on certain nodes) then using a heap will properly order them

replication.go
```diff
@@ -55,19 +56,26 @@ type ReplicationState struct {
 
 	// request iterator
 	requestIterator int
+
+	handler TimeoutHandler
```
yacovm, line 60: `timeoutHandler` ?

replication.go
```diff
@@ -144,8 +157,44 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, node Node
...
+		if qr.GetSequence() > max {
+			max = qr.GetSequence()
+		}
+
+		if qr.GetSequence() < min {
+			min = qr.GetSequence()
+		}
+	}
+
+	id := r.getUniqueTimeoutID(min, max, node)
```
yacovm, line 196: what if we get partial data from the node? then the task is not removed on time and we time out, yet we'll send a request to replicate the previous range of blocks, won't we?

replication_timeout_test.go
```diff
@@ -0,0 +1,194 @@
+package simplex_test
```
yacovm, line 1: license headers missing

timeout_handler.go
```diff
@@ -0,0 +1,148 @@
...
+		if next.Timeout.After(t.now) {
+			t.lock.Unlock()
+			break
+		}
+
+		heap.Pop(&t.heap)
+		delete(t.tasks, next.ID)
+		t.lock.Unlock()
+		t.log.Debug("Executing timeout task", zap.String("taskid", next.ID))
+		next.Task()
```
yacovm, line 77: The `Tick` is invoked by the goroutine that advances the time, which is supposed to be nearly instant. It also affects the [monitor's time advancement](https://github.com/ava-labs/Simplex/blob/main/monitor.go#L44-L50).

I have a slight concern that `r.comm.SendMessage(msg, nodes[index])` done by `sendRequestToNode` can slow down the time advancement goroutine repeatedly and it'll interfere with the monitor.

Is it possible to only update the time, and have the for loop run endlessly in a goroutine of its own? This way if sending the message is too slow, the time will flow naturally.

timeout_handler.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (h *TaskHeap) Len() int { return len(*h) }
+
+// Less returns if the task at index [i] has a lower timeout than the task at index [j]
+func (h *TaskHeap) Less(i, j int) bool { return (*h)[i].Timeout.Before((*h)[j].Timeout) }
+
+// Swap swaps the values at index [i] and [j]
+func (h *TaskHeap) Swap(i, j int) {
+	(*h)[i], (*h)[j] = (*h)[j], (*h)[i]
+	(*h)[i].index = i
```
yacovm, line 122: doesn't the above line (`(*h)[i], (*h)[j] = (*h)[j], (*h)[i]`) also take care of `i` and `j` ?

timeout_handler.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"go.uber.org/zap"
+)
+
+type TimeoutTask struct {
+	ID      string
+	Task    func()
+	Timeout time.Time
+
+	index int // for heap to work more efficiently
```
yacovm, line 16: this is basically the location of the task in the heap, no?

timeout_handler.go
```diff
@@ -0,0 +1,148 @@
...
+	"sync"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+type TimeoutTask struct {
+	ID      string
+	Task    func()
+	Timeout time.Time
```
yacovm, line 14: Deadline?

timeout_handler.go
samliok: yep

timeout_handler.go
samliok: yes, but since we are swapping [i] and [j] we need to reset their inner `index` values so that `remove()` works properly

timeout_handler.go
samliok: rather than have a spinning loop should we just change `advanceTime` in `replicationState` to run in a different go routine?

change
```go
func (r *ReplicationState) AdvanceTime(now time.Time) {
	r.timeoutHandler.Tick(now)
}
```
to
```go
func (r *ReplicationState) AdvanceTime(now time.Time) {
	go r.timeoutHandler.Tick(now)
}
```

replication.go
samliok: if we get something other than the expected range then the ID will be different than expected and we will resend a 
request to a different node. 

So for example if we request seqs `5-10` from a node, but they respond with `5-7` we will timeout on the task. This will resend the replication request for seqs `5-10`(to a different node however). A bigger problem is that if we send req for `5-10` and only get seq `5` and `10` in response. In this case we wont ever try and collect `6-9`.

I think the best way to fix this is to make sure we receive all the seqs in the range of the `id`. If we have some gaps, we should just spin off new timeouts. So if we just receive 5 & 10, we would create a TimeoutTask for 6-9. If we just receive 5-7, we'll create one from 8-10. 

i'll update and also create a test for this.

yacovm: > This will resend the replication request for seqs 5-10(to a different node however). 

Right, this is excessive traffic and I/O. 

>  A bigger problem is that if we send req for 5-10 and only get seq 5 and 10 in response. In this case we wont ever try and collect 6-9.

Yes good point. 

> If we have some gaps, we should just spin off new timeouts. So if we just receive 5 & 10, we would create a TimeoutTask for 6-9. If we just receive 5-7, we'll create one from 8-10.

Makes sense.

timeout_handler.go
yacovm: ohhh right because when we move the `j`'s element to `i` it still has the `j` index in it

timeout_handler.go
```diff
@@ -53,12 +63,14 @@ func (t *TimeoutHandler) GetTime() time.Time {
...
 }
 
-func (t *TimeoutHandler) Tick(now time.Time) {
-	t.lock.Lock()
-	// update the time of the handler
-	t.now = now
-	t.lock.Unlock()
+func (t *TimeoutHandler) run() {
+	for t.shouldRun() {
+		<-t.ticks
```
yacovm, line 68: can we simultaneously select from ticks and close to immediately close and not wait for the tick?

replication.go
```diff
@@ -144,8 +158,91 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, node Node
...
+	seqs := make([]uint64, 0, len(data))
+
+	for _, qr := range data {
+		seqs = append(seqs, qr.GetSequence())
+	}
+
+	slices.Sort(seqs)
+
+	task := r.timeoutHandler.FindTask(node, seqs)
+	if task == nil {
```
yacovm, line 197: can we log this?

timeout_handler.go
```diff
@@ -0,0 +1,231 @@
...
+	index int // for heap to work more efficiently
+}
+
+type TimeoutHandler struct {
+	lock sync.Mutex
+
+	ticks chan time.Time
+	close chan struct{}
+	// nodeids -> range -> task
+	tasks map[string]map[string]*TimeoutTask
```
yacovm, line 32: isn't it simpler to just have:

`tasks map[string]*TimeoutTask 

where the key is:

```
func getTimeoutID(node NodeID, start, end uint64) string {
	return fmt.Sprintf("%x%d%s%d", node, start, delimiter, end)
}
```

? 

saves you an indirection and simplifies the code, in my opinion.

replication.go
```diff
@@ -144,8 +158,91 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, node Node
...
+	}
+	r.timeoutHandler.RemoveTask(node, task.TaskID)
+
+	// we found the timeout, now make sure all seqs were returned
+	missing := findMissingNumbersInRange(task.Start, task.End, seqs)
+	if len(missing) == 0 {
+		return
+	}
+
+	// if not all sequences were returned, create new timeouts
```
yacovm, line 208: can we log that we reached here?

replication.go
```diff
@@ -144,8 +158,91 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, node Node
...
+	nodes := r.highestSequenceObserved.signers.Remove(r.id)
+	lenNodes := len(nodes)
+
+	start := missing[0]
```
yacovm, line 212:
```
	start := missing[0]
	end := missing[0]
	for i, missingSeq := range missing[1:] {
		if missingSeq != missing[i]+1 {
			index := i % lenNodes
			newTask := r.createReplicationTimeoutTask(start, end, nodes, index)
			r.timeoutHandler.AddTask(newTask)
			start = missingSeq
		}

		end = missingSeq
```

Is it possible to abstract out the creation of the sections into a function (let's denote it `f() []struct{start uint64; end uint64}` and add a UT for it, and then just do:

```
for _, startEndPair := range f() { 
		start := startEnd.start
		end := startEnd.end
                 newTask := r.createReplicationTimeoutTask(start, end, nodes, index)
}
```

?

replication.go
```diff
@@ -144,8 +158,91 @@ func (r *ReplicationState) sendRequestToNode(start uint64, end uint64, node Node
...
+	}
+
+	newTask := r.createReplicationTimeoutTask(start, end, nodes, len(missing)%lenNodes)
+	r.timeoutHandler.AddTask(newTask)
+}
+
+// findMissingNumbersInRange finds numbers in an array constructed by [start...end] that are not in [nums]
+// ex. (3, 10, [1,2,3,4,5,6]) -> [7,8,9,10]
+func findMissingNumbersInRange(start, end uint64, nums []uint64) []uint64 {
+	numMap := make(map[uint64]bool)
```
yacovm, line 232: `numMap := make(map[uint64]struct{})` is more idiomatic in Go

timeout_handler.go
```diff
@@ -0,0 +1,231 @@
...
+// FindTask returns the first TimeoutTask assigned to [node] that contains any sequence in [seqs].
+// A sequence is considered "contained" if it falls between a task's Start (inclusive) and End (inclusive).
+func (t *TimeoutHandler) FindTask(node NodeID, seqs []uint64) *TimeoutTask {
+	var task *TimeoutTask
+seq:
+	for _, seq := range seqs {
+		for _, t := range t.tasks[string(node)] {
+			if seq >= t.Start && seq <= t.End {
+				task = t
+				break seq
```
yacovm, line 180: can we just return `t` and return `nil` in line 185 and remove the `seq:` label? Makes things more explicit, and I personally just hate goto labels, maybe that's just me...

timeout_handler.go
```diff
@@ -0,0 +1,231 @@
...
+	old := *h
+	len := h.Len()
+	task := old[len-1]
+	old[len-1] = nil
+	*h = old[0 : len-1]
+	task.index = -1
+	return task
+}
+
+func (h *TaskHeap) Peep() any {
```
yacovm, line 225: why do we need `Peep()` ? I don't think it's used anywhere. 

To implement a head we just need:

```
type Interface interface {
	sort.Interface
	Push(x any) // add x as element Len()
	Pop() any   // remove and return element Len() - 1.
}
```

Also, did you mean `Peek()` ?

timeout_handler.go
```diff
@@ -0,0 +1,231 @@
...
+	tasks := make(map[string]map[string]*TimeoutTask)
+	for _, node := range nodes {
+		tasks[string(node)] = make(map[string]*TimeoutTask)
+	}
+
+	t := &TimeoutHandler{
+		now:   startTime,
+		tasks: tasks,
+		heap:  TaskHeap{},
+		ticks: make(chan time.Time, 2),
```
yacovm, line 51: `ticks: make(chan time.Time, 2),` is this because of the tests?

timeout_handler.go
```diff
@@ -0,0 +1,231 @@
...
+func NewTimeoutHandler(log Logger, startTime time.Time, nodes []NodeID) *TimeoutHandler {
+	tasks := make(map[string]map[string]*TimeoutTask)
+	for _, node := range nodes {
+		tasks[string(node)] = make(map[string]*TimeoutTask)
+	}
+
+	t := &TimeoutHandler{
+		now:   startTime,
+		tasks: tasks,
+		heap:  TaskHeap{},
```
yacovm, line 50: nit, but i don't think we need to initialize `TaskHeap{}`

timeout_handler_test.go
```diff
@@ -0,0 +1,313 @@
...
+	handler.Tick(start.Add(3 * time.Second))
+
+	<-finished
+	<-finished
+
+	mu.Lock()
+	defer mu.Unlock()
+
+	require.Equal(t, 2, len(results))
+	require.Contains(t, results, "first")
```
yacovm, line 121: but this doesn't test their order... 

maybe: 

```
	require.Contains(t, results[:1], "first")
	require.Contains(t, results[1:], "second")
```

or `require.Equal(t, []string{"first", "second"}, results)`

timeout_handler_test.go
```diff
@@ -0,0 +1,313 @@
...
+	require.Contains(t, results, "second")
+}
+
+func TestAddTasksOutOfOrder(t *testing.T) {
+	for range 100 {
+		start := time.Now()
+		l := testutil.MakeLogger(t, 1)
+		nodes := []simplex.NodeID{{1}, {2}}
+		handler := simplex.NewTimeoutHandler(l, start, nodes)
```
yacovm, line 130: can we close the timeout handlers at the end of each test?

timeout_handler_test.go
```diff
@@ -0,0 +1,313 @@
...
+
+func TestTaskOrder(t *testing.T) {
+	start := time.Now()
+	l := testutil.MakeLogger(t, 1)
+	nodes := []simplex.NodeID{{1}, {2}}
+	handler := simplex.NewTimeoutHandler(l, start, nodes)
+	finished := make(chan struct{})
+
+	var mu sync.Mutex
+	results := []string{}
```
yacovm, line 75:
```
	var results []string
```

is more idiomatic

timeout_handler_test.go
```diff
@@ -0,0 +1,313 @@
...
+		require.Equal(t, 1, len(results))
+		require.Contains(t, results, "first")
+		mu.Unlock()
+
+		handler.Tick(start.Add(3 * time.Second))
+		<-finished
+		<-finished
+		mu.Lock()
+		require.Equal(t, 3, len(results))
+		require.Contains(t, results, "second")
```
yacovm, line 195: can we test their order?

replication_timeout_test.go
```diff
@@ -0,0 +1,280 @@
...
+	for i := 0; i <= int(startSeq); i++ {
+		for _, n := range net.instances {
+			if n.e.ID.Equals(laggingNode.e.ID) {
+				continue
+			}
+			n.storage.waitForBlockCommit(uint64(startSeq))
+		}
+	}
+
+	// assert the lagging node has not received any replicaiton requests
```
yacovm, line 58: replicaiton --> replication

replication_timeout_test.go
```diff
@@ -0,0 +1,280 @@
...
+	require.Equal(t, uint64(0), laggingNode.storage.Height())
+
+	// after the timeout, the nodes should respond and the lagging node will replicate
+	net.setAllNodesMessageFilter(allowAllMessages)
+	laggingNode.e.AdvanceTime(laggingNode.e.StartTime.Add(simplex.DefaultReplicationRequestTimeout / 2))
+	require.Equal(t, uint64(0), laggingNode.storage.Height())
+
+	laggingNode.e.AdvanceTime(laggingNode.e.StartTime.Add(simplex.DefaultReplicationRequestTimeout * 2))
```
yacovm, line 66: why are we using `DefaultReplicationRequestTimeout * 2` and not `DefaultReplicationRequestTimeout /2` ?

replication_timeout_test.go
```diff
@@ -0,0 +1,280 @@
...
+			n.storage.waitForBlockCommit(uint64(startSeq))
+		}
+	}
+
+	// ensure lagging node doesn't resend requests
+	mf := &testTimeoutMessageFilter{
+		t: t,
+	}
+	laggingNode.e.Comm.(*testComm).setFilter(mf.failOnReplicationRequest)
+	laggingNode.e.AdvanceTime(laggingNode.e.StartTime.Add(simplex.DefaultReplicationRequestTimeout * 2))
```
yacovm, line 127: can we have several iterations of advanceTime and maybe notarize / finalize another block? Since all nodes run in their own goroutine, the test just ends here and will not give a chance to anyone to process any message.

replication_timeout_test.go
```diff
@@ -0,0 +1,280 @@
...
+	laggingNode.e.AdvanceTime(laggingNode.e.StartTime.Add(simplex.DefaultReplicationRequestTimeout))
+	laggingNode.storage.waitForBlockCommit(startSeq / 3)
+
+	net.setAllNodesMessageFilter(allowAllMessages)
+	// timeout again, now all nodes will respond
+	laggingNode.e.AdvanceTime(laggingNode.e.StartTime.Add(simplex.DefaultReplicationRequestTimeout * 2))
+	laggingNode.storage.waitForBlockCommit(startSeq)
+}
+
+// modifies the replication response to only send every other quorom round
```
yacovm, line 199: quorom --> quorum

timeout_handler.go
```diff
@@ -0,0 +1,222 @@
...
+	default:
+		t.log.Debug("Dropping tick in timeouthandler")
+	}
+}
+
+func (t *TimeoutHandler) AddTask(task *TimeoutTask) {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	if _, ok := t.tasks[string(task.NodeID)]; !ok {
```
yacovm, line 126: can we log this in debug?

timeout_handler.go
```diff
@@ -0,0 +1,222 @@
...
+		return
+	}
+
+	if _, ok := t.tasks[string(nodeID)][ID]; !ok {
+		return
+	}
+
+	// find the task using the task map
+	// remove it from the heap using the index
+	t.log.Debug("Removing timeout task", zap.String("taskid", ID))
```
yacovm, line 156: maybe log the node here as well as we did in the add?

timeout_handler_test.go
```diff
@@ -0,0 +1,318 @@
...
+		},
+	}
+
+	handler.AddTask(task)
+	handler.Tick(start.Add(2 * time.Second))
+	time.Sleep(10 * time.Millisecond)
+
+	require.Zero(t, len(sent))
+	handler.Tick(start.Add(6 * time.Second))
+	<-sent
```
yacovm, line 40: can we add a section that ensures the task is only executed once?

replication_timeout_test.go
```diff
@@ -0,0 +1,286 @@
...
+)
+
+func rejectReplicationRequests(msg *simplex.Message, from simplex.NodeID) bool {
+	return msg.ReplicationRequest == nil && msg.ReplicationResponse == nil && msg.VerifiedReplicationResponse == nil
+}
+
+// A node attempts to request blocks to replicate, but fails to receive them
+func TestReplicationRequestTimeout(t *testing.T) {
+	nodes := []simplex.NodeID{{1}, {2}, {3}, []byte("lagging")}
+	numInitalSeqs := uint64(8)
```
yacovm, line 20: `numInitialSeqs`

timeout_handler.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	// adds a task to the heap and the tasks map
+	if _, ok := t.tasks[string(task.NodeID)][task.TaskID]; ok {
+		t.log.Debug("Trying to add an already included task", zap.Stringer("node", task.NodeID), zap.String("Task ID", task.TaskID))
+		return
+	}
+
+	t.tasks[string(task.NodeID)][task.TaskID] = task
+	t.log.Debug("Adding timeout task", zap.Stringer("node", task.NodeID), zap.String("taskid", task.TaskID))
```
yacovm, line 137: nit, but when i run the test i get:

```
[04-24|00:10:19.239] DEBUG Simplex/timeout_handler.go:137 Adding timeout task {"test": "TestReplicationRequestTimeoutMultiple", "node": 108, "node": "0100000000000000", "taskid": "0_3"}
```

Can we change the `node` here to something else like `from` ? This is confusing...

replication_timeout_test.go
```diff
@@ -0,0 +1,286 @@
...
+	for i := 0; i <= int(startSeq); i++ {
+		for _, n := range net.instances {
+			if n.e.ID.Equals(laggingNode.e.ID) {
+				continue
+			}
+			n.storage.waitForBlockCommit(uint64(startSeq))
+		}
+	}
+
+	// this is done from normalNode2 since the lagging node will request
```
yacovm, line 267: why is this normalNode2 and not the first node? We don't have a time advancement yet, so shouldn't the lagging node ask for the first sequences from the first node?

The first node has its filter set to reject replication requests, but in order for the lagging node to ask from the second node, time needs to pass in the test, no?

yacovm (approved):
Left some nits, LGTM otherwise.

replication_timeout_test.go
samliok: this is helpful to wait for `normalNode2` to send out its first replication response _before_ we get its second response. Otherwise, the test would occasionally flake.

PR #145, opened 2025-04-17
Ensure fCerts aren't dropped during replication

samliok:
we may drop some `fCerts` during replication if they are not the next sequence to commit

the reason `TestReplicationNodeDiverges` fails without the proposed changes is because it hits a code path during replication where we receive an `fCert` for a round that is not the next seq to commit. However, that fcert is for the current round and our node is technically all caught up. 

Similar to the bug fix from the previous pr(#148) we should increase the round, and potentially load future messages so the node doesn't get stuck. The new code changes to `epoch` allow for this to happen, as before we would just drop the finalization certificate and not perform those actions.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -835,6 +835,42 @@ func (e *Epoch) progressRoundsDueToCommit(round uint64) {
 	}
 }
 
+func (e *Epoch) storeFutureFinalizationCertificate(fCert *FinalizationCertificate, replicate bool) error {
+	nextSeqToCommit := e.Storage.Height()
+	recordBytes := NewQuorumRecord(fCert.QC.Bytes(), fCert.Finalization.Bytes(), record.FinalizationRecordType)
+	if err := e.WAL.Append(recordBytes); err != nil {
+		e.Logger.Error("Failed to append finalization certificate record to WAL", zap.Error(err))
+		return err
+	}
+
+	e.Logger.Debug("Persisted finalization certificate to WAL",
+		zap.Uint64("round", fCert.Finalization.Round),
+		zap.Uint64("height", nextSeqToCommit),
+		zap.Int("size", len(recordBytes)),
+		zap.Stringer("digest", fCert.Finalization.BlockHeader.Digest))
+
+	// we receive a finalization certificate for a future round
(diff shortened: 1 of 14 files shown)


epoch.go
```diff
@@ -2485,16 +2484,22 @@ func (e *Epoch) maybeAdvanceRoundFromEmptyNotarizations() (bool, error) {
 // getHighestRound returns the highest round that has either a notarization or finalization
 func (e *Epoch) getHighestRound() *Round {
 	var max uint64
+	var found bool
```
samliok, line 2487: without this we will return round `e.rounds[0]` even if there is no notarization or finalization. Going to make a separate test for this

epoch.go
```diff
@@ -2485,16 +2484,22 @@ func (e *Epoch) maybeAdvanceRoundFromEmptyNotarizations() (bool, error) {
...
 
 	for _, round := range e.rounds {
-		if round.num > max {
+		if round.num >= max {
```
samliok, line 2490: must be >= since we should also check the 0th round

epoch.go
```diff
@@ -835,45 +835,54 @@ func (e *Epoch) progressRoundsDueToCommit(round uint64) {
 	}
 }
 
-func (e *Epoch) persistFinalizationCertificate(fCert FinalizationCertificate) error {
-	e.Logger.Debug("Received enough finalizations to finalize a block", zap.Uint64("round", fCert.Finalization.Round))
-	// Check to see if we should commit this finalization to the storage as part of a block commit,
-	// or otherwise write it to the WAL in order to commit it later.
-	startRound := e.round
+func (e *Epoch) storeFutureFinalizationCertificate(fCert *FinalizationCertificate, replicate bool) error {
```
yacovm, line 838: I don't understand why this `replicate` variable is properly named. 

It seems to me it is inversely named to its intention - when we replicate we pass false, and when we don't replicate we pass true.

samliok: changed to `shouldReplicate`

replication_test.go
```diff
@@ -633,12 +633,118 @@ func testReplicationNotarizationWithoutFinalizations(t *testing.T, numBlocks uin
...
+		if to.Equals(simplex.NodeID{4}) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// TestReplicationNodeDiverges
+func TestReplicationNodeDiverges(t *testing.T) {
```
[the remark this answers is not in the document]
samliok: added to the description of the PR

PR #146, opened 2025-04-20
fix getHighestRound bug

samliok:


epoch.go
```diff
@@ -2485,17 +2485,23 @@ func (e *Epoch) maybeAdvanceRoundFromEmptyNotarizations() (bool, error) {
 // getHighestRound returns the highest round that has either a notarization or finalization
 func (e *Epoch) getHighestRound() *Round {
 	var max uint64
+	var found bool
 
 	for _, round := range e.rounds {
-		if round.num > max {
+		if round.num >= max {
```
yacovm, line 2491: this is only a bug for the case where our ledger is empty, right?

In reality we will always have a genesis block, so our ledger will never be empty, so we will never have the 0 round in our memory.

yacovm (approved):

PR #148, opened 2025-04-20
Increase the round after receiving an fcert

samliok:
In our previous code, if we received a fCert for the current round but it wasn't the next sequence to commit we would not increase the round and start a new one.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -1554,7 +1554,11 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
 
 		// We might have received votes and finalizations from future rounds before we received this block.
 		// So load the messages into our round data structure now that we have created it.
-		e.maybeLoadFutureMessages()
+		err = e.maybeLoadFutureMessages()
+		if err != nil {
+			e.haltedError = err
+			return md.Digest
+		}
 
 		// Check if we have timed out on this round.
 		// Although we store the proposal for this round,
@@ -1868,7 +1872,10 @@ func (e *Epoch) proposeBlock(block VerifiedBlock) error {
 
 	// We might have received votes and finalizations from future rounds before we received this block.
 	// So load the messages into our round data structure now that we have created it.
-	e.maybeLoadFutureMessages()
+	err = e.maybeLoadFutureMessages()
+	if err != nil {
+		return err
+	}
 
 	e.Comm.Broadcast(proposal)
 	e.Logger.Debug("Proposal broadcast",
@@ -1876,7 +1883,7 @@ func (e *Epoch) proposeBlock(block VerifiedBlock) error {
 		zap.Int("size", len(rawBlock)),
 		zap.Stringer("digest", md.Digest))
 
-	return errors.Join(e.handleVoteMessage(&vote, e.ID), e.maybeLoadFutureMessages())
+	return e.handleVoteMessage(&vote, e.ID)
 }
 
 // Metadata returns the metadata of the next expected block of the epoch.
--- a/epoch.go
+++ b/epoch.go
@@ -1514,7 +1514,7 @@ func (e *Epoch) processNotarizedBlock(block Block, notarization *Notarization) e
 			delete(e.rounds, md.Round)
 			return e.processNotarizedBlock(block, notarization)
 		}
-		fmt.Println("*******proceess persist")
+		fmt.Println("*******proceess persist in hereee")
 		if err := e.persistNotarization(*notarization); err != nil {
 			e.Logger.Warn("Failed to persist notarization", zap.Error(err))
 			e.haltedError = err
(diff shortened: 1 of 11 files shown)


epoch_test.go
```diff
@@ -170,6 +170,53 @@ func TestEpochConsecutiveProposalsDoNotGetVerified(t *testing.T) {
...
+
+	e, err := NewEpoch(conf)
+	require.NoError(t, err)
+
+	require.NoError(t, e.Start())
+
+	block, _ := advanceRoundFromNotarization(t, e, bb)
+	advanceRoundFromFinalization(t, e, bb)
+	require.Equal(t, uint64(2), e.Metadata().Round)
+	require.Equal(t, uint64(0), storage.Height())
```
yacovm, line 206: I don't understand why we would want to have a 0 height here. 

Shouldn't we expect the two blocks to commit?

epoch_test.go
```diff
@@ -170,6 +170,53 @@ func TestEpochConsecutiveProposalsDoNotGetVerified(t *testing.T) {
...
+		SignatureAggregator: &testSignatureAggregator{},
+	}
+
+	e, err := NewEpoch(conf)
+	require.NoError(t, err)
+
+	require.NoError(t, e.Start())
+
+	block, _ := advanceRoundFromNotarization(t, e, bb)
+	advanceRoundFromFinalization(t, e, bb)
```
yacovm, line 204: I am not sure we should move to the next round. I think what should happen is we either:

1. Try to collect a finalization for this block via `rebroadcastPastFinalizations`, or:
2. Try to replicate from other nodes.

epoch.go
```diff
@@ -1861,6 +1866,10 @@ func (e *Epoch) proposeBlock(block VerifiedBlock) error {
 		return errors.New("failed to store block proposed by me")
 	}
 
+	// We might have received votes and finalizations from future rounds before we received this block.
+	// So load the messages into our round data structure now that we have created it.
+	e.maybeLoadFutureMessages()
```
yacovm, line 1871: we already call `maybeLoadFutureMessages` in line 1879, why do we need to do this here again?

epoch.go
```diff
@@ -1543,6 +1552,10 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
 			// TODO: timeout
 		}
 
+		// We might have received votes and finalizations from future rounds before we received this block.
+		// So load the messages into our round data structure now that we have created it.
+		e.maybeLoadFutureMessages()
```
yacovm, line 1557: Are we doing it here because we removed it from `storeProposal` ?

epoch.go
```diff
@@ -2097,6 +2105,7 @@ func (e *Epoch) increaseRound() {
 		zap.Uint64("old round", e.round),
 		zap.Uint64("new round", e.round+1),
 		zap.Stringer("leader", leader))
+
```
yacovm, line 2108: can we please revert the empty lines? later they confuse the git blame

epoch.go
```diff
@@ -2463,10 +2470,12 @@ func (e *Epoch) maybeAdvanceRoundFromEmptyNotarizations() (bool, error) {
 	nextSeqQuorum := e.replicationState.GetQuroumRoundWithSeq(expectedSeq)
 	if nextSeqQuorum != nil {
 		// num empty notarizations
-		for range nextSeqQuorum.GetRound() - round {
-			e.increaseRound()
+		if round < nextSeqQuorum.GetRound() {
+			for range nextSeqQuorum.GetRound() - round {
```
yacovm, line 2474: if `round >= nextSeqQuorum.GetRound()` then it's a no-op.

replication_test.go
```diff
@@ -235,10 +235,11 @@ func TestReplicationNotarizations(t *testing.T) {
 			if n.e.ID.Equals(leader) && n.e.ID.Equals(nodes[3]) {
 				continue
 			}
-			n.wal.assertNotarization(uint64(i))
+
+			n.wal.assertNotarizationOrFinalization(uint64(i), n.e.EpochConfig.QCDeserializer)
 		}
 	}
-
+	time.Sleep(10 * time.Millisecond)
```
yacovm, line 242: why do we need this sleep?

util.go
```diff
@@ -181,6 +188,7 @@ func (block *oneTimeVerifiedBlock) Verify(ctx context.Context) (VerifiedBlock, e
 	}()
 
 	if result, exists := block.otv.digests[digest]; exists {
+		block.otv.logger.Warn("Attempted to verify an already verified block", zap.Uint64("round", header.Round))
```
yacovm, line 191: can we please extract this one time verifier change to its own PR?

epoch.go
samliok: no since they are `uints` this will underflow and we will get `for range max.Uint64`

epoch.go
samliok: yea, i thought it would be cleaner to have `storeProposal` not call `maybeLoadFutureMessages`

epoch.go
samliok: ah good point, i added this because `storeProposal` was calling it. I think we should call `maybeLoadFutureMessages` before `handleVoteMessage` so i'll remove the second call

epoch.go
yacovm: oops, yeah you're right

epoch.go
```diff
@@ -1861,13 +1870,20 @@ func (e *Epoch) proposeBlock(block VerifiedBlock) error {
 		return errors.New("failed to store block proposed by me")
 	}
 
+	// We might have received votes and finalizations from future rounds before we received this block.
```
yacovm, line 1873: Why do we load the future messages before broadcasting the proposal? 

If we do it earlier then shouldn't we check if we have advanced to a higher round and if so, abort early and not broadcast the proposal?

yacovm (approved):

PR #151, opened 2025-04-24
Reduce Redundant Sending of Replication Requests

samliok:
If we had 5 nodes and only 3 seqs to request, before we would send requests to all five nodes with the first 4 nodes having the same request of the first sequence. 

Also this PR separates the logic of creating the requests into a helper function with unit tests.


yacovm:
> If we had 5 nodes and only 3 seqs to request, before we would send requests to all five nodes with the first 4 nodes having the same request of the first sequence.
> 
> Also this PR separates the logic of creating the requests into a helper function with unit tests.

because 3/5 is 0?

util.go
```diff
@@ -239,3 +239,42 @@ func CompressSequences(missingSeqs []uint64) []Segment {
...
+
+	if seqsPerNode == 0 {
+		seqsPerNode = 1
+	}
+
+	nodeStart := start
+
+	for i := 0; i < numNodes && nodeStart <= end; i++ {
+		segmentLength := seqsPerNode
+		if remainder > 0 {
```
yacovm, line 264: so here you sprinkle the remainder across the various segments?

yacovm (approved):

yacovm:
LGTM

PR #154, opened 2025-04-29
rebroadcast empty votes until a round advances

samliok:


timeout_handler.go
```diff
@@ -168,23 +172,6 @@ func (t *TimeoutHandler) Close() {
 	}
 }
 
-// FindTask returns the first TimeoutTask assigned to [node] that contains any sequence in [seqs].
```
yacovm, line 171: why did you move this to replication? Just because it doesn't have a `th` receiver?

samliok: yep

PR #160, opened 2025-05-05
Add Silence() method to the test logger

yacovm:


samliok (approved):
i think it would also be helpful to add a helper to `testNode` or `testInstance` to call silence on the underlying logger. but this can be done whenever it comes up

yacovm:
> i think it would also be helpful to add a helper to `testNode` or `testInstance` to call silence on the underlying logger. but this can be done whenever it comes up

done!

PR #162, opened 2025-05-06
indexFinalizationCertificates should not prematurely exit upon a non finalized block

samliok:


epoch_test.go
```diff
@@ -101,6 +102,60 @@ func TestEpochHandleNotarizationFutureRound(t *testing.T) {
 	require.Equal(t, secondBlock, blockCommitted)
 }
 
+// TestEpochIndexFinalizationCertificates ensures that we properly index past finalizations when
+// there has been empty rounds
```
yacovm, line 106: have been

yacovm (approved):

epoch.go
```diff
@@ -930,19 +930,28 @@ func (e *Epoch) rebroadcastPastFinalizations() error {
...
-	round, exists := e.rounds[r]
-	if !exists {
-		e.Logger.Debug("Round not found", zap.Uint64("round", r))
-		return
-	}
-	if round.fCert.Finalization.Seq != e.Storage.Height() {
-		e.Logger.Debug("Finalization certificate does not correspond to the next sequence to commit",
-			zap.Uint64("seq", round.fCert.Finalization.Seq), zap.Uint64("height", e.Storage.Height()))
-		return
+	maxRound := uint64(0)
```
yacovm, line 933: I just noticed that the title of the PR is not aligned with the bug fix, what do you think?

PR #163, opened 2025-05-06
rebroadcastPastFinalizations should not prematurely exit upon a round that is not found

samliok:


replication_test.go
```diff
@@ -176,12 +177,9 @@ func TestReplicationNotarizations(t *testing.T) {
 	net.Disconnect(laggingNode.e.ID)
 	numNotarizations := 9
 	missedSeqs := uint64(0)
-	blocks := []simplex.VerifiedBlock{}
 
 	// finalization for the first block
 	bb.triggerNewBlock()
-	block := <-bb.out
```
yacovm, line 183: why is this needed?

yacovm (commented):
can we add to `TestEpochNotarizeTwiceThenFinalize` an empty notarization in between the notarizations so we test this properly?

replication_test.go
samliok: we weren't using the value anyways

samliok:
> TestEpochNotarizeTwiceThenFinalize

done

yacovm (approved):

PR #168, opened 2025-05-08
Send notarization if we receive an empty vote for a previous round

samliok:


epoch_multinode_test.go
```diff
@@ -38,6 +38,90 @@ func TestSimplexMultiNodeSimple(t *testing.T) {
 	}
 }
 
+func onlySendBlockProposalsAndVotes(nodes []NodeID) messageFilter {
+	return func(m *Message, _, to NodeID) bool {
+		if m.BlockMessage != nil {
+			return true
+		}
+		if to.Equals(NodeID{3}) || to.Equals(NodeID{4}) {
```
yacovm, line 46: is it possible to not hardcode this?

epoch_multinode_test.go
```diff
@@ -38,6 +38,90 @@ func TestSimplexMultiNodeSimple(t *testing.T) {
...
+	}
+
+	net.setAllNodesMessageFilter(allowAllMessages)
+
+	emptyVoteListener := newEmptyVoteListener(splitNode2.e.ID)
+	splitNode2.e.Comm.(*testComm).setFilter(emptyVoteListener.filter)
+	waitForEmptyVote(t, emptyVoteListener.emptyVotes, splitNode2.e, 0, startTimes[2])
+	splitNode2.e.Comm = newTestComm(splitNode2.e.ID, net, allowAllMessages)
+
+	// splitNode3 will receives notarization from splitNode2
```
yacovm, line 102: `// splitNode3 will receives notarization from splitNode2` why?

epoch_multinode_test.go
```diff
@@ -38,6 +38,90 @@ func TestSimplexMultiNodeSimple(t *testing.T) {
...
+			require.Equal(t, uint64(1), n.e.Metadata().Round)
+		}
+	}
+
+	net.setAllNodesMessageFilter(allowAllMessages)
+
+	emptyVoteListener := newEmptyVoteListener(splitNode2.e.ID)
+	splitNode2.e.Comm.(*testComm).setFilter(emptyVoteListener.filter)
+	waitForEmptyVote(t, emptyVoteListener.emptyVotes, splitNode2.e, 0, startTimes[2])
+	splitNode2.e.Comm = newTestComm(splitNode2.e.ID, net, allowAllMessages)
```
yacovm, line 100: re-creating the communication while the node is running is not thread safe, as the node is dispatching requests in a different goroutine, and this replacement doesn't take place with a lock or anything similar.

Why are we replacing the communication? Can you elaborate?

yacovm (commented):
Generally LGTM, couple of questions though.

epoch_multinode_test.go
samliok: because splitNode2 will receive the notarization, then broadcast it to splitnode3

epoch_multinode_test.go
samliok: i wanted to call `waitForEmptyVote` and listen for an empty vote message to be sent by the node. the reason i cant just wait for the `wal` to contain an empty vote is because `splitNode2` might write to the wal & send the empty vote before the normal nodes have processed the notarization.

yacovm (approved):

PR #172, opened 2025-06-10
Consolidate Metadata and Blockheader Serialization

samliok:
We had many functions that performed the same serialization of the blockheader/protocol metadata structs. I moved out the common serialization into separate helper functions.

This helps AvalancheGo, where we re-implement this serialization when building blocks from bytes. It also allows a one-step switch to `asn1` instead of updating serialization in multiple places.


yacovm (approved):
LGTM, two not important nits so proceeding with merging

PR #173, opened 2025-06-11
Export Blockheader Length

samliok:
So we can pass in the correct amount of bytes into methods such as

`Finalization.FromBytes()`


yacovm:
I am not sure we have to know the block header length. We only need to know it if we serialize in a way that doesn't provide explicit field separation. 

But if we use a reflection based serialization scheme like ASN1 or a code generated one like protobuf, then I don't think we have to know the length.

PR #175, opened 2025-06-16
Update Bytes for VerifiedBlock to return error

samliok:
since we are serializing in AvalancheGo, this method _may_ error.


epoch.go
```diff
@@ -1897,7 +1904,11 @@ func (e *Epoch) proposeBlock(block VerifiedBlock) error {
 	// Write record to WAL before broadcasting it, so that
 	// if we crash during broadcasting, we know what we sent.
 
-	rawBlock := block.Bytes()
+	rawBlock, err := block.Bytes()
+	if err != nil {
```
yacovm, line 1908: why aren't we returning an error here?

yacovm (approved):

Issue #164, opened 2025-05-06
Ensure Resend Replication Requests when Response Doesn't Have an FCert


samliok:
If we now that a FinalizationCertificate should have been sent to us via a replication request we should make sure the response includes the fcert before marking that response as received.

PR #180, opened 2025-07-11
Resend ReplicationRequest If Finalization Is Expected But Not Received 

samliok:
Addresses #164

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -2442,8 +2442,6 @@ func (e *Epoch) locateQuorumRecord(seq uint64) *VerifiedQuorumRound {
 					Notarization:  round.notarization,
 				}
 			}
-
-			break
 		}
 	}
 
--- a/epoch.go
+++ b/epoch.go
@@ -2438,7 +2438,7 @@ func (e *Epoch) locateQuorumRecord(seq uint64) *VerifiedQuorumRound {
 					Finalization:  round.finalization,
 				}
 			} else if round.notarization != nil {
-				if round == nil {
+				if notarizedRound == nil {
 					notarizedRound = round
 				} else if round.notarization.Vote.Round > notarizedRound.num {
 					// set the notarized round if it is the highest round we have seen so far
--- a/epoch.go
+++ b/epoch.go
@@ -2428,6 +2428,7 @@ func (e *Epoch) handleReplicationRequest(req *ReplicationRequest, from NodeID) e
 
 // locateQuorumRecord locates a block with a notarization or finalization in the epochs memory or storage.
 func (e *Epoch) locateQuorumRecord(seq uint64) *VerifiedQuorumRound {
+	var notarizedRound *Round
 	for _, round := range e.rounds {
 		blockSeq := round.block.BlockHeader().Seq
 		if blockSeq == seq {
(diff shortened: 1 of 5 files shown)


replication_timeout_test.go
```diff
@@ -217,6 +219,24 @@ func incompleteReplicationResponseFilter(msg *simplex.Message, _, _ simplex.Node
...
 }
 
+// modifies the replication response to only send every other quorum round
+func removeFinalizationsFromReplicationResponses(msg *simplex.Message, _, _ simplex.NodeID) bool {
+	if msg.VerifiedReplicationResponse != nil || msg.ReplicationResponse != nil {
+		newData := make([]simplex.VerifiedQuorumRound, 0, len(msg.VerifiedReplicationResponse.Data))
+
+		for i := 0; i < len(msg.VerifiedReplicationResponse.Data); i++ {
+			qr := msg.VerifiedReplicationResponse.Data[i]
+			// to keep the request well-formed, we remove the finalization
```
yacovm, line 229: This test got me thinking - why do we even send both?

```
diff --git a/epoch.go b/epoch.go
index 10e54f0..6a65fec 100644
--- a/epoch.go
+++ b/epoch.go
@@ -2418,6 +2418,9 @@ func (e *Epoch) handleReplicationRequest(req *ReplicationRequest, from NodeID) e
                }
 
                data[i] = *quorumRound
+               if data[i].Finalization != nil {
+                       data[i].Notarization = nil
+               }
        }
 
        response.Data = data
```

With the above patch, the tests pass... shouldn't we only send the finalization if we have it? Why send the notarization too?

replication_timeout_test.go
```diff
@@ -285,3 +305,76 @@ func TestReplicationRequestIncompleteResponses(t *testing.T) {
...
+
+	// wait until the lagging nodes sends replication requests
+	time.Sleep(100 * time.Millisecond)
+	// due to the removeFinalizationsFromReplicationResponses message filter
+	// the lagging node should not have processed any replication requests
+	require.Equal(t, uint64(0), laggingNode.storage.Height())
+
+	// we should still have these replication requests in the timeout handler
+	laggingNode.e.AdvanceTime(laggingNode.e.StartTime.Add(simplex.DefaultReplicationRequestTimeout * 2))
+	time.Sleep(100 * time.Millisecond) // wait for the tick
```
yacovm, line 372: ditto

epoch.go
```diff
@@ -2431,13 +2431,16 @@ func (e *Epoch) locateQuorumRecord(seq uint64) *VerifiedQuorumRound {
...
-			return &VerifiedQuorumRound{
-				VerifiedBlock: round.block,
-				Notarization:  round.notarization,
-				Finalization:  round.finalization,
+			if round.finalization != nil {
+				return &VerifiedQuorumRound{
+					VerifiedBlock: round.block,
+					Finalization:  round.finalization,
+				}
+			} else if round.notarization != nil {
```
samliok, line 2439: so its possible we have multiple rounds in the rounds map with the same sequence. Technically if we return a notarization here, we could miss a sequence with a finalization later in the rounds map.

replication_timeout_test.go
```diff
@@ -285,3 +290,147 @@ func TestReplicationRequestIncompleteResponses(t *testing.T) {
...
+
+type collectNotarizationComm struct {
+	lock          *sync.Mutex
+	notarizations map[uint64]*simplex.Notarization
+	testNetworkCommunication
+
+	replicationResponses chan struct{}
+}
+
+func newCollectNotarizationComm(nodeID simplex.NodeID, net *inMemNetwork, notarizations map[uint64]*simplex.Notarization, lock *sync.Mutex) *collectNotarizationComm {
```
yacovm, line 302: Can we just move all the logic in these ad-hoc structs to a closure instantiated inside the test? 

It feels wrong to have so much logic defined just for one test.

replication_timeout_test.go
```diff
@@ -285,3 +290,147 @@ func TestReplicationRequestIncompleteResponses(t *testing.T) {
...
+func (c *collectNotarizationComm) Send(msg *simplex.Message, to simplex.NodeID) {
+	if msg.Notarization != nil {
+		c.lock.Lock()
+		c.notarizations[msg.Notarization.Vote.Round] = msg.Notarization
+		c.lock.Unlock()
+	}
+	c.testNetworkCommunication.Send(msg, to)
+}
+
+func (c *collectNotarizationComm) Broadcast(msg *simplex.Message) {
```
yacovm, line 320: you can just use the `recordingComm` we already have defined in `epoch_failover_test.go`. It fits what you're doing here perfectly.

yacovm (approved):

PR #182, opened 2025-07-15
Don't Expect Sorted Nodes From Communication Interface

samliok:


api.go
```diff
@@ -54,7 +54,7 @@ type Storage interface {
 
 type Communication interface {
 
-	// Nodes returns all nodes known to the application.
+	// Nodes returns all nodes that participate in the epoch.
```
samliok, line 57: https://github.com/ava-labs/avalanchego/pull/3998#discussion_r2207851765

PR #189, opened 2025-07-25
Skip notarization votes inconsistent to the block

yacovm:
This commit makes Simplex disregard votes that aren't related to the block received.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -2446,17 +2446,31 @@ func (e *Epoch) handleReplicationRequest(req *ReplicationRequest, from NodeID) e
 
 // locateQuorumRecord locates a block with a notarization or finalization in the epochs memory or storage.
 func (e *Epoch) locateQuorumRecord(seq uint64) *VerifiedQuorumRound {
+	var notarizedRound *Round
 	for _, round := range e.rounds {
 		blockSeq := round.block.BlockHeader().Seq
 		if blockSeq == seq {
-			if round.finalization == nil && round.notarization == nil {
-				break
+			if round.finalization != nil {
+				return &VerifiedQuorumRound{
+					VerifiedBlock: round.block,
+					Finalization:  round.finalization,
+				}
+			} else if round.notarization != nil {
+				if notarizedRound == nil {
+					notarizedRound = round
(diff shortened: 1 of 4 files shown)


notarization.go
```diff
@@ -48,6 +48,11 @@ func NewNotarization(logger Logger, signatureAggregator SignatureAggregator, vot
 
 	var toBeSignedVote *ToBeSignedVote
 	for _, vote := range votesForCurrentRound {
+		if !bytes.Equal(vote.Vote.Digest[:], blockHeader.Digest[:]) {
```
samliok, line 51: should we keep the same style as your other changes and check the block headers for equality?
```suggestion
		if vote.BlockHeader == blockHeader {
```

epoch_test.go
```diff
@@ -1158,6 +1158,93 @@ func TestMetadataProposedRound(t *testing.T) {
...
+				for _, signer := range msg.Notarization.QC.(testQC).Signers() {
+					require.NotEqual(t, nodes[1], signer, "Node 1 should not be in the notarization QC")
+				}
+				return true
+			}
+		default:
+			return false
+		}
+		return false
+	}, time.Second, time.Millisecond*10)
```
samliok, line 1244: these `require.Never` and `require.Eventually` feel very gross to me. I also suspect that they handle things asynchronously, which `assertions.Eventually` seems to do(it calls the condition function in a separate go routine). 
I added a time.Sleep(500ms) to the end of this test and it ended up failing, so i think thats the case.

Also don't we know the next broadcasted message should be a notarization? We should be able to grab the next message in the channel, or if not loop over the channel using a timer.

```go
for {
	select {
	case msg := <-recordedMessages:
		if msg.Notarization != nil {
			for _, signer := range msg.Notarization.QC.(testQC).Signers() {
					require.NotEqual(t, nodes[1], signer, "Node 1 should not be in the notarization QC")
			}
			break outer
		}
	case <-timer.C:
		t.Fatal("Timed out waiting for notarization")
	default:
		continue
	}
}
```

samliok (approved):

PR #190, opened 2025-07-25
Trigger empty block agreement upon invalid block

yacovm:
This commit makes nodes agreeing on the empty block once they fail to verify a block sent from the leader

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -1185,22 +1185,21 @@ func (e *Epoch) maybeCollectNotarization() error {
 		return e.maybeLoadFutureMessages()
 	}
 
-	// TODO: store votes before receiving the block
 	block := e.rounds[e.round].block
-	expectedDigest := block.BlockHeader().Digest
+	md := block.BlockHeader()
 
-	// Ensure we have enough votes for the same digest
-	var voteCountForOurDigest int
+	// Ensure we have enough votes for the same block header.
+	var voteCountForOurBlock int
 	for _, vote := range votesForCurrentRound {
-		if bytes.Equal(expectedDigest[:], vote.Vote.Digest[:]) {
-			voteCountForOurDigest++
+		if md == vote.Vote.BlockHeader {
+			voteCountForOurBlock++
 		}
 	}
 
-	if voteCountForOurDigest < e.quorumSize {
-		e.Logger.Warn("Counting votes for the digest we received from the leader",
+	if voteCountForOurBlock < e.quorumSize {
+		e.Logger.Warn("Counting votes for the block we received from the leader",
 			zap.Uint64("round", e.round),
-			zap.Int("voteForOurDigests", voteCountForOurDigest),
+			zap.Int("voteForOurBlock", voteCountForOurBlock),
 			zap.Int("total votes", voteCount))
 
 		// As a last resort, check if we have received a notarization message for this round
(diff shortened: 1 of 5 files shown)


epoch.go
```diff
@@ -1649,6 +1649,7 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
 		verifiedBlock, err := block.Verify(context.Background())
 		if err != nil {
 			e.Logger.Debug("Failed verifying block", zap.Error(err))
+			e.triggerEmptyBlockNotarization(md.Round, fmt.Sprintf("Invalid block: %s", err))
```
samliok, line 1652: nit: feel free to not update but I feel like its weird to add a `reason` param here. Especially since we log the error, then call `triggerEmptyBlockNotarization` and log the error again. Maybe we just add to the log above that we are triggering an empty block notarization?

samliok, line 1652: nit: feel free to ignore, but I feel like its weird to add a reason param here. Especially since we log the error, then call triggerEmptyBlockNotarization and log the error again. Maybe we just add to the log above that we are triggering an empty block notarization?

```suggestion
			e.Logger.Debug("Failed verifying block, triggering empty block notarization", zap.Error(err))
			e.triggerEmptyBlockNotarization(md.Round)
```

epoch_failover_test.go
```diff
@@ -657,6 +656,73 @@ func TestEpochLeaderFailoverTwice(t *testing.T) {
...
+
+	require.NoError(t, e.Start())
+
+	notarizeAndFinalizeRound(t, e, bb)
+
+	md := e.Metadata()
+	_, ok := bb.BuildBlock(context.Background(), md)
+	require.True(t, ok)
+	block := <-bb.out
+	block.verificationError = fmt.Errorf("invalid block")
```
samliok, line 698: nit: I think errors.New() is preferred if there is not formatting directives

epoch_failover_test.go
```diff
@@ -657,6 +656,73 @@ func TestEpochLeaderFailoverTwice(t *testing.T) {
 	})
 }
 
+func TestEpochLeaderFailoverBecauseOfBadBlock(t *testing.T) {
```
samliok, line 659: i think a small comment here would be nice, so that if we come back to this function down the line we can more quickly understand what the test does

util.go
```diff
@@ -195,9 +195,11 @@ func (block *oneTimeVerifiedBlock) Verify(ctx context.Context) (VerifiedBlock, e
 
 	vb, err := block.Block.Verify(ctx)
 
-	block.otv.digests[digest] = verifiedResult{
-		vb:  vb,
-		err: err,
+	if err == nil {
+		block.otv.digests[digest] = verifiedResult{
+			vb:  vb,
+			err: err,
```
samliok, line 201: whats the point of storing an err if it will always be nil? 

secondly i think it still makes sense to store the error, so that we don't  waste time verifying the block again. We just need to make sure we check the `verifiedResult` doesn't have an error before access fields in the `VerifiedBlock`

yacovm: Makes sense, done.

epoch_failover_test.go
yacovm: Done

epoch.go
yacovm: removed the parameter and moved the log to the caller functions

samliok (approved):

PR #205, opened 2025-08-01
Blacklist implementation

yacovm:
This commit implements a blacklist that performs book-keeping of nodes that are blacklisted. The blacklist contains a node count, and a bit vector the size of the node count.

When a node is blacklisted, its corresponding bit is set to 1 in the bit vector.

A node can be removed from the blacklist by being voted to be redeemed by f+1 nodes.

The blacklist keeps track of redemptions via a bit vector per blacklisted node.

If a more than f nodes are blacklisted, a random node is removed from the blacklist.

### DIFF
--- a/blacklist.go
+++ b/blacklist.go
@@ -1,0 +1,242 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package simplex
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"math/big"
+	"math/bits"
+)
+
+type Blacklist struct {
+	NodeCount        uint16
+	BlacklistedNodes bitVec
+	RedeemedNodes    []struct {
+		redeemedNode uint16
+		reporters    *bitVec
+	}
+}
+
+func (bl *Blacklist) blacklistSize() int {
+	return bits.OnesCount16(uint16(bl.BlacklistedNodes))
+}
+
+func (bl *Blacklist) IsBlacklisted(i uint16) bool {
+	if i < 0 || i >= bl.NodeCount {
+		panic(fmt.Sprintf("index out of range [0, %d): %d", bl.NodeCount, i))
+	}
+
+	return bl.BlacklistedNodes.IsSet(i)
+}
+
+func (bl *Blacklist) BlacklistNode(i uint16) error {
+	if i < 0 || i >= bl.NodeCount {
+		return fmt.Errorf("index out of range [0, %d): %d", bl.NodeCount, i)
+	}
+
+	bl.BlacklistedNodes.Set(i)
+
+	threshold := (bl.NodeCount - 1) / 3 // f
+
+	// If we have too many blacklisted nodes, we should not allow more,
+	// as it would lead to a situation where we cannot reach consensus.
+	// So we pick a pseudo-random node to redeem.
+	if bits.OnesCount16(uint16(bl.BlacklistedNodes)) > int(threshold) {
+		bl.redeemRandomNode()
+	}
+
+	return nil
+}
+
+func (bl *Blacklist) redeemRandomNode() {
+	digest := sha256.Sum256(bl.Bytes())
+	for {
+		n := big.NewInt(0).SetBytes(digest[:])
+		n.Mod(n, big.NewInt(int64(bl.NodeCount)))
+		if bl.BlacklistedNodes.IsSet(uint16(n.Int64())) {
+			bl.BlacklistedNodes.Clear(uint16(n.Int64()))
+			bl.compactRedeemedNodes()
+			break
+		}
+		// Else, the node is not blacklisted, so we need to pick a different one.
+		digest = sha256.Sum256(digest[:])
+	}
+}
+
+func (bl *Blacklist) RedeemNode(redeemedNodeIndex uint16, reporterNodeIndex uint16) error {
+	if (bl.NodeCount) == 0 {
+		return fmt.Errorf("blacklist node count is zero")
+	}
+
+	threshold := (bl.NodeCount-1)/3 + 1 // f + 1
+
+	if redeemedNodeIndex < 0 || redeemedNodeIndex >= bl.NodeCount {
+		return fmt.Errorf("redeemed node index out of range [0, %d): %d", bl.NodeCount, redeemedNodeIndex)
+	}
+
+	if reporterNodeIndex < 0 || reporterNodeIndex >= bl.NodeCount {
+		return fmt.Errorf("reporter node index out of range [0, %d): %d", bl.NodeCount, reporterNodeIndex)
+	}
+
+	if !bl.BlacklistedNodes.IsSet(redeemedNodeIndex) {
+		return nil
+	}
+
+	bl.markNodeReportedRedeemed(redeemedNodeIndex, reporterNodeIndex)
+
+	if bl.isNodeRedeemed(int(threshold), redeemedNodeIndex) {
+		bl.BlacklistedNodes.Clear(redeemedNodeIndex)
+		bl.compactRedeemedNodes()
+	}
+
+	return nil
+}
+
+func (bl *Blacklist) isNodeRedeemed(threshold int, redeemedNodeIndex uint16) bool {
+	for _, rn := range bl.RedeemedNodes {
+		if rn.redeemedNode == redeemedNodeIndex {
+			if bits.OnesCount16(uint16(*rn.reporters)) >= threshold {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func (bl *Blacklist) markNodeReportedRedeemed(redeemedNodeIndex uint16, reporterNodeIndex uint16) {
+	for _, rn := range bl.RedeemedNodes {
+		if rn.redeemedNode == redeemedNodeIndex {
+			rn.reporters.Set(reporterNodeIndex)
+			return
+		}
+	}
+
+	var zeroVec bitVec
+	bl.RedeemedNodes = append(bl.RedeemedNodes, struct {
+		redeemedNode uint16
+		reporters    *bitVec
+	}{
+		redeemedNode: redeemedNodeIndex,
+		reporters:    &zeroVec,
+	})
+
+	bl.RedeemedNodes[len(bl.RedeemedNodes)-1].reporters.Set(reporterNodeIndex)
+}
+
+func (bl *Blacklist) compactRedeemedNodes() {
+	newRedeemedNodes := make([]struct {
+		redeemedNode uint16
+		reporters    *bitVec
+	}, 0, len(bl.RedeemedNodes))
+	for _, rn := range bl.RedeemedNodes {
+		if bl.BlacklistedNodes.IsSet(rn.redeemedNode) {
+			newRedeemedNodes = append(newRedeemedNodes, rn)
+		}
+	}
+	bl.RedeemedNodes = newRedeemedNodes
+}
+
+func (bl *Blacklist) String() string {
+	redeemed := bytes.Buffer{}
+	for _, rn := range bl.RedeemedNodes {
+		redeemed.WriteString(fmt.Sprintf("{%d %d},", rn.redeemedNode, *rn.reporters))
+	}
+	redeemedStr := redeemed.String()
+	return fmt.Sprintf("NodeCount: %d, BlacklistedNodes: %d, RedeemedNodes: [%s]",
+		bl.NodeCount, bl.BlacklistedNodes, redeemedStr[0:len(redeemedStr)-1]) // remove trailing comma
+}
+
+func (bl *Blacklist) Bytes() []byte {
+	bytes := make([]byte, 4+4*len(bl.RedeemedNodes))
+	bytes[0] = byte(bl.NodeCount)
+	bytes[1] = byte(bl.NodeCount >> 8)
+	bytes[2] = byte(bl.BlacklistedNodes)
+	bytes[3] = byte(bl.BlacklistedNodes >> 8)
+
+	pos := 4
+
+	for _, bv := range bl.RedeemedNodes {
+		bytes[pos] = byte(bv.redeemedNode)
+		bytes[pos+1] = byte(bv.redeemedNode >> 8)
+		bytes[pos+2] = byte(*bv.reporters)
+		bytes[pos+3] = byte(*bv.reporters >> 8)
+		pos += 4
+	}
+
+	return bytes
+}
+
+func (bl *Blacklist) FromBytes(bytes []byte) error {
+	if len(bytes) < 4 {
+		return fmt.Errorf("blacklist bytes too short: %d < 4", len(bytes))
+	}
+
+	if len(bytes)%4 != 0 {
+		return fmt.Errorf("blacklist bytes length must be a factor of 4, got %d", len(bytes))
+	}
+
+	bl.NodeCount = uint16(bytes[0]) | uint16(bytes[1])<<8
+	bl.BlacklistedNodes = bitVec(uint16(bytes[2]) | uint16(bytes[3])<<8)
+
+	bytes = bytes[4:]
+
+	bl.RedeemedNodes = make([]struct {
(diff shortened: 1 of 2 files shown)


blacklist.go
```diff
@@ -0,0 +1,565 @@
...
+
+import (
+	"encoding/asn1"
+	"encoding/binary"
+	"fmt"
+	"slices"
+)
+
+// Orbit computes the number of rounds since genesis until and including the given round a node was the leader.
+func Orbit(round uint64, nodeIndex uint16, nodeCount uint16) uint64 {
```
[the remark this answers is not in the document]
yacovm: We can just reset the blacklist upon an epoch change.

blacklist_test.go
```diff
@@ -0,0 +1,670 @@
...
+		round             uint64
+		expectedErr       string
+	}{
+		{
+			name:      "node count mismatch",
+			nodeCount: 5,
+			proposedBlacklist: Blacklist{
+				NodeCount: 13,
+			},
+			expectedErr: "block contains an invalid blacklist node count, expected 5, got 13",
```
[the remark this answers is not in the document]
yacovm: But  then we can't have the extra information in the error.

blacklist.go
```diff
@@ -0,0 +1,565 @@
...
+		if u1.Type != u2.Type || u1.NodeIndex != u2.NodeIndex {
+			return false
+		}
+	}
+	return true
+}
+
+// SerializedBlacklist is used to transmit a blacklist over the wire.
+type SerializedBlacklist struct {
+	NewBlacklist []byte
```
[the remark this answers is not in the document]
yacovm: Removed the struct

blacklist.go
```diff
@@ -0,0 +1,565 @@
...
+
+	f := (bl.NodeCount - 1) / 3
+	threshold := f + 1
+
+	// Count how many total nodes are suspected. If we have f or more nodes suspected,
+	// then abort because we don't want the blacklist to be too big.
+
+	var suspectedCount int
+	for _, sn := range bl.SuspectedNodes {
+		// Node is already suspected.
```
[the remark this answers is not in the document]
yacovm: Fixed

blacklist_test.go
```diff
@@ -0,0 +1,829 @@
...
+		err := updates.FromBytes(data)
+		if err == nil {
+			require.Equal(t, data, updates.Bytes())
+			return
+		}
+		require.Error(t, err)
+	})
+}
+
+func TestSupectedNodesTrailer(t *testing.T) {
```
samliok, line 274:
```suggestion
func TestSuspectedNodesTrailer(t *testing.T) {
```

blacklist.go
```diff
@@ -0,0 +1,795 @@
...
+	copy(newBlacklist.Updates, updates)
+
+	return newBlacklist
+}
+
+// advanceRound advances the blacklist to the given round.
+// It will remove nodes that are no longer suspected or have been redeemed from the blacklist.
+// It will also garbage-collect any votes from past orbits to blacklist or redeem nodes, unless
+// they have surpassed the threshold of f+1.
+func (bl *Blacklist) advanceRound(round uint64) {
```
samliok, line 207: i thought about the logic here and I think it can be simplified a lot plus remove a bunch of for loops.

i created a branch if you want to see my thoughts on it. 
https://github.com/ava-labs/Simplex/compare/blacklist...blacklist-advance-round?expand=1

blacklist.go
```diff
@@ -0,0 +1,795 @@
...
+	threshold := bl.computeThreshold()
+
+	// Sanity check in case we get a too high node index.
+	// This shouldn't happen, but if it does, we should just ignore it.
+	if nodeIndex >= bl.NodeCount {
+		return
+	}
+	// First, look to see if the node is not suspected.
+	// If it is not, then we shouldn't try to redeem it.
+	if !bl.IsNodeSuspected(nodeIndex) {
```
samliok, line 340: so we can only redeem a node that is suspected? 

I'm thinking of this scenario

we get `threshold - 1` updates to suspect a node, and then get _over_ a `threshold` updates to redeem a node. Because the node is never suspected for the first `threshold-1` votes all the redeeming updates will get thrown out. Then when we get 1 more `suspected` update, our node will need to wait for a `threshold` of redeeming updates before it is unsuspecting. Is this fine or wasteful/potentially a problem?

or am i missing something?

blacklist.go
```diff
@@ -0,0 +1,795 @@
...
+		if sn.OrbitToRedeem < orbit {
+			newSN.RedeemingCount = 0
+			newSN.OrbitToRedeem = 0
+		}
+		// Update the suspected node
+		suspectedNodes[i] = newSN
+	}
+}
+
+func (bl *Blacklist) nodeSuspected(orbit uint64, nodeIndex uint16) {
```
samliok, line 264: small nit but I feel like `nodeSuspected` and `isNodeSuspected` are too similar. maybe `setNodeSuspected` or `markNodeSuspected`? fine with keeping it as is too

blacklist.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+	// First, look to see if the node is already suspected.
+	// If it is, then we shouldn't further increment its count.
+	if bl.IsNodeSuspected(nodeIndex) {
+		return
+	}
+
+	threshold := bl.computeThreshold()
+
+	var suspectedCount int
```
samliok, line 283: nit: setting this to `uint16` could save us a cast on line 312

blacklist.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+}
+
+func (bl *Blacklist) IsEmpty() bool {
+	return bl.NodeCount == 0
+}
+
+// ApplyUpdates applies the given updates in the given round to the current blacklist and returns a new blacklist.
+// The current blacklist is not modified.
+func (bl *Blacklist) ApplyUpdates(updates []BlacklistUpdate, round uint64) Blacklist {
```
samliok, line 178: this method pattern/flow seems a bit weird to me. Essentially we are creating a new `Blacklist` in this method and are expecting not to modify the receiver `bl`. 

We also have `bl.advanceRound` and `bl.advanceOrbit` which are only called during execution of `ApplyUpdates`. If we are just using `advanceRound` and `advanceOrbit` _only_ for the purpose of creating a new blacklist(given some parameters) shouldn't we not associate them with `bl` receiver?

blacklist.go
samliok: here is a pr i created for it 
https://github.com/ava-labs/Simplex/pull/252

samliok (commented):
going to pick up the rest of the review tomorrow, getting late for me

blacklist.go
yacovm: but why would you get a threshold of updates to redeem a node that isn't blacklisted? The purpose of the redeem is to redeem a node that is blacklisted. A node is blacklisted only if it was supposed to be the leader of a round but in that round an empty block was notarized. Therefore, it is only when a node is blacklisted that it can be redeemed, but that also means it cannot be blacklisted, so the threshold - 1 votes cannot occur until it finishes its redemption. 

Having said that, it matters not what happens within a single orbit, as in the next orbit if the node isn't blacklisted and is failed, then the node will be blacklisted in this orbit irrespectively of the previous orbit because when we move between orbits we garbage collect votes that haven't passed the threshold.

epoch.go
```diff
@@ -1284,6 +1320,7 @@ func (e *Epoch) handleEmptyNotarizationMessage(emptyNotarization *EmptyNotarizat
 
 	// Ignore votes for previous rounds
 	if !e.isVoteRoundValid(vote.Round) {
+		e.Logger.Debug("Empty notarization is invalid")
```
samliok, line 1323: if we are adding this debug log i think we should add vote.round and potentially e.round?

epoch.go
```diff
@@ -1912,7 +1956,40 @@ func (e *Epoch) locateBlock(seq uint64, digest []byte) (VerifiedBlock, bool) {
...
-	buildTheBlock := e.createBlockBuildingTask(metadata)
+	prevBlacklist, ok := e.retrieveBlacklistOfParentBlock(metadata)
+	if !ok {
+		return
+	}
+
+	// If I'm blacklisted, I cannot propose a block.
+	if prevBlacklist.IsNodeSuspected(uint16(e.nodes.IndexOf(e.ID))) {
+		e.Logger.Info("I'm blacklisted, cannot propose a block", zap.Uint64("round", metadata.Round))
+		e.triggerEmptyBlockNotarization(metadata.Round)
```
samliok, line 1967: hmmm, i think our current logic can cause us to trigger an `emptyBlockNotarization` twice. Once will be here, and the other time will be in `monitorProgress` which we will call right after `buildBlock`

blacklist.go
yacovm: Are you asking why there is asymmetry between modifying the blacklist in different instance methods?

epoch.go
```diff
@@ -2682,18 +2820,6 @@ func (e *Epoch) maybeAdvanceRoundFromEmptyNotarizations() (bool, error) {
 		}
 	}
 
-	// if there is no sequence, then maybe there is one with the same sequence but an empty notarization
```
samliok, line 2685: since the changes are unrelated should we remove this in a separate pr?

samliok (commented):
reviewed all files modulo tests

blacklist.go
samliok: i was trying to say the receiver methods `advanceRound` and `advanceOrbit` feel unnecessary but
i think my suggestions on https://github.com/ava-labs/Simplex/pull/252 make this comment not applicable anymore.

epoch.go
```diff
@@ -1172,11 +1176,33 @@ func (e *Epoch) persistEmptyNotarization(emptyNotarization *EmptyNotarization, s
 			zap.Uint64("round", emptyNotarization.Vote.Round))
 	}
 
+	err := e.maybeMarkLeaderAsTimedOutForFutureBlacklisting(emptyNotarization)
```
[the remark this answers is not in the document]
yacovm: what do you mean by updating the blacklist? You can only update the blacklist as a block proposer, which means you are not replicating.

epoch.go
```diff
@@ -1250,6 +1276,16 @@ func (e *Epoch) persistNotarization(notarization Notarization) error {
 		return err
 	}
 
+	round := notarization.Vote.Round
+	for _, signer := range notarization.QC.Signers() {
+		if signerIndex := e.nodes.IndexOf(signer); signerIndex != -1 {
+			e.Logger.Debug("Potentially redeeming node", zap.Stringer("signer", signer), zap.Uint64("round", round))
+			e.redeemedRounds[uint16(signerIndex)] = round
```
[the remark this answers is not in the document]
yacovm: In `ComputeBlacklistUpdates` we don't pay attention to redeemed nodes that are not suspected. 

It is done optimistically for simplicity.

blacklist.go
```diff
@@ -0,0 +1,784 @@
...
+	copy(newBlacklist.Updates, updates)
+
+	return newBlacklist
+}
+
+// garbageCollectSuspectedNodes returns a new list of suspected nodes for the given round.
+// Nodes that are no longer suspected or have been redeemed, will not be included in the returned suspected nodes.
+// It will also garbage-collect any redeem votes from past orbits, unless hey have surpassed the threshold of f+1.
+// It does not modify the current blacklist.
+func (bl *Blacklist) garbageCollectSuspectedNodes(round uint64) SuspectedNodes {
```
samliok, line 206: i think this method is a lot better + great name 🚀

blacklist_test.go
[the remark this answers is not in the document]
samliok: i'm still seeing the expected strings for this test

blacklist_test.go
```diff
@@ -0,0 +1,894 @@
...
+		},
+		{
+			name:        "too short data",
+			data:        []byte{0, 1, 2},
+			expectedErr: "buffer too short (3) to contain blacklist",
+		},
+		{
+			name:        "invalid suspected nodes",
+			data:        []byte{0, 4, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+			expectedErr: "failed to parse suspected nodes: failed to read suspected node at position 0: invalid bitmask: lower 4 bits must be zero, got 00000010",
```
samliok, line 240: i feel like having the expectedErr be strings makes things too hardcoded.  From what I've learned from AvalancheGo PRs, we should probably define these errors as it makes it easier to change the error message without causing tests to fail. 

Having said that, if we are just working on the repo I guess its fine to keep but I think eventually when we try and merge simplex into AvalancheGo someone will comment on this 😅

blacklist_test.go
```diff
@@ -0,0 +1,894 @@
...
+		t.Run(fmt.Sprintf("round %d", round), func(t *testing.T) {
+			blacklist := Blacklist{
+				NodeCount:      uint16(len(nodes)),
+				SuspectedNodes: testCases[round].suspectedNodesBefore,
+			}
+
+			suspectingNodes := blacklist.garbageCollectSuspectedNodes(round)
+
+			if testCases[round].noChange {
+				require.Equal(t, testCases[round].suspectedNodesBefore, suspectingNodes)
```
samliok, line 555: if `suspectingNodesBefore` is the same for all the tests should we just remove them from the `testCases` struct?

blacklist_test.go
```diff
@@ -0,0 +1,894 @@
...
+
+	blacklist = blacklist.ApplyUpdates([]BlacklistUpdate{{Type: BlacklistOpType_NodeSuspected, NodeIndex: 3}}, 8)
+	require.True(t, blacklist.IsNodeSuspected(3))
+}
+
+func TestBlacklistSimulateNetwork(t *testing.T) {
+	for nodes := 1; nodes <= 128; nodes *= 2 {
+		for rounds := 10; rounds <= 1000; rounds *= 10 {
+			t.Run(fmt.Sprintf("n=%d/r=%d", nodes, rounds), func(t *testing.T) {
+				testBlacklistSimulateNetwork(t, nodes, rounds)
```
samliok, line 680: we could run these in parallel for a speedup
```suggestion
                t.Parallel()
				testBlacklistSimulateNetwork(t, nodes, rounds)
```

blacklist_test.go
```diff
@@ -0,0 +1,894 @@
...
+			recoveredInRound: leaderRecoveredInRound,
+			crashedNodes:     crashedNodes,
+			nodeCount:        nodeCount,
+			prevBlacklist:    blacklist,
+		})
+	}
+}
+
+const (
+	noNodeCrashed = math.MaxUint16
```
samliok, line 755: this const is unused

blacklist_test.go
```diff
@@ -0,0 +1,894 @@
...
+	for nodes := 1; nodes <= 128; nodes *= 2 {
+		for rounds := 10; rounds <= 1000; rounds *= 10 {
+			t.Run(fmt.Sprintf("n=%d/r=%d", nodes, rounds), func(t *testing.T) {
+				testBlacklistSimulateNetwork(t, nodes, rounds)
+			})
+		}
+	}
+}
+
+func testBlacklistSimulateNetwork(t *testing.T, nodeCount int, rounds int) {
```
samliok, line 686: great test 🙌

epoch.go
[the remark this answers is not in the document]
samliok: by updating i meant the epoch variables related to the blacklist such as `e.timedOutRounds`

epoch.go
samliok: oh oops i missed the `return nil`

samliok (approved):
approved with some nits i'm not impartial towards that you can ignore if you want

blacklist_test.go
yacovm: cuts the time in 50%

epoch.go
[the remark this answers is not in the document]
yacovm: Ah, no I don't think we should. If we're replicating an empty notarization or a regular notarization then it means that `2f+1` other nodes created it, out of which `f+1` nodes are correct, so we should have enough threshold to suspect or redeem.

PR #208, opened 2025-08-05
Add Context to BlockBuilding

samliok:
Also renames `finishCtx`


epoch.go
```diff
@@ -172,7 +176,8 @@ func (e *Epoch) init() error {
 	e.sched = NewScheduler(e.Logger)
 	e.monitor = NewMonitor(e.StartTime, e.Logger)
 	e.cancelWaitForBlockNotarization = func() {}
-	e.finishCtx, e.finishFn = context.WithCancel(context.Background())
```
yacovm, line 175: can we revert this change? I know this sounds like Finnish context but `e.ctx` doesn't say what it's used for.

epoch.go
```diff
@@ -172,7 +176,8 @@ func (e *Epoch) init() error {
...
+	e.ctx, e.cancelCtx = context.WithCancel(context.Background())
+	e.blockBuilderCtx, e.blockBuilderCancelFunc = context.WithCancel(e.ctx)
```
yacovm, line 180: why initialize the builder context here and then overwrite it again? furthermore I think a context should be cancelled before we overwrite it so doesn't this create a leak?

epoch.go
```diff
@@ -516,7 +521,7 @@ func (e *Epoch) loadLastBlock() error {
 }
 
 func (e *Epoch) Stop() {
-	e.finishFn()
```
yacovm, line 519: can we revert this?

epoch.go
```diff
@@ -1959,7 +1964,8 @@ func (e *Epoch) buildBlock() {
 
 func (e *Epoch) createBlockBuildingTask(metadata ProtocolMetadata) func() Digest {
 	return func() Digest {
-		block, ok := e.BlockBuilder.BuildBlock(e.finishCtx, metadata)
+		e.blockBuilderCtx, e.blockBuilderCancelFunc = context.WithCancel(e.ctx)
+		block, ok := e.BlockBuilder.BuildBlock(e.blockBuilderCtx, metadata)
```
yacovm, line 1968: I think we need to `defer` cancel the context anyway, even if we fail building the block.

yacovm (approved):

PR #222, opened 2025-08-25
Fix data race when timing out upon bad blocks

yacovm:
This commit fixes a data race that happens when we time out on a block that cannot be verified. The triggering of the empty block notarization needs to take place under the lock.

The test is modified slightly to trigger the data race.


samliok (approved):
good catch!

Issue #214, opened 2025-08-11
Simplex Sending out Unnecessary Replication Requests


samliok:
Our current replication logic keeps track of a `highestSequenceObserved`. This gets initially set when we receive a `finalization` from a future round, but can be updated during the replication process if a node sends us a `LatestRound` with a higher sequence.

Setting `LatestRound` see [here](https://github.com/ava-labs/Simplex/blob/39740e5dddf56b2466e44ca3ae4afc6fac110a23/epoch.go#L2464)
Updating `highestSequenceObserved` with `LatestRound` see [here](https://github.com/ava-labs/Simplex/blob/39740e5dddf56b2466e44ca3ae4afc6fac110a23/epoch.go#L2617-L2618)

We assume `highestSequenceObserved`  keep tracks of the latest `Finalized` sequence and will therefore re-send replication requests until we get a finalization for that sequence. However, it's possible that we have actually caught up since the `LatestRound` sent prior was a notarization.

yacovm:
I think we should just mark whether the latest round is a finalization or a notarization and then if you have replicated a notarization of that round but not the finalization you simply won't retry.

samliok:
>  I think we should just mark whether the latest round is a finalization or a notarization and then if you have replicated a notarization of that round but not the finalization you simply won't retry.

saw we are at seq `10` and get a future finalization of seq `20`. then right after we get a notarization of `21` we wont resend requests for sequences 10-20

yacovm:
> > I think we should just mark whether the latest round is a finalization or a notarization and then if you have replicated a notarization of that round but not the finalization you simply won't retry.
> 
> saw we are at seq `10` and get a future finalization of seq `20`. then right after we get a notarization of `21` we wont resend requests for sequences 10-20

why?

Issue #187, opened 2025-07-22
Re-add ReplicationTask if verification fails


samliok:
Currently, we remove the replication task when we receive a replication response with a sequence we are expecting. 

https://github.com/ava-labs/Simplex/blob/6e87c4b242d78ca09e048003f503ee7fcdb566a8/replication.go#L191

However, if block verification fails or someone sent us a bad `Notarization` or `Finalization` we will not re-request.

PR #224, opened 2025-08-28
Resend replication requests for finalized blocks that fail to verify

samliok:
I could add this for notarizations, but it seems like we are not planning on re[-requesting notarizations](https://github.com/ava-labs/Simplex/issues/214#issuecomment-3232314089)?

This solves #187

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -1746,6 +1746,7 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block Block, finalization F
 
 		verifiedBlock, err := block.Verify(context.Background())
 		if err != nil {
+			e.Logger.Debug("Failed verifying block", zap.Error(err))
 			// if we fail to verify the block, we should re-add to request timeout
 			index := int(md.Round) % len(finalization.QC.Signers())
 			e.replicationState.sendRequestToNode(md.Seq, md.Seq, finalization.QC.Signers(), index)
@@ -1769,7 +1770,6 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block Block, finalization F
 			e.Logger.Error("Failed to index finalization", zap.Error(err))
 			return md.Digest
 		}
-		e.Logger.Info("Finalized block", zap.Uint64("round", md.Round), zap.Uint64("seq", md.Seq))
 		err = e.processReplicationState()
 
 		if err != nil {
--- a/epoch.go
+++ b/epoch.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"math/rand/v2"
 	"slices"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


yacovm:
> I could add this for notarizations, but it seems like we are not planning on re[-requesting notarizations](https://github.com/ava-labs/Simplex/issues/214#issuecomment-3232314089)?
> 
> This solves #187

I don't think we have this problem for notarizations, because we first [filter out](https://github.com/ava-labs/Simplex/blob/16717cf375f2db26a69019adb4c42399e6f9679e/epoch.go#L2551) the notarizations that are bad, don't we?

epoch.go
```diff
@@ -1746,7 +1746,9 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block Block, finalization F
 
 		verifiedBlock, err := block.Verify(context.Background())
 		if err != nil {
-			e.Logger.Debug("Failed verifying block", zap.Error(err))
+			// if we fail to verify the block, we should re-add to request timeout
+			index := int(md.Round) % len(finalization.QC.Signers())
```
yacovm, line 1750: why do we send the request to the same node all the time?

epoch.go
```diff
@@ -1767,6 +1769,7 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block Block, finalization F
 			e.Logger.Error("Failed to index finalization", zap.Error(err))
 			return md.Digest
 		}
+		e.Logger.Info("Finalized block", zap.Uint64("round", md.Round), zap.Uint64("seq", md.Seq))
```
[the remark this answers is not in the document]
samliok: yep, removing

epoch.go
samliok: ok i see what you're saying. Updated so its selecting a random index now

epoch.go
```diff
@@ -1747,6 +1748,9 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block Block, finalization F
 		verifiedBlock, err := block.Verify(context.Background())
 		if err != nil {
 			e.Logger.Debug("Failed verifying block", zap.Error(err))
+			// if we fail to verify the block, we re-add to request timeout
+			index := rand.Int() % len(finalization.QC.Signers())
```
yacovm, line 1752: this `rand` is fairly deterministic in tests, and also not random in production.

Can we just read 2 random bytes from crypto random and reduce it modulo the node count? It'll be close enough to random, while not uniform, but good enough.

samliok: updated

epoch.go
```diff
@@ -1747,6 +1749,16 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block Block, finalization F
 		verifiedBlock, err := block.Verify(context.Background())
 		if err != nil {
 			e.Logger.Debug("Failed verifying block", zap.Error(err))
+			// if we fail to verify the block, we re-add to request timeout
+			numSigners := int64(len(finalization.QC.Signers()))
+			index, err := rand.Int(rand.Reader, big.NewInt(numSigners))
```
yacovm, line 1754: I don't understand why sample a number from 0 to num signers. Don't we need to sample the number from `0` to `n-1` ?

yacovm (approved):

PR #226, opened 2025-09-14
Fix unit test TestEpochQCSignedByNonExistentNodes

yacovm:


util.go
```diff
@@ -34,9 +34,6 @@ func IsFinalizationValid(eligibleSigners map[string]struct{}, finalization *Fina
 	if !valid {
 		return false
 	}
-	if !valid {
```
samliok, line 37: lol why did we have this twice 😂 such a funny thing to not notice

util.go
```diff
@@ -34,9 +34,6 @@ func IsFinalizationValid(eligibleSigners map[string]struct{}, finalization *Fina
...
-		return false
-	}
 
 	return true
```
samliok, line 38: we dont need the if at all, could just return below
```suggestion
	return valid
```

samliok (approved):

PR #236, opened 2025-09-19
Align log levels and fail CI if warn/error is found

yacovm:


ut.sh
samliok, line 1: can we rename this to unit-test.sh or something more descriptive?

unit-tests.sh
yacovm: done

PR #240, opened 2025-09-23
Remove Shared Block Builder

samliok:
This PR removes the shared `BlockBuilder` from the testing helpers. Instead, the network and `testNodes` manage each nodes `BlockBuilder`. This makes the tests more precise and less prone to flakes.

### DIFF
--- a/epoch_multinode_test.go
+++ b/epoch_multinode_test.go
@@ -218,7 +218,10 @@ type testNode struct {
 
 // triggerBlockShouldBeBuilt signals this nodes block builder it is expecting a block to be built.
 func (t *testNode) triggerBlockShouldBeBuilt() {
-	t.bb.blockShouldBeBuilt <- struct{}{}
+	select {
+	case t.bb.blockShouldBeBuilt <- struct{}{}:
+	default:
+	}
 }
 
 func (t *testNode) Silence() {


replication_test.go
```diff
@@ -946,12 +925,12 @@ func advanceWithoutLeader(t *testing.T, net *inMemNetwork, bb *testControlledBlo
...
 	}
 
-	for range net.instances {
-		select {
-		case bb.blockShouldBeBuilt <- struct{}{}:
-		default:
-
+	for _, n := range net.instances {
+		leader := n.e.ID.Equals(simplex.LeaderForRound(net.nodes, n.e.Metadata().Round))
+		if leader || laggingNodeId.Equals(n.e.ID) {
```
yacovm, line 930: the leader now also tracks the round so we should remove this if.

epoch_multinode_test.go
```diff
@@ -212,8 +211,14 @@ type testNode struct {
...
-	l *testutil.TestLogger
-	t *testing.T
+	l  *testutil.TestLogger
+	t  *testing.T
+	bb *testControlledBlockBuilder
+}
+
+// triggerBlockShouldBeBuilt signals this nodes block builder it is expecting a block to be built.
+func (t *testNode) triggerBlockShouldBeBuilt() {
+	t.bb.blockShouldBeBuilt <- struct{}{}
```
yacovm, line 221: I think we should have a select here:

```
select {
    case  t.bb.blockShouldBeBuilt <- struct{}{}
    default:
```

Because if we ever move between rounds too quickly, we don't get to dispatch the monitor task, so we never need to read from this channel.

yacovm (approved):

PR #261, opened 2025-10-02
Check if round has progressed before calling monitorProgress

samliok:
1. we are building a block
2. while we build a block we get an empty notarization
  a .we increase the round,
  b. the block builder gets cancelled
  c. we call `monitorProgress` on the new round when we enter `startRound`
3. we call the `monitorProgress` in the buildBlock task after. this would cancel the current monitorProgress and monitors a stale round

therefore we should check if the round has increased while building a block and not call monitorProgress() if so


yacovm:
https://github.com/ava-labs/Simplex/pull/267 has a test :)

PR #274, opened 2025-10-09
Full Replication

### LINKED ISSUE #238: Replicate empty rounds that have no notarization / finalization with higher rounds

### LINKED ISSUE #247: Vote for Notarized Blocks during Replication
Keep Track of `highestFinalizedSeqObserved` and `highestNotarizedSequenceObserved`. This way we can broadcast our votes for notarizations during replication(if we know the sequence has not been finalized yet).

### LINKED ISSUE #254: Allow for full replication
Replication is currently kickstarted when a node receives a `finalization` for a future round. It then knows it is missing `x` amount of sequences so it requests them from other nodes. A node will respond with the requested sequences and also send its latest quorum round to ensure the requesting node has the most recent sequence. This works for finalizations, but is not enough since we aren't guaranteed to receive the most up-to-date notarization(without a finalization) or empty notarization. Furthermore, we need to kickstart the process of replication when we receive future `Notarizations` and `EmptyNotarizations` since the network might be stuck and waiting for the lagging node to complete the `QC`. 

To do so, we should extend our replication logic to distinguish between replicating `finalized` sequences and future rounds with `Notarizations/EmptyNotarizations`. By keeping track of the `latestFinalizedSeq` and `latestKnownRound`, our node also knows if it needs to vote on rounds while replicating(noted [here](https://github.com/ava-labs/Simplex/issues/247)). 

With this proposed logic, a lagging node would being replication when it receives a `EmptyNotarization`, `Finalization` or `Notarization` from a future round. In the case where a lagging node reconnects into a halted network, it will send an `emptyVote` message which will be responded to via https://github.com/ava-labs/Simplex/blob/2030b910c34ad2cb77b6b52cb6eb258ff55e19ba/epoch.go#L709. This logic should be updated to also send the `latestQuorumRound`.

samliok:
### Summary  
This PR finalizes the replication scheme by clearly distinguishing between **replicating rounds** and **replicating sequences**. It also refactors related components for clarity and reliability. Overall, the replication logic is much more straight-forward and simple which will hopefully lead to less bugs(a few small ones I have found while making the pr)

### Replication Overview  
In Simplex, when a node receives a `notarization`, `emptyNotarization`, or `finalization` for a round or sequence ahead of its current state, it indicates the node is behind. The process of catching up(retrieving missing rounds or sequences) is called **Replication**.  

The `replicationState` struct manages this process. Replication is triggered when:  
- A `notarization` or `emptyNotarization` for a higher round is received, or  
- A `finalization` for a future sequence arrives.  
- It can also be triggered when a lagging node sends an empty vote for an older round. In this case, the normal node will send its highest `notarization` and `finalization` to potentially trigger replication for the lagging node. 

The `replicationState` tracks missing **rounds** and **sequences**, resends requests as needed, and removes completed items. When a `QuorumRound` (notarization, empty notarization, or finalization) is received, it’s added to the relevant state. Finalizations always supersede earlier quorum rounds, allowing us to prune older states from both the rounds and sequence trackers.  

`processReplicationState` advances rounds or sequences by checking for available quorum rounds. If the next sequence is complete, it’s committed; otherwise, the function checks for the next round. Empty notarizations can sometimes be inferred to advance rounds via `maybeAdvanceRoundFromEmptyNotarizations`.

### Key Changes  
- **Introduced `replicator` struct**, which:  
  - Stores received `QuorumRounds` keyed by rounds or sequenc es.  
  - Limits requests to a maximum of `maxRoundWindow` ahead of the current position.  
  - Manages timeouts for missing quorum rounds.  
- **Simplified `TimeoutManager`:**  
  - As long as tasks exist in `taskMap`, the `TimeoutHandler` periodically sends them to the `TaskRunner` every `runInterval`.  
  - This greatly simplifies task management and scheduling logic.


testutil/node.go
```diff
@@ -106,3 +107,26 @@ func (t *TestNode) handleMessages() {
 		}
 	}
 }
+
+// TimeoutOnRound advances time until the node times out of the given round.
+func (t *TestNode) TimeoutOnRound(round uint64) {
+	startTime := t.E.StartTime
```
yacovm: I think it makes sense to have a time as part of the `TestNode` struct, otherwise you can only use this method properly once. When this method returns, the start time of the node is out of sync with the `E.StartTime` and you will start from the old start time the next time you invoke this method. Instead I think we should store the current time in the `TestNode` and when you update the time you also save it in the `TestNode`. This way, next time you invoke this method, you will fetch the updated start time.

epoch_failover_test.go
```diff
@@ -82,6 +82,7 @@ func TestEpochLeaderFailoverWithEmptyNotarization(t *testing.T) {
 		EmptyNotarization: emptyNotarization,
 	}, nodes[1])
 
+	fmt.Println("Sent empty notarization for round 2")
```
yacovm: we can remove this

replicator.go
```diff
@@ -0,0 +1,304 @@
+package simplex
```
yacovm: // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
// See the file LICENSE for licensing terms.

epoch.go
```diff
@@ -2242,23 +2336,32 @@ func (e *Epoch) triggerEmptyBlockNotarization(round uint64) {
...
+		e.Logger.Debug("Our empty vote not found in the set to rebroadcast, yet expected to rebroadcast", zap.Uint64("round", e.round))
+		return
 	}
 
-	e.timeoutHandler.AddTask(task)
+	e.Logger.Debug("Rebroadcasting empty vote because round has not advanced", zap.Uint64("round", ourVote.Vote.Round))
+	e.Comm.Broadcast(&Message{EmptyVoteMessage: ourVote})
+}
+
+func (e *Epoch) addEmptyVoteRebroadcastTimeout(vote *EmptyVote) {
```
yacovm:
```
func (e *Epoch) addEmptyVoteRebroadcastTimeout() {
```

replicator.go
```diff
@@ -0,0 +1,304 @@
...
+	// For example this may happen if our node has sent a finalized vote
+	// for this round and has not received the
+	// finalization from the network.
+	ss.signers = ss.signers.Remove(myNodeID)
+	return ss, nil
+}
+
+func newSignedRoundOrSeqFromFinalization(finalization *Finalization, myNodeID NodeID) *signerRoundOrSeq {
+	ss := &signerRoundOrSeq{
+		round:   finalization.Finalization.Round,
```
yacovm: why do we need the information about the round if we replicate a finalized seq?

replicator.go
```diff
@@ -0,0 +1,304 @@
...
+		if storedRound < newValue {
+			delete(r.receivedQuorumRounds, storedRound)
+		}
+	}
+}
+
+// sendReplicationRequests sends requests for missing sequences for the
+// range of sequences [start, end] <- inclusive. It does so by splitting the
+// range of sequences equally amount the nodes that have signed [highestSequenceObserved].
+func (r *replicator) sendReplicationRequests(start uint64, end uint64) {
```
yacovm: This was just moved over from `replication.go` right?

yacovm (commented):
Made a pass, will make another pass later.

replication.go
```diff
@@ -4,342 +4,157 @@
...
-	seqs := make([]uint64, (end+1)-start)
-	for i := start; i <= end; i++ {
-		seqs[i-start] = i
-	}
-	request := &ReplicationRequest{
-		Seqs:        seqs,
-		LatestRound: r.highestSequenceObserved.seq,
+	// otherwise we are storing a round without finalization
+	// don't bother storing rounds that are older than the highest finalized round we know
+	// todo: grab a lock for sequence replicator
```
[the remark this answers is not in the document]
samliok: ah i ended up using the epoch lock so i can just remove this todo

replicator.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	// next time we send requests, we start with a different permutation
+	r.requestIterator++
+}
+
+// sendRequestToNode requests the sequences [start, end] from nodes[index].
+// In case the nodes[index] does not respond, we create a timeout that will
+// re-send the request.
+func (r *replicator) sendRequestToNode(start uint64, end uint64, nodes []NodeID, index int) {
```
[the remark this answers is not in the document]
samliok: it would be a no-op though since the task is already in the scheduler. I can add a flag to `sendRequestToNode` such as `shouldAddTimeout` and only add the timeout if set.

replicator.go
```diff
@@ -0,0 +1,304 @@
...
+	if r.highestObserved != nil {
+		return r.highestObserved.round
+	}
+	return 0
+}
+
+// maybeSendMoreReplicationRequests checks if we need to send more replication requests given an observed round or sequence.
+// it limits the amount of outstanding requests to be at most [maxRoundWindow] ahead of [base] which is
+// either nextSeqToCommit or currentRound depending on if we are replicating sequences or rounds.
+func (r *replicator) maybeSendMoreReplicationRequests(observed *signerRoundOrSeq, base uint64) {
```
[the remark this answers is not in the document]
yacovm: `roundOrSequence` ?

replicator.go
```diff
@@ -0,0 +1,310 @@
...
+	val := observed.value()
+
+	// we've observed something we've already requested
+	if r.highestRequested >= val && r.highestObserved != nil {
```
yacovm: I am really concerned that we're going to compare apples to oranges if we compare a highest observed sequence to a round, and vice versa. 

`r.highestObserved` can be either a round or a sequence. 

Can this not cause an incorrect return?

replicator.go
```diff
@@ -0,0 +1,304 @@
...
+		return
+	}
+
+	// if this is the highest observed sequence or round, update our state
+	if r.highestObserved == nil || val > r.highestObserved.value() {
+		r.highestObserved = observed
+	}
+
+	startSeq := math.Max(float64(base), float64(r.highestRequested))
+	// we limit the number of outstanding requests to be at most maxRoundWindow ahead of nextSeqToCommit
```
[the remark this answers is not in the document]
yacovm: I get the memory concern, I am just wondering if it makes sense to use the same config parameter for both the standard execution flow and the replication one. 

I guess we can put them the same for now, maybe revisit later.

replicator.go
```diff
@@ -0,0 +1,310 @@
...
+		r.logger.Debug("Already requested observed value, skipping", zap.Uint64("value", val), zap.Bool("isRound", observed.isRound))
+		return
+	}
+
+	// if this is the highest observed sequence or round, update our state
+	if r.highestObserved == nil || val > r.highestObserved.value() {
+		r.highestObserved = observed
+	}
+
+	startSeq := math.Max(float64(base), float64(r.highestRequested))
```
yacovm: I don't understand - if `highestRequested` is a round, how can we treat it as a sequence? 

Can we perhaps maintain separate highestObservedX for rounds and sequences? I think it would make things more safer.

epoch.go
```diff
@@ -2832,27 +2987,35 @@ func (e *Epoch) processEmptyNotarization(emptyNotarization *EmptyNotarization) e
 	return e.processReplicationState()
 }
 
-func (e *Epoch) processLatestRoundReceived(latestRound *QuorumRound, from NodeID) error {
+// processQuorumRound processes a quorum round received from another node.
+// It verifies the quorum round and stores it in the replication state if valid.
+func (e *Epoch) processQuorumRound(latestRound *QuorumRound, from NodeID) error {
```
yacovm: Not related to this PR, but thought I should mention it.

If we replicate a notarized block $b_i$ that was only notarized by the node that sent us the block and all other nodes notarized the empty round, the next notarized block we receive after that $b_{i+1}$, can be a block building on top of the parent of $b_i$. 

Conversely, we can replicate an empty notarization while everyone else but the sender node has a regular notarization for the same round. Then we might get stuck, won't we?

yacovm (commented):
Will continue to review later, here are some more comments.

PR #283, opened 2025-10-20
[not to be merged] Missed Block Bug

samliok:
This relates to Yacov's closed PR, which we should probably re-open 😀. Maybe we should consider sending the notarization with the block? not saying we should, but it could be easier 🤷

https://github.com/ava-labs/Simplex/pull/264/files


samliok:
i think this can be closed since the nodes will send an empty vote after a timeout and they will all progress to a new round.

PR #285, opened 2025-10-23
Only verify blocks once their parent is notarized or finalized

yacovm:


epoch_test.go
```diff
@@ -32,6 +33,71 @@ var (
 	}
 )
 
+func TestBlockNotVerifiedIfParentNotNotarized(t *testing.T) {
```
samliok: can we also check that the block eventually gets verified once we get the notarization/finalization? either in this test or another

PR #293, opened 2025-11-03
Verify After Receiving Empty Notarizations

samliok:
This PR enforces that blocks are only verified if their parent digests are verified(notarized or finalized) as well as any empty notarizations that are in between the parent and the to be verified block.

It does so by adjusting the scheduler to wait for two dependencies to be resolved: the parent digest, and any missing empty notarizations. When empty notarizations get persisted, we notify the scheduler. To do so, we must allow empty notarizations from previous rounds to be processed and ensure we don't delete empty notarizations from memory prematurely. 

In a future PR, I will add the ability for a node to request these missing empty notarizations.

### DIFF
--- a/sched.go
+++ b/sched.go
@@ -218,12 +218,9 @@ func (d *dependencies) RemoveDigest(id Digest) []Task {
 
 	for _, task := range d.tasks {
 		var removed bool
-		if task.ParentBlockDependency == nil {
-			continue
-		}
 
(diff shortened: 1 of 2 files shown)


sched.go
samliok: I think the wording of `Dependencies` feels a little weird since `Dependencies` just have an array of tasks now.

PR #294, opened 2025-11-06
Simplify scheduling in presence of gaps of empty rounds

yacovm:


util.go
```diff
@@ -342,3 +343,91 @@ func (nt *NotarizationTime) CheckForNotFinalizedNotarizedBlocks(now time.Time) {
...
+		delete(erd, digest)
+	}
+}
+
+func (erd EmptyRoundsDependencies) MissingEmptyRounds() []uint64 {
+	missing := make(map[uint64]struct{})
+	for _, roundRanges := range erd { // Digest --> []RoundRange
+		for _, roundRange := range roundRanges { // []RoundRange
+			for round := range roundRange { // map[uint64]struct{}
+				missing[round] = struct{}{}
```
samliok: i think your original idea for this was to keep the RoundRanges as [start, end]?

PR #301, opened 2025-11-21
Implement PoS

yacovm:


pos_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"bytes"
+	"github.com/ava-labs/simplex"
+	"github.com/ava-labs/simplex/testutil"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestPoS(t *testing.T) {
```
samliok: would it be helpful to add a test which cuts off >60% of the stake and ensures it cannot make progress?

samliok (approved):
Approved with optional questions

pos_test.go
yacovm: Good call, done.

PR #302, opened 2025-11-24
Full Replication Logic Pt 2

samliok:
### Summary  
This PR finalizes(hopefully) the replication scheme by clearly distinguishing between **replicating rounds** and **replicating sequences**. It also refactors related components for clarity and reliability. Overall, the replication logic is much more explicit which will hopefully lead to less bugs(a few small ones I have found while making the pr)

### Replication Overview  
In Simplex, when a node receives a `notarization`, `emptyNotarization`, or `finalization` for a round or sequence ahead of its current state, it indicates the node is behind. The process of catching up(retrieving missing rounds or sequences) is called **Replication**.  

The `replicationState` struct manages this process. Replication is triggered when:  
- A `notarization` or `emptyNotarization` for a higher round is received, or  
- A `finalization` for a future sequence arrives.  
- It can also be triggered when a lagging node sends an empty vote for an older round. In this case, the normal node will send its highest `notarization` and `finalization` to potentially trigger replication for the lagging node. 

The `replicationState` tracks missing **rounds** and **sequences**, resends requests as needed, and removes completed items. When a `QuorumRound` (notarization, empty notarization, or finalization) is received, it’s added to the relevant state. Finalizations always supersede earlier quorum rounds, allowing us to prune older states from both the rounds and sequence trackers.  

`processReplicationState` advances rounds or sequences by checking for available quorum rounds. If the next sequence is complete, it’s committed; otherwise, the function checks for the lowest quorum round. Note that the lowest quorum round can be lower than the current round. This is because we may have advanced to the current round through replication through a "dead" chain. I.e. we may have received a notarization/empty notarization that other block builders decided not to build off of, yet we received it as a valid replication response.  Empty notarizations can sometimes be inferred to advance rounds via `maybeAdvanceRoundFromEmptyNotarizations`.

### Key Changes  
- **Introduced `requestor` struct**, which fetches quorum rounds up from the network.
- **Generic & Simple `TimeoutManager`:**  
  - As long as tasks exist in `taskMap`, the `TimeoutHandler` periodically sends them to the `TaskRunner` every `runInterval`.  
 - **Epoch Message Handling**
   - `notarization` and `emptyNotarization` messages are now able to be processed for previous, non-finalized rounds. 
  - **Quorum Round Validity**
    - Quorum rounds can now include both empty notarizations & notarizations. 

TODOS
- We need a message type to specifically request `Digests` from peers. This is important for replication but also the case where we receive a notarization for our current round but are yet to receive the block.

### DIFF
--- a/replication_test.go
+++ b/replication_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zapcore"
 
+	"github.com/ava-labs/simplex/record"
 	"github.com/ava-labs/simplex/testutil"
 	. "github.com/ava-labs/simplex/testutil"
 )
--- a/epoch_test.go
+++ b/epoch_test.go
@@ -1297,7 +1297,7 @@ func TestEpochVotesForEquivocatedVotes(t *testing.T) {
 // Ensures we don't double increment the round on persisting a notarization
 func TestDoubleIncrementOnPersistNotarization(t *testing.T) {
 	// add an empty notarization, then a notarization for a previous round
-	bb := &testutil.TestBlockBuilder{Out: make(chan *testutil.TestBlock, 1)}
+	bb := testutil.NewTestBlockBuilder()
 	nodes := []NodeID{{1}, {2}, {3}, {4}}
 	conf, wal, _ := testutil.DefaultTestNodeEpochConfig(t, nodes[3], testutil.NewNoopComm(nodes), bb)
 	conf.ReplicationEnabled = true
@@ -1319,7 +1319,7 @@ func TestDoubleIncrementOnPersistNotarization(t *testing.T) {
 	_, ok := bb.BuildBlock(context.Background(), md, emptyBlacklist)
 	require.True(t, ok)
 
-	block := <-bb.Out
+	block := bb.GetBuiltBlock()
 	notarization, err := testutil.NewNotarization(conf.Logger, conf.SignatureAggregator, block, nodes)
 	require.NoError(t, err)
 
--- a/pos_test.go
+++ b/pos_test.go
@@ -5,7 +5,6 @@ package simplex_test
 
 import (
 	"bytes"
-	"fmt"
 	"testing"
 	"time"
 
--- a/replication_request_test.go
+++ b/replication_request_test.go
@@ -215,7 +215,7 @@ func TestReplicationRequestMixed(t *testing.T) {
 }
 
 func TestReplicationRequestTailingEmptyNotarizations(t *testing.T) {
-	bb := &testutil.TestBlockBuilder{Out: make(chan *testutil.TestBlock, 1)}
+	bb := testutil.NewTestBlockBuilder()
 	nodes := []simplex.NodeID{{1}, {2}, {3}, {4}}
 	comm := NewListenerComm(nodes)
 	conf, wal, _ := testutil.DefaultTestNodeEpochConfig(t, nodes[0], comm, bb)
@@ -270,7 +270,7 @@ func TestReplicationRequestTailingEmptyNotarizations(t *testing.T) {
 }
 
 func TestReplicationRequestUnknownSeqsAndRounds(t *testing.T) {
-	bb := &testutil.TestBlockBuilder{Out: make(chan *testutil.TestBlock, 1)}
+	bb := testutil.NewTestBlockBuilder()
 	nodes := []simplex.NodeID{{1}, {2}, {3}, {4}}
 	comm := NewListenerComm(nodes)
 	conf, _, _ := testutil.DefaultTestNodeEpochConfig(t, nodes[0], comm, bb)


replication_state.go
```diff
@@ -0,0 +1,320 @@
...
+	r.finalizationRequestor.sendRequestToNode(seq, seq, signers[index.Int64()])
+	return nil
+}
+
+// CreateDependencyTasks creates tasks to refetch the given parent digest and empty rounds. If there are no
+// dependencies, no tasks are created.
+func (r *ReplicationState) CreateDependencyTasks(parent *Digest, parentSeq uint64, emptyRounds []uint64) {
+	if parent != nil {
+		r.digestTimeouts.AddTask(*parent)
+		r.seqsToDigests[parentSeq] = *parent
```
yacovm: where are you removing from the map? This will eventually make the memory blow up, no?

replication_state.go
```diff
@@ -0,0 +1,320 @@
...
+// deleteOldRounds cleans up the replication state for round replication after receiving a finalized round.
+func (r *ReplicationState) deleteOldRounds(finalizedRound uint64) {
+	for round := range r.rounds {
+		if round <= finalizedRound {
+			r.logger.Debug("Replication State Deleing Old Round", zap.Uint64("round", round))
+			delete(r.rounds, round)
+		}
+	}
+
+	r.roundRequestor.removeOldTasks(finalizedRound)
```
yacovm: what about `r.finalizationRequestor.removeOldTasks(finalizedRound)` ?  Why don't we need to do that too?

requestor.go
```diff
@@ -0,0 +1,256 @@
...
+		r.sendRequestToNode(seqsOrRounds.Start, seqsOrRounds.End, nodes[index])
+	}
+
+	r.requestIterator++
+}
+
+// maybeSendMoreReplicationRequests checks if we need to send more replication requests given an observed quorum.
+// it limits the amount of outstanding requests to be at most [maxRoundWindow] ahead of [currentSeqOrRound].
+func (r *requestor) maybeSendMoreReplicationRequests(observed *signedQuorum, currentSeqOrRound uint64) {
+	observedSeqOrRound := r.getSeqOrRound(observed)
```
yacovm: can we put a comment that we don't mix rounds and sequences because we have a different and mutually exclusive instance for each?

requestor.go
```diff
@@ -0,0 +1,256 @@
...
+			signers: NodeIDs(qr.Notarization.QC.Signers()).Remove(myNodeID),
+			round:   qr.Notarization.Vote.Round,
+			seq:     qr.Notarization.Vote.Seq,
+		}
+	default:
+		return nil
+	}
+}
+
+func newSingedQuorumFromFinalization(finalization *Finalization, nodeID NodeID) *signedQuorum {
```
yacovm: signed, not singed.

epoch.go
```diff
@@ -2854,72 +3026,72 @@ func (e *Epoch) processEmptyNotarization(emptyNotarization *EmptyNotarization) e
...
-func (e *Epoch) processLatestRoundReceived(latestRound *QuorumRound, from NodeID) error {
-	if latestRound == nil {
+// processQuorumRound processes a quorum round received from another node.
+// It verifies the quorum round and stores it in the replication state if valid.
+func (e *Epoch) processQuorumRound(round *QuorumRound, from NodeID) error {
+	if round == nil {
 		return nil
 	}
 
 	// make sure the latest round is well formed
```
yacovm: this is not the latest round, this is just a round

epoch.go
```diff
@@ -2854,72 +3026,72 @@ func (e *Epoch) processEmptyNotarization(emptyNotarization *EmptyNotarization) e
...
-		return err
+	if err := round.IsWellFormed(); err != nil {
+		return fmt.Errorf("received malformed latest round: %w", err)
 	}
 
-	if err := e.verifyQuorumRound(*latestRound, from); err != nil {
-		e.Logger.Debug("Received invalid latest round", zap.Error(err))
-		return err
+	if round.Finalization == nil && e.isVoteForFinalizedRound(round.GetRound()) {
+		return fmt.Errorf("received a quorum round for a too far behind. round: %d; seq: %d", round.GetRound(), round.GetSequence())
```
yacovm: "received a quorum round for a too far behind round".

But let's just state explicitly that the round has already been finalized.

epoch.go
```diff
@@ -2854,72 +3026,72 @@ func (e *Epoch) processEmptyNotarization(emptyNotarization *EmptyNotarization) e
 	return e.processReplicationState()
 }
 
-func (e *Epoch) processLatestRoundReceived(latestRound *QuorumRound, from NodeID) error {
-	if latestRound == nil {
+// processQuorumRound processes a quorum round received from another node.
+// It verifies the quorum round and stores it in the replication state if valid.
+func (e *Epoch) processQuorumRound(round *QuorumRound, from NodeID) error {
+	if round == nil {
```
yacovm: I think we might as well also just check if we have committed this round. Otherwise why bother processing it?

```
	if round.Finalization != nil && e.lastBlock != nil && e.lastBlock.VerifiedBlock.BlockHeader().Seq > round.Finalization.Finalization.Seq {
		return fmt.Errorf("received a finalized round for a committed sequence. round: %d; seq: %d", round.GetRound(), round.GetSequence())
	}
```

requestor.go
```diff
@@ -0,0 +1,256 @@
...
+	segments := CompressSequences(missingIds)
+	for i, seqsOrRounds := range segments {
+		index := (i + r.requestIterator) % numNodes
+		r.sendRequestToNode(seqsOrRounds.Start, seqsOrRounds.End, nodes[index])
+	}
+
+	r.requestIterator++
+}
+
+// maybeSendMoreReplicationRequests checks if we need to send more replication requests given an observed quorum.
```
yacovm: IIUC this is the only place where we send the requests so maybe this should be called `maybeSendReplicationRequests` ?

requestor.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// we've observed something we've already requested
+	if r.highestRequested >= observedSeqOrRound && r.highestObserved != nil {
+		r.logger.Debug("Already requested observed value, skipping", zap.Uint64("value", observedSeqOrRound), zap.Bool("Seq Replication", r.replicateSeqs))
+		return
+	}
+
+	// if this is the highest observed sequence, update our state
+	if r.highestObserved == nil || observedSeqOrRound > r.highestObserved.seq {
+		r.highestObserved = observed
```
yacovm: I would not expect a method named `maybeSendFoo` to have a side effect that updates the highest observed sequence/round. 

In all invocations of this method, we do it last. 

I'm not sure though how to refactor the code to make it less bloated overall but what do you think of this refactoring?

```

// maybeSendMoreReplicationRequests checks if we need to send more replication requests given an observed quorum.
// it limits the amount of outstanding requests to be at most [maxRoundWindow] ahead of [currentSeqOrRound].
func (r *requestor) maybeSendMoreReplicationRequests(observed *signedQuorum, currentSeqOrRound uint64) {
	observedSeqOrRound := r.getSeqOrRound(observed)


	// if this is the highest observed sequence, update our state
	if r.highestObserved == nil || r.highestObserved.seq < observedSeqOrRound {
		r.highestObserved = observed
	}

	// we've observed something we've already requested
	if r.highestRequested >= observedSeqOrRound && r.highestObserved != nil {
		r.logger.Debug("Already requested observed value, skipping", zap.Uint64("value", observedSeqOrRound), zap.Bool("Seq Replication", r.replicateSeqs))
		return
	}

	r.sendMoreReplicationRequests(currentSeqOrRound, observedSeqOrRound)
}

// maybeSendMoreReplicationRequests checks if we need to send more replication requests given an observed quorum.
// it limits the amount of outstanding requests to be at most [maxRoundWindow] ahead of [currentSeqOrRound].
func (r *requestor) sendMoreReplicationRequests(currentSeqOrRound uint64, observedSeqOrRound uint64) {
	start := math.Max(float64(currentSeqOrRound), float64(r.highestRequested))
	// we limit the number of outstanding requests to be at most maxRoundWindow ahead of nextSeqToCommit
	end := math.Min(float64(observedSeqOrRound), float64(r.maxRoundWindow+currentSeqOrRound))

	r.logger.Debug("Node is behind, attempting to request missing values", zap.Uint64("value", observedSeqOrRound), zap.Uint64("start", uint64(start)), zap.Uint64("end", uint64(end)), zap.Bool("seq requestor", r.replicateSeqs))
	r.sendReplicationRequests(uint64(start), uint64(end))
}
```

requestor.go
```diff
@@ -0,0 +1,256 @@
...
+	// ensure we don't have concurrent access to highestObserved
+	r.epochLock.Lock()
+	defer r.epochLock.Unlock()
+
+	nodes := r.highestObserved.signers
+	numNodes := len(nodes)
+	slices.Sort(missingIds)
+	segments := CompressSequences(missingIds)
+	for i, seqsOrRounds := range segments {
+		index := (i + r.requestIterator) % numNodes
```
yacovm: Would it make sense to export this for loop to a function and then just call it both here and in `sendReplicationRequests` ?

epoch.go
```diff
@@ -1691,15 +1777,31 @@ func (e *Epoch) processNotarizedBlock(block Block, notarization *Notarization) e
...
 		return e.processReplicationState()
 	}
 
 	// Create a task that will verify the block in the future, after its predecessors have also been verified.
 	task := e.createNotarizedBlockVerificationTask(e.oneTimeVerifier.Wrap(block), *notarization)
 	blockDependency, missingRounds := e.blockDependencies(md)
 
-	// TODO: if we have dependencies during replication, we are stuck since this means a node
-	// must have sent us an empty notarization for a round this block depends on.
+	e.replicationState.CreateDependencyTasks(blockDependency, md.Seq-1, missingRounds)
```
yacovm: This is a todo right?

replication_state.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	return blockWithFinalization.block, blockWithFinalization.finalization, true
+}
+
+func (r *ReplicationState) GetLowestRound() *QuorumRound {
+	var lowestRound *QuorumRound
+
+	for round, qr := range r.rounds {
+		fmt.Println("round", round)
```
yacovm: please clean up the Println

epoch.go
```diff
@@ -749,10 +752,60 @@ func (e *Epoch) handleEmptyVoteMessage(message *EmptyVote, from NodeID) error {
...
+		}
+		e.Logger.Debug("Node appears behind, sending them the highest empty notarized round", zap.Stringer("to", to), zap.Uint64("round", latestQR.EmptyNotarization.Vote.Round))
+		e.Comm.Send(msg, to)
+		return
+	}
+}
+
+// send notarization or finalization for this round as well
 func (e *Epoch) maybeSendNotarizationOrFinalization(to NodeID, round uint64) {
 	r, ok := e.rounds[round]
```
yacovm: what if the round is finalized but in the storage and not in memory?

epoch.go
```diff
@@ -1641,11 +1717,20 @@ func (e *Epoch) processFinalizedBlock(block Block, finalization Finalization) er
 		return e.processReplicationState()
 	}
 
+	blockDependency, missingRounds := e.blockDependencies(block.BlockHeader())
+	// because its finalized we don't care about empty rounds
+	if blockDependency != nil {
+		e.Logger.Error(
```
yacovm: Can't we somehow call `processFinalizedBlock` on out of order blocks and reach here while we don't have the parent block?

epoch.go
```diff
@@ -1691,15 +1777,31 @@ func (e *Epoch) processNotarizedBlock(block Block, notarization *Notarization) e
 			return nil
 		}
 
+		// if we haven't timed out on the round, send a finalized vote message
```
yacovm: why do we need to do this? I don't think this is mandatory. We re-broadcast dropped finalized votes if we have a tail of notarized blocks that's not finalized. We will eventually re-broadcast this after we finish replication, if it's needed.

epoch.go
```diff
@@ -2666,34 +2785,59 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
 	}
 	response := &VerifiedReplicationResponse{}
 
-	latestRound := e.getLatestVerifiedQuorumRound()
-
-	if latestRound != nil && latestRound.GetRound() > req.LatestRound {
-		response.LatestRound = latestRound
+	if req.LatestRound > 0 {
+		latestRound := e.getLatestVerifiedQuorumRound()
+		if latestRound != nil && latestRound.GetRound() > req.LatestRound {
```
yacovm: can't we just always assign the latest round?

epoch.go
samliok: i think its fine to not fetch from storage here. If the node needs the finalization, it will get it from the replication path. 

Plus we would need to traverse the storage for the `round` since we can only fetch by sequence numbers

requestor.go
```diff
@@ -0,0 +1,256 @@
...
+// requestor fetches quorum rounds up to [highestObserved] from the network,
+// allowing up to [maxRoundWindow] concurrent requests to limit memory use.
+// Ensures all rounds/sequences are eventually received.
+type requestor struct {
+	epochLock *sync.Mutex
+
+	// highestSequenceRequested prevents duplicates and limits outstanding requests.
+	highestRequested uint64
+
+	// the requestor stops requesting once all sequences/rounds up to an including `highestObserved` have been received.
```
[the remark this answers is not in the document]
samliok: `highestObserved` needs to be of `signedQuorum` because we use the signers to notify the requestor who to send requests to.

timeout_handler.go
```diff
@@ -113,72 +101,53 @@ func (t *TimeoutHandler) shouldRun() bool {
...
 
-	tasks, exists := t.tasks[nodeID]
-	if !exists {
-		return
-	}
-
-	for _, task := range tasks {
-		f(task)
+	for id := range t.tasks {
+		if t.shouldRemove(id, task) {
```
yacovm: This shouldRemove is awkward to use. Why not have it be a predicate on a specific task and just pass in a function? 

It's much cleaner this way:

```
func (t *TimeoutHandler[T]) RemoveOldTasks(shouldRemove func(id T, _ struct{}) bool) {
	t.lock.Lock()
	defer t.lock.Unlock()
	
	maps.DeleteFunc(t.tasks, shouldRemove)
}
```

And it's clear in the caller what we're trying to do:
```
	r.emptyRoundTimeouts.RemoveOldTasks(func(id uint64, _ struct{}) bool {
		return id <= finalizedRound
	})
```

As opposed to:

```
func shouldDelete(value, target uint64) bool {
	return value <= target
}
``` 

which lacks context and it's not clear what value and target are supposed to convey.

epoch.go
```diff
@@ -600,9 +597,9 @@ func (e *Epoch) handleFinalizationMessage(message *Finalization, from NodeID) er
 }
 
 func (e *Epoch) handleFinalizationForPendingOrFutureRound(message *Finalization, round uint64, nextSeqToCommit uint64) {
-	if round == e.round {
+	if round <= e.round {
```
yacovm: I don't have a real objection against this change, but can you tell what made you do this? Was there a flake you encountered or a test that fails? I ran the tests and they pass when reverting the change. 

Asking because the only way we're still verifying a proposal for a past round is if we've advanced a round because we received an empty notarization, but since we received a finalization, this is a contradiction.

epoch.go
```diff
@@ -2630,34 +2718,59 @@ func (e *Epoch) storeProposal(block VerifiedBlock) bool {
...
+	if req.LatestFinalizedSeq > 0 {
+		if e.lastBlock != nil && e.lastBlock.Finalization.Finalization.Seq > req.LatestFinalizedSeq {
+			response.LatestFinalizedSeq = &VerifiedQuorumRound{
+				VerifiedBlock: e.lastBlock.VerifiedBlock,
+				Finalization:  &e.lastBlock.Finalization,
+			}
+		}
 	}
 
 	seqs := req.Seqs
```
yacovm: can we check that we don't have more than `e.maxRoundWindow` sequences? Otherwise it's a DoS if we get a million sequences to retrieve.

epoch.go
```diff
@@ -2737,42 +2883,36 @@ func (e *Epoch) handleReplicationResponse(resp *ReplicationResponse, from NodeID
...
-		if data.GetSequence() > nextSeqToCommit+e.maxRoundWindow {
+		if data.Finalization != nil && data.GetSequence() > nextSeqToCommit+e.maxRoundWindow {
 			e.Logger.Debug("Received quorum round for a seq that is too far ahead", zap.Uint64("seq", data.GetSequence()))
 			// we are too far behind, we should ignore this message
 			continue
 		}
 
-		if err := e.verifyQuorumRound(data, from); err != nil {
-			e.Logger.Debug("Received invalid quorum round", zap.Uint64("seq", data.GetSequence()), zap.Stringer("from", from))
+		// TODO: if empty notarizations occur for long periods, we may receive a nextSeqToCommit that has a round considered too far ahead.
```
yacovm: this may happen just because we're simply really far behind, no?

epoch.go
```diff
@@ -1560,10 +1634,11 @@ func (e *Epoch) blockDependencies(bh BlockHeader) (*Digest, []uint64) {
 	prevBlock, notarizationOrFinalization, found := e.locateBlock(bh.Seq-1, bh.Prev[:])
 	if !found {
 		// should never happen since we check this when we verify the proposal metadata
-		e.Logger.Error("Could not find predecessor block for proposal scheduling",
+		e.Logger.Info("Could not find predecessor block for proposal scheduling",
```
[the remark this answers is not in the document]
samliok: because `processNotarizeBlock` may actually hit this code path. 
Ex. In round `2` we receive an empty round during replication  instead of a notarized one, and then process round `3` which relies on the notarized round for `2`. We don't have the predecessor block which is why we can end up here.

timeout_handler.go
samliok: i like this 💯. Added

epoch.go
samliok: yep, will do the same with rounds. we may want to use a smaller value because sending `e.maxRoundWindow` blocks + finalizations in one message could get big.  added

epoch.go
yacovm: > but we can still have a round that has an empty notarization & finalization?

We may have it only due to the fact that we finalize each block recursively.

But if a round was timed out by `f+1` correct nodes and an empty notarization was produced, this round is blocked from being finalized. 

We may later notarized and finalize a more advanced round and then we will recursively collect a finalization. 
However if that happens, it means we're behind, and we will replicate because we should receive the finalization with the higher round, no?

epoch.go
samliok: updated the comment

timeout_handler.go
```diff
@@ -4,72 +4,63 @@
...
-type TimeoutHandler struct {
-	lock sync.Mutex
-
-	ticks chan time.Time
-	close chan struct{}
-	// nodeids -> range -> task
-	tasks map[string]map[string]*TimeoutTask
-	heap  TaskHeap
+type timeoutRunner[T comparable] func(ids []T)
+type shouldRemoveFunc[T comparable] func(a T, b T) bool
```
yacovm: This is now unused

replication_test.go
```diff
@@ -1085,3 +1087,331 @@ func TestReplicationVerifyEmptyNotarization(t *testing.T) {
 		return wal.ContainsEmptyNotarization(0)
 	}, time.Millisecond*500, time.Millisecond*10, "Did not expect an empty notarization with a corrupt QC to be written to the WAL")
 }
+
+// almostFinalizeBlocks is a message filter that allows all messages except for finalized votes
```
yacovm: Can we move this method to the test it's used in, or alternatively, parametrize `simplex.NodeID{1}` ? It seems highly specific to the test it's used at, and therefore it should really be defined in the test.

epoch.go
```diff
@@ -747,10 +750,60 @@ func (e *Epoch) handleEmptyVoteMessage(message *EmptyVote, from NodeID) error {
...
 	r, ok := e.rounds[round]
 
 	if !ok {
+		// round could be an empty notarized round
+		evs, ok := e.emptyVotes[round]
+		if ok && evs.emptyNotarization != nil {
+			msg := &Message{
+				EmptyNotarization: evs.emptyNotarization,
+			}
+			e.Logger.Debug("Node appears behind, sending them an empty notarization", zap.Stringer("to", to), zap.Uint64("round", round))
```
yacovm: sending it

epoch.go
```diff
@@ -1327,7 +1381,9 @@ func (e *Epoch) persistNotarization(notarization Notarization) error {
 		}
 	}
 
-	e.increaseRound()
+	if notarization.Vote.Round == e.round && r.finalization == nil {
```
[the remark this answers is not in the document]
yacovm: I get why we need the notarization check, I just don't get why we need `r.finalization == nil`. 

Because if we have received a finalization for this round and didn't increase the round, we won't be increasing this round again.

epoch.go
```diff
@@ -2232,31 +2311,40 @@ func (e *Epoch) triggerEmptyBlockNotarization(round uint64) {
...
-		TaskID:   EmptyVoteTimeoutID,
-		Deadline: e.timeoutHandler.GetTime().Add(e.EpochConfig.MaxRebroadcastWait),
-		Task: func() {
-			e.Logger.Debug("Rebroadcasting empty vote because round has not advanced", zap.Uint64("round", vote.Vote.Round))
-			e.Comm.Broadcast(&Message{EmptyVoteMessage: vote})
-			e.addEmptyVoteRebroadcastTimeout(vote)
-		},
+func (e *Epoch) emptyVoteTimeoutTaskRunner(_ []string) {
+	e.lock.Lock()
+	roundVotes, ok := e.emptyVotes[e.round]
```
yacovm: we need to lock for the entirety of this function, not just here. 

What if `roundVotes.votes` is accessed concurrently?

epoch.go
```diff
@@ -2737,42 +2890,35 @@ func (e *Epoch) handleReplicationResponse(resp *ReplicationResponse, from NodeID
...
+		if data.Finalization != nil && data.GetSequence() > nextSeqToCommit+e.maxRoundWindow {
 			e.Logger.Debug("Received quorum round for a seq that is too far ahead", zap.Uint64("seq", data.GetSequence()))
 			// we are too far behind, we should ignore this message
 			continue
 		}
 
-		if err := e.verifyQuorumRound(data, from); err != nil {
-			e.Logger.Debug("Received invalid quorum round", zap.Uint64("seq", data.GetSequence()), zap.Stringer("from", from))
+		// We may be really far behind, so we shouldn't process sequences unless they are the nextSequenceToCommit
+		if data.GetRound() > e.round+e.maxRoundWindow && data.GetSequence() != nextSeqToCommit {
```
yacovm: I don't understand why would `data.GetSequence()` be necessary equal to the next sequence to commit?

PR #308, opened 2025-12-08
Request Block Digests if Needed

samliok:
This PR adds the `BlockDigestRequest` to Simplex. It allows blocks to be specifically requested which is important for nodes who may not have received them yet.

### DIFF
--- a/replication_test.go
+++ b/replication_test.go
@@ -253,6 +253,8 @@ func TestRebroadcastingWithReplication(t *testing.T) {
 // many empty notarizations in a row.
 // This test sometimes takes > 30 sec
 func TestReplicationEmptyNotarizations(t *testing.T) {
+	t.Skip()
+
 	nodes := []simplex.NodeID{{1}, {2}, {3}, {4}, {5}, {6}}
 
 	for endRound := uint64(2); endRound <= 2*simplex.DefaultMaxRoundWindow; endRound++ {
@@ -1222,6 +1224,7 @@ func TestReplicationVotesForNotarizations(t *testing.T) {
 // TestReplicationEmptyNotarizations ensures a lagging node will properly replicate
 // a tail of empty notarizations.
 func TestReplicationEmptyNotarizationsTail(t *testing.T) {
+	t.Skip()
 	nodes := []simplex.NodeID{{1}, {2}, {3}, {4}, {5}, {6}}
 
 	for endRound := uint64(2); endRound <= 2*simplex.DefaultMaxRoundWindow; endRound++ {


epoch.go
```diff
@@ -1033,6 +1035,7 @@ func (e *Epoch) persistFinalization(finalization Finalization) error {
 
 	e.blockVerificationScheduler.ExecuteBlockDependents(finalization.Finalization.Digest)
 	e.blockVerificationScheduler.RemoveOldTasks(finalization.Finalization.Seq)
+	e.replicationState.clearBlockDependencyTasks(finalization.Finalization.Digest, finalization.Finalization.Seq)
```
yacovm: why not remove everything before the sequence? `clearBlockDependencyTasks` only removes a specific sequence / digest.

replication_state.go
```diff
@@ -164,6 +182,9 @@ func (r *ReplicationState) StoreQuorumRound(round *QuorumRound) {
 	if round.EmptyNotarization != nil {
 		r.emptyRoundTimeouts.RemoveTask(round.GetRound())
 	}
+	if round.Block != nil {
+		r.digestTimeouts.RemoveTask(seqAndDigestFromBlock(round.Block))
```
yacovm: we're calling `RemoveTask` here, which is fine.

But we're also calling this in many other places. Isn't it redundant? 

We call it in `storeSequence` and in `storeRound` and in `persistFinalization` and in `persistNotarization`. 

I think we should call it here and also remove past tasks when we persist a finalization, and also call it when we persist a notarization. 

Thoughts?

replication_state.go
```diff
@@ -296,9 +326,41 @@ func (r *ReplicationState) GetBlockWithSeq(seq uint64) Block {
...
+	// grab the lock since this is called in the timeout handler goroutine
+	r.epochLock.Lock()
+	defer r.epochLock.Unlock()
+
+	signedQuorum := r.roundRequestor.getHighestObserved()
+	if signedQuorum == nil {
+		signedQuorum = r.finalizationRequestor.getHighestObserved()
+	}
+	if signedQuorum == nil {
+		r.logger.Warn("Replication State cannot request missing block digests, no known nodes to request from")
```
yacovm: we call `requestDigests` because we called `processNotarizedBlock`. 

Can we do this?

```
	e.replicationState.roundRequestor.observedSignedQuorum(&signedQuorum{
		round: notarization.Vote.Round,
		signers: notarization.QC.Signers(),
	}, notarization.Vote.Round)
	e.blockVerificationScheduler.ScheduleTaskWithDependencies(task, md.Seq, blockDependency, missingRounds)
```

Won't that eliminate the chances the `==` check will fail?

replication_state.go
```diff
@@ -296,9 +326,41 @@ func (r *ReplicationState) GetBlockWithSeq(seq uint64) Block {
...
+		return
+	}
+
```
yacovm: can we add a sanity check with an error log in case the signers is empty?

epoch.go
samliok: actually, i wonder if we can do that. Its possible we have two digests that need to be received, one for seq `3` and another for seq `4`. We could persist a notarization for seq `4` and with this logic remove the digest request for `seq 3`. What if we need `seq 3` in order to make progress in later rounds? 

Therefore, I think we should keep `clearBlockDependencyTasks` as is, and ensure to cleanup the digest requests in `deleteOldRounds`

replication_state.go
samliok: why not call it in `storeSequence`?

epoch_test.go
```diff
@@ -1203,6 +1203,79 @@ func TestEpochBlockTooHighRound(t *testing.T) {
 	})
 }
 
+// TestEpochSendsBlockDigestRequest ensures that we send a block digest request
+// when we receive a notarization for a block we don't have
+func TestEpochSendsBlockDigestRequest(t *testing.T) {
```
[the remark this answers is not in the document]
samliok: `TestReplicationChain`  checks this code path. if you comment out `CreateDependencyTasks` and run that test it should fail.

PR #309, opened 2025-12-09
Request Empty Rounds From Block Proposer

samliok:
Requests any empty notarization dependencies by sending a replication request to the block proposer.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -1820,6 +1820,13 @@ func (e *Epoch) createBlockVerificationTask(block Block, from NodeID, vote Vote)
 			return md.Digest
 		}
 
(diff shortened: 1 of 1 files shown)


yacovm (approved):
It's kind of annoying that is sends redundant requests after receiving the empty notarization:

```
[12-24|22:30:25.621] DEBUG Simplex/requestor.go:159 Node is behind, attempting to request missing values {"test": "TestEpochRequestsEmptyRoundDependency", "myNodeID": 4, "value": 1, "start": 2, "end": 1, "seq requestor": false}
```

But given that verification of blocks happens asynchronously, I don't see an easy way of fixing this. I guess we'll have to live with it 🤷

PR #315, opened 2025-12-29
Properly stop epoch instance

yacovm:


samliok (approved):
going to need to run CI again. `TestReplicationEmptyNotarizations` i think is an unrelated flake, i will look into it

PR #323, opened 2026-01-09
Pass In round to `maybeAssembleEmptyNotarization`

samliok:
`createBlockVerificationTask` is run and scheduled asynchronously from the main Epoch thread. This is leads to a potential error if we have advanced the round in between scheduling the task and running it.

This scenario leads to us halting the epoch. 

1. schedule block verification for round 0.
2. advance round (say from empty notarization for round 0)
3. begin block verification for round 0.
4. block verification fails so we call `triggerEmptyBlockNotarization`.
5. `triggerEmptyBlockNotarization` adds the empty vote for round 0, but then calls `maybeAssembleEmptyNotarization`.
6. `maybeAssembleEmptyNotarization` uses the epoch round and tries to assemble an empty notarization for round 1.
7. round 1 does not exist in the empty round map so we error.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -2328,6 +2328,12 @@ func (e *Epoch) metadata() ProtocolMetadata {
 }
 
 func (e *Epoch) triggerEmptyBlockNotarization(round uint64) {
+	if e.round > round {
+		e.Logger.Debug("Not triggering empty block notarization because we advanced to a higher round",
+			zap.Uint64("round", round), zap.Uint64("currentRound", e.round))
+		return
+	}
+
 	emptyVote := ToBeSignedEmptyVote{EmptyVoteMetadata: EmptyVoteMetadata{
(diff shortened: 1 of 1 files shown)


epoch.go
```diff
@@ -2410,12 +2416,6 @@ func (e *Epoch) monitorProgress(round uint64) {
 		e.lock.Lock()
 		defer e.lock.Unlock()
 
-		// Check if we have advanced to a higher round in the meantime while this task was dispatched.
```
yacovm: But without this, we will print:

```
		e.Logger.Debug("Triggering empty block agreement",
			zap.String("reason", "Timed out on block agreement"),
			zap.Uint64("round", round),
			zap.Stringer("leader", leader))
```

and this is confusing. Can we keep this?

yacovm (approved):

PR #328, opened 2026-01-14
Wal Recovery Bugs

samliok:
This PR fixes several issues in WAL recovery and round resumption logic that could lead to incorrect state restoration and stalled consensus.

**Incorrect WAL resume point**

Previously, resumeFromWal assumed the last appended WAL record was always the correct record to resume from. This assumption is invalid because WAL entries may be appended out of order (e.g., higher-round records written before lower-round ones).

As a result, we could resume from an incorrect record, leading to unintended behavior. For example, if the last record is a notarization for a lower round, we may call startRound twice for the same round(once before a crash and once after) potentially producing conflicting blocks. Similar issues can occur with empty notarizations and finalization records.

This PR ensures we resume from the correct WAL record rather than blindly using the last appended entry.

**EmptyVoteRecord resume bug**

When resuming from an EmptyVoteRecord, we were resending the empty vote but failing to create a timeout for it. If the empty vote message is dropped (which could be likely during recovery), we would never resend it. This can stall consensus if other nodes are waiting on our empty vote to complete an empty notarization. The fix ensures proper timeout handling so empty votes are retried as expected.

**Missing startRound / progress monitoring after notarization**

In some notarization cases, we were not calling startRound, which could prevent progress. Maybe other nodes are requiring us to time out and send an empty vote message. 

> Note: In `resumeFromWal` when highestRoundRecord.block != nil, we may need to explicitly monitor progress; otherwise, we may never send an empty vote when other nodes are waiting on us.

### DIFF
--- a/recovery_test.go
+++ b/recovery_test.go
@@ -614,6 +614,10 @@ func TestRecoveryReVerifiesBlocks(t *testing.T) {
 	require.Len(t, deserializer.DelayedVerification, 0)
 }
 
+// TestWalRecoveryTriggersEmptyVoteTimeout tests that when recovering from a wal
+// we properly broadcast an EmptyVote and create a timeout to rebroadcast it.
+// Without this timeout, other nodes may be stuck waiting for an empty notarization that
+// can only be completed if they receive our empty vote.
 func TestWalRecoveryTriggersEmptyVoteTimeout(t *testing.T) {
 	bb := testutil.NewTestBlockBuilder()
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


recovery_test.go
```diff
@@ -613,3 +613,332 @@ func TestRecoveryReVerifiesBlocks(t *testing.T) {
...
+	require.True(t, ok)
+	bBytes, err := block.Bytes()
+	require.NoError(t, err)
+	blockRecord := BlockRecord(block.BlockHeader(), bBytes)
+
+	// write block blockRecord to wal
+	require.NoError(t, wal.Append(blockRecord))
+
+	// lets add some notarizations
+	notarizationRecord, err := testutil.NewNotarizationRecord(conf.Logger, conf.SignatureAggregator, block, nodes[0:quorum])
```
yacovm: why do we need this notarization?

recovery_test.go
```diff
@@ -725,3 +725,220 @@ func TestWalRecoveryMonitorsProgress(t *testing.T) {
 		}
 	}
 }
+
+func TestWalRecoverySetsRoundCorrectly(t *testing.T) {
```
yacovm: good test 👍

recovery_test.go
samliok: dont need it, i just wanted to advance the round for this test.

yacovm (approved):

PR #330, opened 2026-01-20
Use configurable random source in epoch

samliok:


epoch.go
```diff
@@ -183,6 +186,17 @@ func (e *Epoch) HandleMessage(msg *Message, from NodeID) error {
...
 	e.maybeAssignDefaultConfig()
+
+	// Initialize random number generator
+	var seed int64
+	if e.RandomSeed != nil {
+		seed = *e.RandomSeed
+	} else {
+		seed = time.Now().UnixNano()
+	}
+	source := rand.NewPCG(uint64(seed), uint64(seed))
```
yacovm: It's better to use a 32 byte seed with ChaCha8, which has a bigger initial random seed (32 bytes vs 16 bytes) and also since it is a real cipher, it is more "unpredictable". 

```
	source := rand.NewChaCha8(seed)
```

For the tests we can still use a uint64 seed, but I think it's better that internally we'll have a 32 byte seed. We can just fill the top 8 bytes.

yacovm (approved):

PR #341, opened 2026-02-09
Broadcast Replication Request On Startup

samliok:
This PR adds a broadcasting a replication request when a node starts in order to check if it has synced up with the network. If we are behind, we will get a replication response that will kick start the replication process. If we are not behind, then other nodes will disregard our request.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -273,6 +273,7 @@ func (e *Epoch) Start() error {
 	}
 	// Only init receiving messages once you have initialized the data structures required for it.
 	defer func() {
+		e.broadcastReplicationSync()
 		e.canReceiveMessages.Store(true)
 	}()
 	return e.restoreFromWal()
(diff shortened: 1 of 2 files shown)


epoch.go
```diff
@@ -473,10 +473,36 @@ func (e *Epoch) loadFinalizationRecord(r []byte) error {
...
+		},
+	}
+	e.Comm.Broadcast(replicationRequest)
+}
+
 // resumeFromWal resumes the epoch from the records of the write ahead log.
 func (e *Epoch) resumeFromWal(highestRoundRecord *walRound) error {
 	e.Logger.Info("Most relevant record recovered from WAL", zap.Uint64("round", highestRoundRecord.round), zap.Stringer("relevant", highestRoundRecord))
 
+	e.broadcastReplicationSync()
```
yacovm: If we move it here:

```
func (e *Epoch) Start() error {
	if e.canReceiveMessages.Load() {
		return ErrAlreadyStarted
	}
	// Only init receiving messages once you have initialized the data structures required for it.
	defer func() {
		e.canReceiveMessages.Store(true)
		e.broadcastReplicationSync()
	}()
	return e.restoreFromWal()
}
```


We won't need to rely on the WAL to have messages in it, in order for this to be invoked.

epoch.go
```diff
@@ -273,6 +273,7 @@ func (e *Epoch) Start() error {
 	}
 	// Only init receiving messages once you have initialized the data structures required for it.
 	defer func() {
+		e.broadcastReplicationSync()
```
yacovm: why not move it after can receive messages?

yacovm (approved):

PR #350, opened 2026-02-17
Prioritize new tasks over older ones

samliok:


sched.go
```diff
@@ -82,4 +82,27 @@ func (as *BasicScheduler) Schedule(task Task) {
 	}
 }
 
+// ScheduleOrReplace schedules a task, replacing any queued task if full.
+// The cancel function is called before draining the old task.
+func (as *BasicScheduler) ScheduleOrReplace(task Task, cancel func()) {
```
yacovm: why not just invoke `e.blockBuilderCancelFunc` outside, in the epoch?

samliok: yea we do, removed

yacovm (approved):

Issue #352, opened 2026-03-03
Non Validating Nodes Medium Level Design Doc


samliok:
# Supporting Non-Validator Nodes

This medium-level design document outlines the objectives, background, and key considerations for supporting non-validator nodes. It also proposes a clear implementation approach to achieve the desired functionality and outcomes.

**What is a non-validator?**

A non-validator is a node that does not participate in consensus but still receives consensus messages in order to track the current state of the network. 

Examples include:

-   **API nodes**, which serve information about confirmed transactions and blocks.
    
-   **Full nodes**, which maintain the blockchain state at specific heights but do not vote in consensus.
    

Although these nodes do not contribute to consensus, they must remain fully synchronized with finalized blocks and validator set changes.

## Goal

Simplex currently supports validators through the `Epoch` implementation and will soon support a [multi-epoch configuration](https://github.com/ava-labs/avalanchego/blob/ed0b2ad3d1e058c48ce94ab8312cd9a1caa0ef8f/simplex/docs/reconfiguration.md#L1).

However, running an instance of `Epoch` requires participating in consensus, which non-validator nodes are not required to do. At a high level, non-validator nodes should: Sync and replicate only finalized blocks and their corresponding finalization messages. Non-validator nodes must also synchronize finalizations spanning multiple epochs (i.e., blocks signed by different validator sets).

Per the reconfiguration spec,  the epoch number of a block is equal to the block sequence of the sealing block from the previous epoch.

Given a block, the sealing block that sealed the previous epoch can be located by checking the epoch number of the block.

A sealing block of an epoch encodes the validator set of the next epoch.

## Bootstrapping

At startup:

-   The non-validator is configured with a set of bootstrapping NodeIDs.
    
-   It queries these nodes for their latest sealing block by fetching the latest block and determining the sealing block's sequence.
    
-   Once it receives at least `f + 1` identical sealing blocks, it can safely treat that sealing block as correct.

## Recursively Fetching Sealing Blocks

This provides a backward pointer mechanism:
-   The sealing block of an epoch contains the sequence number of the previous sealing block.
-   Using this pointer, we can recursively fetch prior sealing blocks
-   This continues until we reach the ever first Simplex block, which its predecessor is either a genesis block or the last Snowman block.

At this stage, the node reconstructs the full chain of sealing blocks across epochs in memory(sealing block hashes or full block).

## Fetching Blocks Within Epochs

[...]

PR #358, opened 2026-03-13
Non Validators Happy Path

samliok:


nonvalidator/non_validator.go
```diff
@@ -0,0 +1,241 @@
...
+func (n *NonValidator) handleBlock(block simplex.FullBlock, from simplex.NodeID) error {
+	bh := block.BlockHeader()
+
+	if n.latestEpoch.epoch != bh.Epoch {
+		n.Logger.Debug("Received a block not from the current epoch", zap.Uint64("Current epoch", n.latestEpoch.epoch), zap.Uint64("Epoch", bh.Epoch), zap.Stringer("From", from))
+	}
+
+	// TODO: do we need this check for non-validators? I added because otherwise we would need to store all blocks received until we receive a finalization
+	// but this could be a dos since a malicious node can spam us with bad blocks.
+	if !bytes.Equal(simplex.LeaderForRound(n.latestEpoch.validators, bh.Round), from) {
```
samliok: do we need this check for non-validators? I added because otherwise we would need to store all blocks received until we receive a finalization but this could be a dos since a malicious node can spam us with bad blocks.

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,241 @@
+package nonvalidator
```
yacovm, line 4: license header

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,241 @@
...
+	n.broadcastLatestEpoch()
+}
+
+// TODO: Broadcast the last known epoch to bootstrap the node. Collect responses marking the latest seal block
+// Keep rebroadcasting requests for that sealing block until we have enough responses to verify.
+func (n *NonValidator) broadcastLatestEpoch() {
+
+}
+
+func skipMessage(msg *simplex.Message) bool {
```
yacovm: isn't it easier to write when we do not skip the message?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,241 @@
...
+	case msg.FinalizeVote != nil:
+		return true
+	default:
+		return false
+	}
+}
+
+func (n *NonValidator) HandleMessage(msg *simplex.Message, from simplex.NodeID) error {
+	switch {
+	case skipMessage(msg):
```
yacovm: but why do we need this `skipMessage`? can't we just skip the message in the `default`?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,252 @@
...
+func (n *NonValidator) handleFinalization(finalization *simplex.Finalization, from simplex.NodeID) error {
+	bh := finalization.Finalization.BlockHeader
+
+	if bh.Epoch < n.latestVerifiedEpoch.epoch {
+		n.Logger.Debug("Received a finalization from an earlier epoch", zap.Uint64("Current epoch", n.latestVerifiedEpoch.epoch), zap.Uint64("Epoch", bh.Epoch), zap.Stringer("From", from))
+		return nil
+	}
+
+	// verify the finalization
+	if err := finalization.Verify(); err != nil {
```
yacovm: this just checks the signature. What about checking that the signers constitute a quorum of the validator set?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,252 @@
...
+
+type epochMetadata struct {
+	validators []simplex.NodeID // the validators of this epoch
+	epoch      uint64
+}
+
+type Config struct {
+	Logger            simplex.Logger
+	Storage           simplex.FullStorage
+	GenesisValidators []simplex.NodeID
```
yacovm: why genesis? the genesis validators might be long gone

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,252 @@
...
+		n.Logger.Debug("Stored incomplete sequence", zap.Stringer("Sequence", incompleteSeq))
+		return nil
+	}
+
+	// Duplicate finalization received.
+	if incomplete.finalization != nil {
+		// sanity check: should never happen.
+		if !bytes.Equal(incomplete.finalization.Finalization.Bytes(), finalization.Finalization.Bytes()) {
+			n.Logger.Warn(
+				"Mismatching finalizations",
```
yacovm, line 248: this should be an error that will halt the chain

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	// No block received yet for this sequence.
+	if incomplete.block == nil {
+		return nil
+	}
+
+	digest := incomplete.block.BlockHeader().Digest
+	if !bytes.Equal(bh.Digest[:], digest[:]) {
+		// TODO: this means the leader has equivocated and sent us a wrong block while another has been finalized.
```
yacovm, line 268: Just delete the block in such a case. The finalization takes precedence, right?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,252 @@
...
+		n.Logger.Info(
+			"Received a block from the leader of a round whose digest mismatches the finalization",
+			zap.Stringer("Finalization Digest", bh.Digest),
+			zap.Stringer("Block digest", digest),
+			zap.Stringer("From", from),
+		)
+		return nil
+	}
+
+	return n.indexBlock(incomplete.block, finalization)
```
yacovm: if we index the block, can't we garbage collect from the incompleteSeqs map?

nonvalidator/verifier.go
```diff
@@ -0,0 +1,96 @@
...
+
+// alreadyVerifiedSeq checks if `seq` has already been verified.
+func (v *Verifier) alreadyVerifiedSeq(seq uint64) bool {
+	return seq < v.nextSeqToVerify()
+}
+
+// triggerVerify wakes up the verifier by attempting to verify the next seq to be verified.
+// We verify `block` if it is the next sequence to be verified, otherwise we try to retrieve the next
+// block to be verified from storage.
+func (v *Verifier) triggerVerify(block simplex.Block) error {
```
yacovm: We don't really need this asynchronous scheduling. We can just verify() and index right after.

yacovm (commented):
I made a pass. Generally there is still a lot more to do here, as you noted in TODOs. 

I think we should add to the PR a markdown document. Essentially moving the [issue](https://github.com/ava-labs/Simplex/issues/352) you opened to markdown can be a good start.

PR #360, opened 2026-03-23
Reconfig

yacovm:


msm/fake_node_test.go
```diff
@@ -0,0 +1,432 @@
...
+	// Build a new block and check that the node has transitioned to the new epoch,
+	// rather than building a block on top of the old epoch.
+	height := node.Height()
+
+	for node.Height() == height {
+		node.act()
+	}
+	require.Greater(t, node.Epoch(), uint64(1))
+
+	return
```
samliok: why do we return here?

yacovm: oops, artifact from debugging 🤦‍♂️ 

Fixed.

msm/msm_test.go
```diff
@@ -0,0 +1,1054 @@
...
+	"github.com/ava-labs/simplex/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+type outerBlock struct {
+	finalization *simplex.Finalization
+	block        metadata.StateMachineBlock
+}
+
+type blockStore map[uint64]*outerBlock
```
samliok, line 25: can we store helper structs like this in their own files? alongside functions like `newStateMachine()`?

it's hard to review since im bouncing around these large files trying to separate/review the helpers from the tests

msm/msm_test.go
```diff
@@ -0,0 +1,1054 @@
...
+
+func (bs blockStore) clone() blockStore {
+	newStore := make(blockStore)
+	for k, v := range bs {
+		newStore[k] = v
+	}
+	return newStore
+}
```
samliok, line 33:
```suggestion
func (bs blockStore) clone() blockStore {
	newStore := make(blockStore)
	maps.Copy(newStore, bs)
	return newStore
}
```

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/360/commits/53742ffb7d15b4f5cdcdaa8cdd67e498a6257a25

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+		sm.Logger.Error("Error building block, building block without inner block instead", zap.Error(err))
+	}
+	if impatientContext.Err() != nil {
+		sm.Logger.Debug("Timed out waiting for block to be built, building block without inner block instead",
+			zap.Duration("elapsed", time.Since(start)), zap.Duration("maxBlockBuildingWaitTime", sm.MaxBlockBuildingWaitTime))
+	}
+	return sm.wrapBlock(parentBlock, childBlock, simplexEpochInfo, pChainHeight, simplexMetadata, simplexBlacklist), nil
+}
+
+func (sm *StateMachine) createSealingBlock(ctx context.Context, parentBlock StateMachineBlock, simplexMetadata []byte, simplexBlacklist []byte, simplexEpochInfo SimplexEpochInfo, newApprovals *approvals, pChainHeight uint64) (*StateMachineBlock, error) {
```
samliok, line 671: newApprovals is unused

msm/encoding.go
```diff
@@ -0,0 +1,347 @@
...
+	return cloned
+}
+
+func (nbms NodeBLSMappings) TotalWeight() (uint64, error) {
+	return nbms.SumWeights(func(int, NodeBLSMapping) bool {
+		return true
+	})
+}
+
+func (nbms NodeBLSMappings) ForEach(selector func(int, NodeBLSMapping)) {
```
samliok, line 266: why can't we just do 

```go
for i, nbm := range NodeBLSMappings {
		
}
```

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+	// TimeSkewLimit is the maximum allowed time difference between a block's timestamp and the current time.
+	TimeSkewLimit time.Duration
+	// GetTime returns the current time.
+	GetTime func() time.Time
+	// ComputeICMEpoch computes the ICM epoch for the next block.
+	ComputeICMEpoch ICMEpochTransition
+	// GetPChainHeight returns the latest known P-chain height.
+	GetPChainHeight func() uint64
+	// GetUpgrades returns the current upgrade configuration.
+	GetUpgrades func() UpgradeConfig
```
samliok, line 133: im confused by this function. The only time we set it is in a test, and it is set to a noop. If it is not necessary now, can we remove it from this pr?

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+		newSimplexEpochInfo.NextPChainReferenceHeight = pChainHeight
+		return sm.wrapBlock(parentBlock, nil, newSimplexEpochInfo, pChainHeight, simplexMetadata, simplexBlacklist), nil
+	case blockBuildingDecisionContextCanceled:
+		return nil, ctx.Err()
+	default:
+		return nil, fmt.Errorf("unknown block building decision %d", decisionToBuildBlock)
+	}
+}
+
+func (sm *StateMachine) createBlockBuildingDecider(parentBlock StateMachineBlock) blockBuildingDecider {
```
samliok, line 436: instead of having `sm.createBlockBuildingDecider()` i think i t would be better to have `newBlockBuildeingDecider(sm, parentBlock)`? I think it would make it easier to reason about its initialization and move the code to a more relevant area

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+			if !currentValidatorSet.Equal(newValidatorSet) {
+				return true, nil
+			}
+			return false, nil
+		},
+	}
+	return blockBuildingDecider
+}
+
+func (sm *StateMachine) buildBlockAndMaybeTransitionEpoch(ctx context.Context, parentBlock StateMachineBlock, simplexMetadata []byte, simplexBlacklist []byte, childBlock VMBlock, decisionToBuildBlock blockBuildingDecision, newSimplexEpochInfo SimplexEpochInfo, pChainHeight uint64) (*StateMachineBlock, error) {
```
samliok, line 468: this is such a large function signature 😬

samliok: also we are passing in a `childBlock VMBlock` but child block seems to always be nil?

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+
+	switch currentState {
+	case stateFirstSimplexBlock:
+		return sm.buildBlockZero(ctx, parentBlock, simplexMetadataBytes, simplexBlacklistBytes)
+	case stateBuildBlockNormalOp:
+		return sm.buildBlockNormalOp(ctx, parentBlock, simplexMetadataBytes, simplexBlacklistBytes, prevBlockSeq)
+	case stateBuildCollectingApprovals:
+		return sm.buildBlockCollectingApprovals(ctx, parentBlock, simplexMetadataBytes, simplexBlacklistBytes, prevBlockSeq)
+	case stateBuildBlockEpochSealed:
+		return sm.buildBlockEpochSealed(ctx, parentBlock, simplexMetadataBytes, simplexBlacklistBytes, prevBlockSeq)
```
samliok, line 244: no action just a thought: we have this pattern of passing `parentBlock, simplexMetadataBytes, simplexBlacklistBytes` and drilling these params all the way down to all the methods. I think it clutters the function signatures, im wondering if there is a better way for this

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+type BlockBuilder interface {
+	BuildBlock(ctx context.Context, pChainHeight uint64) (VMBlock, error)
+
+	// WaitForPendingBlock returns when either the given context is cancelled,
+	// or when the VM signals that a block should be built.
+	WaitForPendingBlock(ctx context.Context)
+}
+
+// StateMachine manages block building and verification across epoch transitions.
+type StateMachine struct {
```
samliok, line 118: should we make a constructor for this? `NewStateMachine` and pass in a config? seems dangerous to not have one

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+
+func (sm *StateMachine) maybeInit() {
+	if sm.initialized {
+		return
+	}
+	sm.init()
+	sm.initialized = true
+}
+
+func (sm *StateMachine) init() {
```
samliok, line 298: what is the benefit of having an `init` and `maybeInit`? is it possible that `VerifyBlock` or `BuildBlock` gets called concurrently, in that case we could have a pretty bad race since we might call init() twice? if we wanted to go down the lazy init route should we use `sync.Once` instead?

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+
+	// verifiers is the list of verifiers used to verify proposed blocks.
+	// Each verifier is responsible for verifying a specific aspect of the block's metadata.
+	verifiers   []verifier
+}
+
+type state uint8
+
+const (
+	stateFirstSimplexBlock state = iota
```
samliok, line 165:
```suggestion
	stateFirstSimplexBlock state = iota + 1
```
should we make this change to ensure we are initializing it? same with `BlockType`?

samliok: it would probably make more sense to have this for `BlockType` since it is exported

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+)
+
+type BlockType uint8
+
+const (
+	BlockTypeNormal BlockType = iota
+	BlockTypeTelock
+	BlockTypeSealing
+	BlockTypeNewEpoch
+)
```
samliok, line 178: should we move block type to where it is used mainly? i.e in `verification.go`? furthermore, i think we `IdentifyBlockType` can be moved maybe to a util file, just to not clutter the `msm.go` file.

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+	if decisionToBuildBlock == blockBuildingDecisionBuildBlockAndTransitionEpoch {
+		// We need to also transition to a new epoch, in addition to building an inner block,
+		// so set the next P-chain reference height for the new epoch info.
+		newSimplexEpochInfo.NextPChainReferenceHeight = pChainHeight
+	}
+
+	return sm.wrapBlock(parentBlock, childBlock, newSimplexEpochInfo, pChainHeight, simplexMetadata, simplexBlacklist), nil
+}
+
+func IdentifyBlockType(nextBlockMD StateMachineMetadata, prevBlockMD StateMachineMetadata, prevSeq uint64) BlockType {
```
samliok: why is this exported? also if this is a self contained function, i think we should have tests for it.

msm/msm_test.go
```diff
@@ -0,0 +1,1054 @@
...
+			blockTime := lastBlock.InnerBlock.Timestamp().Add(time.Second)
+
+			var icmEpochInvokeCount int
+
+			sm1.ComputeICMEpoch = func(_ any, input metadata.ICMEpochInput) metadata.ICMEpoch {
+				icmEpochInvokeCount++
+				require.Equal(t, metadata.ICMEpochInput{
+					ParentPChainHeight: 100,
+					ChildTimestamp:     blockTime,
+					ParentTimestamp:    time.Unix(int64(lastBlock.Metadata.Timestamp), 0),
```
samliok: we read this value using `time.UnixMilli` so this needs to be `time.UnixMilli`

msm/encoding.go
```diff
@@ -0,0 +1,347 @@
...
+	for i, nbm := range nbms {
+		selector(i, nbm)
+	}
+}
+
+func (nbms NodeBLSMappings) SumWeights(selector func(int, NodeBLSMapping) bool) (uint64, error) {
```
samliok, line 272: im confused by this? why are we creating this function if the only time we use it selector always returns true? isn't it just easier to do a simple for loop?

msm/encoding.go
```diff
@@ -0,0 +1,347 @@
...
+	var total uint64
+	var err error
+	nbms.ForEach(func(i int, nbm NodeBLSMapping) {
+		if err != nil {
+			return
+		}
+		if selector(i, nbm) {
+			total, err = safeAdd(total, nbm.Weight)
```
samliok, line 280: why not just return the error right away?

this function feels so ai generated 😅 so much unnecessary complexity for a simple for loop + sum counter.  Unless i am missing something?

msm/encoding.go
```diff
@@ -0,0 +1,347 @@
...
+	AuxInfoSeqDigest [32]byte `canoto:"fixed bytes,2"`
+	PChainHeight     uint64   `canoto:"uint,3"`
+	Signature        []byte   `canoto:"bytes,4"`
+
+	canotoData canotoData_ValidatorSetApproval
+}
+
+type ValidatorSetApprovals []ValidatorSetApproval
+
+func (vsa ValidatorSetApprovals) ForEach(f func(int, ValidatorSetApproval)) {
```
samliok, line 321: same here

msm/encoding.go
```diff
@@ -0,0 +1,347 @@
...
+	seen := make(map[nodeID]struct{})
+	result := make(ValidatorSetApprovals, 0, len(vsa))
+	vsa.ForEach(func(i int, v ValidatorSetApproval) {
+		if _, exists := seen[v.NodeID]; !exists {
+			seen[v.NodeID] = struct{}{}
+			result = append(result, v)
+		}
+	})
+	return result
+}
```
samliok, line 347: needs a new line at the end

msm/encoding.go
```diff
@@ -0,0 +1,347 @@
...
+		return slices.Compare(a.NodeID[:], b.NodeID[:])
+	})
+
+	for i := range nbmsClone {
+		if !nbmsClone[i].Equals(&otherClone[i]) {
+			return false
+		}
+	}
+	return true
+}
```
samliok, line 308:
```suggestion
 func (nbms NodeBLSMappings) Equal(other NodeBLSMappings) bool {                                                                                                                                                                                                                                          
        if len(nbms) != len(other) {                                                                                                                                                                                                                                                                       
                return false                                                                                                                                                                                                                                                                               
        }                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                           
        sortByNodeID := func(a, b NodeBLSMapping) int {                                                                                                                                                                                                                                                    
                return slices.Compare(a.NodeID[:], b.NodeID[:])                                                                                                                                                                                                                                            
        }                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                           
        nbmsClone := nbms.Clone()                                                                                                                                                                                                                                                                       
        otherClone := other.Clone()                                                                                                                                                                                                                                                                        
        slices.SortFunc(nbmsClone, sortByNodeID)                                                                                                                                                                                                                                                           
        slices.SortFunc(otherClone, sortByNodeID)                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                           
        return slices.EqualFunc(nbmsClone, otherClone, func(a, b NodeBLSMapping) bool {                                                                                                                                                                                                                    
                return a.Equals(&b)                                                                                                                                                                                                                                                                        
        })                                                                                                                                                                                                                                                                                                 
  }    
```

also why don't we just keep these sets sorted? potentially we could require `GetValidatorSet` to return a sorted array, or create a wrapper like `GetSortedValidatorSet` which would save us a bunch of slices.SortFunc() calls as `equal` can be called multiple times

msm/encoding.go
```diff
@@ -0,0 +1,347 @@
...
+type ValidatorSetApproval struct {
+	NodeID           nodeID   `canoto:"fixed bytes,1"`
+	AuxInfoSeqDigest [32]byte `canoto:"fixed bytes,2"`
+	PChainHeight     uint64   `canoto:"uint,3"`
+	Signature        []byte   `canoto:"bytes,4"`
+
+	canotoData canotoData_ValidatorSetApproval
+}
+
+type ValidatorSetApprovals []ValidatorSetApproval
```
samliok, line 319: why is this an array and not a map with the nodeIDs being keys? seems like `UniqueByNodeID` would benefit from this.

msm/encoding.go
samliok: also we could just store the `NodeBLSMappings` as a map and then assert equality in linear time rather than sorting both functions and then iterating(O(nlogn) * 2 + n) ?

msm/msm.go
yacovm: It's needed by ICM epochs

msm/msm.go
yacovm: This package isn't supposed to be used concurrently

msm/encoding.go
yacovm: It's not unique. A node may send several approvals, for different payloads. 

That's why we filter them:

```
	newApprovals = newApprovals.Filter(func(i int, approval ValidatorSetApproval) bool {
		// Pick only approvals that agree with our candidate auxiliary info digest and P-Chain height
		return approval.PChainHeight == pChainHeight && approval.AuxInfoSeqDigest == candidateAuxInfoDigest
	})
```

msm/msm.go
samliok: what about during replication? is it possible we are verifying an older block while also building a new one? 

Even though its not used concurrently, I still feel like the lazy init pattern adds unnecessary complexity. Would it be better to have a constructor that does this initialization(would also resolve the comment above)?

yacovm: > I still feel like the lazy init pattern adds unnecessary complexity. Would it be better to have a constructor that does this initialization(would also resolve the comment above)?

Yeah I can add a constructor, sure.

> what about during replication? is it possible we are verifying an older block while also building a new one?

So, I think that it makes no sense to verify the MSM state transition for a notarized or finalized block. We should therefore only actually verify the state transition if it's a block we've received from the leader in the normal consensus path.

msm/README.md
```diff
@@ -0,0 +1,454 @@
...
+
+- The `next_epoch_approvals` is a canoto message that contains the approvals of the next epoch by at least `n-f` nodes.
+  It is required in order to create the sealing block of the epoch and indicates that the nodes comprising the next epoch's validator set
+  are ready to start the next epoch. It prevents cases where nodes in the current epoch creating a sealing block and moving to the next epoch
+  while the nodes of the next epoch still replicating blocks and cannot yet participate in the consensus protocol.
+
+- The genesis block does not contain the Simplex epoch information, and it is implicitly set to be the zero values of the fields above.
+
+
+### Auxiliary information encoding
```
samliok, line 107: alongside separating out the ICM stuff from this PR, I think we should also remove the auxillary information   to reduce the diff of this PR. thoughts?

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func (fn *fakeNode) tryFinalizeNextBlock() {
+	nextIndex := len(fn.finalizedBlocks)
+
+	if fn.isNextBlockTelock() {
+		return
+	}
+
+	block := fn.notarizedBlocks[nextIndex]
```
samliok, line 280: it seems like we can some index out of bounds exceptions/nil panics if `notarizedBlocks` & `finalizedBlocks` are out of sync. For example this test panics

```go
func TestTryFinalize(t *testing.T) {
	fn := newFakeNode(t)

	fn.tryFinalizeNextBlock()
}
```

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+	require.Greater(t, node.Epoch(), epoch)
+	require.Equal(t, node.Height(), uint64(30))
+}
+
+type innerBlock struct {
+	metadata.InnerBlock
+	Prev [32]byte
+}
+
+type fakeNode struct {
```
samliok, line 172: im not a fan of the name `fakeNode`, it is too generic and non-descriptive as it doesn't really tell me much about what the node actually does, i.e. the part being `faked`. 

we already have `BasicNode`, `LongRunningNode`, `ControlledNode`. Maybe `MultiEpochNode`?

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/ava-labs/simplex"
+	metadata "github.com/ava-labs/simplex/msm"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFakeNode(t *testing.T) {
```
samliok, line 19: this test was initially hard to follow. `TestFakeNode` makes it seem like we are trying to test the fake node implementation, even though we don't really care about testing the fake node implementation. Maybe `TestStateMachineEpochTransition` is better?

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+	metadata.InnerBlock
+	Prev [32]byte
+}
+
+type fakeNode struct {
+	t               *testing.T
+	sm              metadata.StateMachine
+	mempoolEmpty    bool
+	notarizedBlocks []metadata.StateMachineBlock
+	finalizedBlocks []metadata.StateMachineBlock
```
samliok, line 177: should we consider consolidating these into a single data structure?

msm/fake_node_test.go
samliok: is there any benefit to having the block in two places? can we have something like 

```go
 type blockState struct {
      block     metadata.StateMachineBlock
      finalized bool
  }
```

and then remove the two arrays and just have a singular array in the node state?

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+
+// RetrievingOpts specifies the options for retrieving a block by height and/or digest.
+type RetrievingOpts struct {
+	// Height is the sequence number of the block to retrieve.
+	Height uint64
+	// Digest is the expected hash of the block, used for validation.
+	Digest [32]byte
+	// ShouldBeFinalized indicates whether the block being retrieved is expected to be finalized.
+	// This optimizes retrieval by allowing the retriever to go directly to persisted storage.
+	ShouldBeFinalized bool
```
samliok: is this optimization needed? Finalized blocks can also be in memory so if we are going directly to persisted storage in this case we are being unoptimized 🤔. I feel like it would be cleaner to remove this struct altogether and pass in Seq & Digest, similarly to how we do for `BlockDigestRequests`

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	fn.sm.BlockBuilder = fn
+	fn.sm.PChainProgressListener = fn
+
+	fn.sm.GetBlock = func(opts metadata.RetrievingOpts) (metadata.StateMachineBlock, *simplex.Finalization, error) {
+		if opts.Height == 0 {
+			return genesisBlock, nil, nil
+		}
+		for _, block := range fn.finalizedBlocks {
```
samliok, line 216: since finalized blocks is just a copy of notarized blocks, shouldn't we only check `fn.notarizedBlocks`. If we combined the two fields though, we wouldn't need to worry about this.

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+
+func (fn *fakeNode) Height() uint64 {
+	return uint64(len(fn.finalizedBlocks))
+}
+
+func (fn *fakeNode) Epoch() uint64 {
+	return fn.notarizedBlocks[len(fn.notarizedBlocks)-1].Metadata.SimplexEpochInfo.EpochNumber
+}
+
+func (fn *fakeNode) act() {
```
samliok, line 256:
```suggestion
// act randomly either finalizes a notarized block, builds and notarizes a new block, or does nothing.
func (fn *fakeNode) act() {
```

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+	Prev [32]byte
+}
+
+type fakeNode struct {
+	t               *testing.T
+	sm              metadata.StateMachine
+	mempoolEmpty    bool
+	notarizedBlocks []metadata.StateMachineBlock
+	finalizedBlocks []metadata.StateMachineBlock
+	innerChain      []innerBlock
```
samliok, line 178: this seems unused?

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+	fn.t.Logf("Building a block on top of %s parent with epoch %d", finalizedString, parentBlock.Metadata.SimplexEpochInfo.EpochNumber)
+
+	block, err := fn.sm.BuildBlock(context.Background(), parentBlock, simplex.ProtocolMetadata{
+		Seq:   lastMD.Seq + 1,
+		Round: lastMD.Round + 1,
+		Prev:  prevBlockDigest,
+	}, nil)
+	require.NoError(fn.t, err)
+
+	return block.InnerBlock, block
```
samliok, line 339: i think we can just return `block` since block.InnerBlock is unused?

msm/fake_node_test.go
```diff
@@ -0,0 +1,429 @@
...
+	lastMD, prevBlockDigest := fn.prepareMetadataAndPrevBlockDigest()
+
+	_, finalizatin, err := fn.sm.GetBlock(metadata.RetrievingOpts{
+		Digest: prevBlockDigest,
+		Height: lastMD.Seq,
+	})
+	require.NoError(fn.t, err)
+
+	finalizedString := "not finalized"
+	if finalizatin != nil {
```
samliok:
```suggestion
	if finalization != nil {
```

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+
+	defer func() {
+		elapsed := time.Since(start)
+		sm.Logger.Debug("Built block",
+			zap.Uint64("seq", simplexMetadata.Seq),
+			zap.Uint64("epoch", simplexMetadata.Epoch),
+			zap.Stringer("prevHash", simplexMetadata.Prev),
+			zap.Duration("elapsed", elapsed),
+		)
+	}()
```
samliok, line 219: just curious here, but I know in `avalanchego` we have `vms/tracedvm` that keeps track of the time of function calls. Should we use the same pattern here? we could have a separate wrapper `TracedStateMachine` that handles these logs & potentially uses the same stack as avalanchego?

could be extra, but if you think its worth doing, should we note it as an issue and do it in a separate pr?

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+
+const (
+	BlockTypeNormal BlockType = iota
+	BlockTypeTelock
+	BlockTypeSealing
+	BlockTypeNewEpoch
+)
+
+// BuildBlock constructs the next block on top of the given parent block, and passes in the provided simplex metadata and blacklist.
+func (sm *StateMachine) BuildBlock(ctx context.Context, parentBlock StateMachineBlock, simplexMetadata simplex.ProtocolMetadata, simplexBlacklist *simplex.Blacklist) (*StateMachineBlock, error) {
```
samliok, line 196: i think we can rename instances of `simplexBlacklist` to `blacklist` since we know its part of`simplex`

samliok, line 196: `simplexBlacklist` doesn't have to be a pointer here + we don't have any tests that try building a block with a non empty blacklist. Should we include some?

samliok, line 196: is it redundant to pass in `parentBlock`? we can just get the parent block by using the `simplexMetadata`?

samliok: i also dont think we ever ensure `parentBlock` is consistent with simplexMetadata.prev

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+	sm.maybeInit()
+
+	// The zero sequence number is reserved for the genesis block, which should never be built.
+	if simplexMetadata.Seq == 0 {
```
samliok, line 200: should we add a test for this case?

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+	}
+
+	if block.InnerBlock == nil {
+		return nil
+	}
+
+	return block.InnerBlock.Verify(ctx)
+}
+
+func (sm *StateMachine) identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) (state, error) {
```
```suggestion
  func epochState(epochInfo SimplexEpochInfo) (state, error) {
```
i dont think this function should be associated with `StateMachine`.

samliok: I think this suggestion is even better than the previous. Also we don't need to return `error` regardless
```suggestion
func (i *SimplexEpochInfo) CurrentState() state {
```

we should also probably write tests for this function

msm/msm.go
```diff
@@ -0,0 +1,1007 @@
...
+		return stateBuildBlockEpochSealed, nil
+	}
+
+	// In any other case, NextPChainReferenceHeight > 0 but the previous block is not a Telock or sealing block,
+	// it means we are in the process of collecting approvals for the next epoch.
+	return stateBuildCollectingApprovals, nil
+}
+
+func computePrevVMBlockSeq(parentBlock StateMachineBlock, prevBlockSeq uint64) uint64 {
+	// Either our parent block has no inner block, in which case we just inherit its previous VM block sequence,
```
```suggestion
	// Either our parent block has no inner block, in which case we just inherit its previous VM block sequence.
```

msm/msm.go
samliok: if we pass in the simplexMetadata as `simplex.Metadata` instead of bytes, we can remove `prevBlockSeq` from all the function headers and just use `simplexMetadata.Seq - 1`. Then when we need the bytes, we should just call `simplexMetadata.Bytes()`

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/501e0f1e24be313c6d570623820a844062ed54a1

msm/msm_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/10188dc510ab11cefd743d0f4e47adb8dffbb064

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/501e0f1e24be313c6d570623820a844062ed54a1

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/c77a88b1f1729585d30b0a07f4148a22b208d5df

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/763761539f47d6a5d276dc9dba4350e1391a3d59

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/59bcbb0022b6ef6f4923272dc1c2f9a1eb516899

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/d59e5e3d8769fb1c1544ede3828a92852b4a64d2

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/commits/8be5e9b03f6e06aa053ba1855b322e02835e6713

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/20a72cd32d0deea372799420278038003c53b605

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/79d6a85ed63a635a798e65d33ee7e73a148b3de4

msm/msm_test.go
yacovm: I removed ICM epochs

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/c77a88b1f1729585d30b0a07f4148a22b208d5df

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/c77a88b1f1729585d30b0a07f4148a22b208d5df

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/c77a88b1f1729585d30b0a07f4148a22b208d5df

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/95df444dc242f67c6dd6e1bf7ade98b951ef6d0d

> also we could just store the NodeBLSMappings as a map and then assert equality in linear time rather than sorting both functions and then iterating(O(nlogn) * 2 + n) ?

As a general rule I prefer not to store maps because iteration on them is non deterministic and we are signing over payloads in the MSM so we'll have conflicts

msm/README.md
yacovm: OK I removed the ICM and aux info stuff out of this PR.

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/2ac42b51ff2c513959ec664d9618495852bbaf65

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/commits/44e73fbd4df4cc9b0216a0d23b82c68ec7e64bc1

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/commits/93abf7a9e85537f26773d158765c92a9acc298d2

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/2ac42b51ff2c513959ec664d9618495852bbaf65

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/commits/2ac42b51ff2c513959ec664d9618495852bbaf65

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/5c2d1a02aad1b341dac22eaf51a78259f53f2af7

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/f2dda7dfa4f0625d353b2c44e6d14e197d32946d

msm/fake_node_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/72cd9c844467fdb0edab211b56a8918d422dd8b9

msm/fake_node_test.go
yacovm: Fixed

msm/msm.go
yacovm: It makes sense to do that if we have a ton of methods. We also have something for the engine, I think.

But we have two methods here - Build and Verify. I don't think it justifies a logging wrapper.

Also, traces are used not for logging but for measuring events across functions and nodes. 
This isn't what we're trying to do here at this point.

yacovm: > could be extra, but if you think its worth doing, should we note it as an issue and do it in a separate pr?

I think we should do it if we decide it's necessary.

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/2b869d5d73d8f7ee08cf9cbd73252ee05a8b61d7

yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/2b869d5d73d8f7ee08cf9cbd73252ee05a8b61d7

Made in a non-pointer (well, Claude did...) 

> we don't have any tests that try building a block with a non empty blacklist. Should we include some?

All we do with the blacklist is call `Bytes()` on it. I don't think it's needed because there is no significance to its content in the code.

yacovm: > is it redundant to pass in parentBlock? we can just get the parent block by using the simplexMetadata?

Yeah, because i need access to the inner block (the VM block) to compute the prev VM sequence. 

> i also dont think we ever ensure parentBlock is consistent with simplexMetadata.prev

Simplex computes this and then calls BuildBlock on the MSM. Why would we check what Simplex computed in the previous function call?

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/commits/ef828b2add8c672ded364ff5d7d208d9f3df6f2a

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/ff79a8386a9c078d176e544b3f2d844973de6bd5

yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/ff79a8386a9c078d176e544b3f2d844973de6bd5

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/7d7d0c1c2833e9e898124a52ca39fb3f890f48d1

msm/msm.go
yacovm: Removed it https://github.com/ava-labs/Simplex/pull/366/commits/32bf41e2196998936159916d0b1962c151aca8d5

PR #363, opened 2026-04-16
Simplex reconfiguration framework - Part I (Helpers)

yacovm:
This commit adds some helpers for the Simplex reconfiguration framework.
    
This commit entails:
    
 - README.md changes to align with latest implementation
 - misc.go contains helpers which will be removed once we moved the code to avalanchego.
 - builds_decision.go contains a wrapper to the mempool that also listens to P-chain changes,
      and returns either when the mempool needs to build a block or re-configuration is in order.
 - encoding.go contains the types that encode Simplex metadata and blocks.


msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+// blockBuildingDecision represents the decision of whether we should build a block at the current time,
+// and if so, whether we should also transition to a new epoch along the way.
+type blockBuildingDecision int8
+
+const (
+	blockBuildingDecisionUndefined                    blockBuildingDecision = iota
+	blockBuildingDecisionBuildBlock                                         // We should build a block, and we don't need to transition to a new epoch.
+	blockBuildingDecisionTransitionEpoch                                    // We should transition to a new epoch immediately, but we don't need to build a block.
+	blockBuildingDecisionBuildBlockAndTransitionEpoch                       // We should build a block and transition to a new epoch along the way.
+	blockBuildingDecisionContextCanceled
```
samliok: i think this is probably overly verbose. is this better?

```suggestion
	  const (
      decisionUndefined decision = iota                                                                                               
      decisionBuild                  // Build a block; stay in current epoch.                                                       
      decisionTransition             // Transition epoch; don't build.       
      decisionBuildAndTransition     // Build a block and transition epoch.                                                           
      decisionCanceled                                                                                                                
  )  
```

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+// See the file LICENSE for licensing terms.
+
+package metadata
+
+import (
+	"bytes"
+	"slices"
+)
+
+// go:generate go run github.com/StephenButtolph/canoto/canoto encoding.go
```
samliok:
```suggestion
//go:generate go run github.com/StephenButtolph/canoto/canoto encoding.go
```

msm/misc.go
```diff
@@ -0,0 +1,122 @@
...
+
+func bitmaskFromBytes(bytes []byte) bitmask {
+	var bm bitmask
+	(*big.Int)(&bm).SetBytes(bytes)
+	return bm
+}
+
+// Logger defines the interface that is used to keep a record of all events that
+// happen to the program
+type Logger interface {
```
samliok: dont we already have this interface defined in `api.go`?

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+	canotoData canotoData_OuterBlock
+}
+
+// StateMachineMetadata defines the metadata that the StateMachine uses to transition between epochs,
+// and maintain ICM epoch information.
+// TODO: change SimplexProtocolMetadata and SimplexBlacklist to be non-opaque types.
+// TODO: This requires to encode the protocol metadata and blacklist using canoto.
+type StateMachineMetadata struct {
+	// SimplexEpochInfo is the metadata that the StateMachine uses for its own epoching.
+	SimplexEpochInfo SimplexEpochInfo `canoto:"value,1"`
```
samliok, line 19: i feel like `EpochInfo` is probably fine. `Simplex` is sort of assumed since its in the simplex repo, and we don't have any other epoch structs?

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+	// It is set only in the sealing block and the blocks preceding it starting from a block that has a NextPChainReferenceHeight set.
+	NextEpochApprovals *NextEpochApprovals `canoto:"pointer,7"`
+	// SealingBlockSeq is the block sequence of the sealing block of the current epoch.
+	// It defines the validator set of the next epoch.
+	SealingBlockSeq uint64 `canoto:"uint,8"`
+
+	canotoData canotoData_SimplexEpochInfo
+}
+
+func (sei *SimplexEpochInfo) IsZero() bool {
```
samliok, line 68: `IsEmpty` naming matches the blacklists pattern.

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+		return false
+	}
+	if !slices.Equal(nbm.BLSKey, other.BLSKey) {
+		return false
+	}
+	if nbm.Weight != other.Weight {
+		return false
+	}
+	return true
+}
```
samliok, line 138: does something like this make sense for `Equal`? This would allow us to delete the clone functions
```go
func (nbms NodeBLSMappings) Equal(other NodeBLSMappings) bool {                                                                     
      if len(nbms) != len(other) {                                                                                                    
          return false                                                                                                                
      }                     
      index := make(map[nodeID]*NodeBLSMapping, len(nbms))
      for i := range nbms {                               
          index[nbms[i].NodeID] = &nbms[i]                                                                                            
      }                                   
      for i := range other {                                                                                                          
          m, ok := index[other[i].NodeID]
          if !ok || !m.Equals(&other[i]) {                                                                                            
              return false                
          }                                                                                                                           
      }    
      return true                                                                                                                     
  }          
```

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+
+type BlockValidationDescriptor struct {
+	AggregatedMembership AggregatedMembership `canoto:"value,1"`
+
+	canotoData canotoData_BlockValidationDescriptor
+}
+
+func (bvd *BlockValidationDescriptor) Equals(other *BlockValidationDescriptor) bool {
+	if bvd == nil && other == nil {
```
samliok, line 147: bdv can never be nil since we would panic if calling nil.Equal()

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+	go func() {
+		defer wg.Done()
+		err := bbd.pChainlistener.WaitForProgress(pChainAwareContext, pChainHeight)
+		if err != nil && pChainAwareContext.Err() == nil {
+			bbd.logger.Warn("error while waiting for P-chain progress", zap.Error(err))
+		}
+		cancel()
+	}()
+
+	bbd.waitForPendingBlock(pChainAwareContext)
```
samliok, line 126: we have an invariant where `bbd.waitForPendingBlock(pChainAwareContext)` must return when the context is cancelled. if this is common in golang can just ignore this comment, otherwise should we note it  somewhere?

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	defer wg.Wait()
+	defer cancel()
+
+	go func() {
+		defer wg.Done()
+		err := bbd.pChainlistener.WaitForProgress(pChainAwareContext, pChainHeight)
+		if err != nil && pChainAwareContext.Err() == nil {
```
samliok, line 120: Should we propagate this error to `shouldBuildBlock`? Otherwise, there seems to be a bug: if pChainlistener.WaitForProgress returns an error, we exit `waitForPChainChangeOrPendingBlock`. Since only the inner context was canceled, we won’t return `blockBuildingDecisionContextCanceled`. And because the P-chain height hasn’t changed, we end up returning:

`return blockBuildingDecisionBuildBlock, pChainHeight, nil`

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+
+		// Else, we have reached here because a block is ready to be built, and the P-chain height has not changed,
+		// which means we should build a block.
+
+		return blockBuildingDecisionBuildBlock, pChainHeight, nil
+	}
+}
+
+// waitForPChainChangeOrPendingBlock waits until either the given P-chain height changes from the provided pChainHeight,
+// or a block is ready to be built.
```
samliok:
```suggestion
// waitForPChainChangeOrPendingBlock waits until either the given P-chain height changes from the provided pChainHeight,
// or a block is ready to be built, or if `ctx` gets cancelled.
```

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+// PChainProgressListener listens for changes in the P-chain height.
+type PChainProgressListener interface {
+	// WaitForProgress should block until either the context is cancelled, or the P-chain height has increased from the provided pChainHeight.
+	WaitForProgress(ctx context.Context, pChainHeight uint64) error
+}
+
+type blockBuildingDecider struct {
+	logger                   Logger
+	maxBlockBuildingWaitTime time.Duration
+	pChainlistener           PChainProgressListener
```
samliok:
```suggestion
	pChainListener           PChainProgressListener
```

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		// We have returned from waitForPendingBlock because the context has timed out, which means we don't need to build a block.
+		return blockBuildingDecisionTransitionEpoch
+	}
+
+	// Block is ready to be built
+	return blockBuildingDecisionBuildBlockAndTransitionEpoch
+}
+
+func (bbd *blockBuildingDecider) wasContextCanceled(ctx context.Context) bool {
```
samliok: i think we can remove this and just check ctx.Err()?

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+	bbd.waitForPendingBlock(impatientContext)
+
+	if impatientContext.Err() != nil {
+		// Check if we have returned because the parent context was cancelled
+		if bbd.wasContextCanceled(ctx) {
+			return blockBuildingDecisionContextCanceled
+		}
+		// We have returned from waitForPendingBlock because the context has timed out, which means we don't need to build a block.
+		return blockBuildingDecisionTransitionEpoch
+	}
```
samliok, line 149:
```suggestion
	if ctx.Err() != nil {
		return blockBuildingDecisionContextCanceled
	}
	if impatientContext.Err() != nil {
		// Timed out without a block — transition epoch only.
		return blockBuildingDecisionTransitionEpoch
	}

```

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+			// If we should transition to a new epoch, maybe we can also build a block along the way.
+			return bbd.maybeBuildBlockWithEpochTransition(ctx), pChainHeight, nil
+		}
+
+		// Else, we don't need to transition to a new epoch, but maybe we should build a block.
+		// We wait for either the P-chain height to change, or for a block to be ready to be built.
+		bbd.waitForPChainChangeOrPendingBlock(ctx, pChainHeight)
+
+		// If the context was cancelled in the meantime, abandon evaluation.
+		if bbd.wasContextCanceled(ctx) {
```
samliok: similarly does this work?
```suggestion
		if ctx.Err() != nil {
```

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+	shouldTransitionEpoch    func(pChainHeight uint64) (bool, error)
+	getPChainHeight          func() uint64
+}
+
+// shouldBuildBlock determines whether we should build a block at the current time,
+// based on the current P-chain height and whether we should transition to a new epoch.
+// It returns a blockBuildingDecision, the current P-chain height sampled at the time of deciding,
+// and an error if the decision cannot be made.
+// The P-chain height is returned because sampling the P-chain height afterwards might be inconsistent with the decision that was made.
+func (bbd *blockBuildingDecider) shouldBuildBlock(
```
samliok, line 66: this is essentially the same functionality as `WaitForPendingBlock`?  i think we should name it similarly to keep consistency

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+		// If the context was cancelled in the meantime, abandon evaluation.
+		if bbd.wasContextCanceled(ctx) {
+			return blockBuildingDecisionContextCanceled, 0, nil
+		}
+
+		// If we've reached here, either the P-chain height has changed, or a block is ready to be built.
+
+		// If the P-chain height changed, re-evaluate again whether we should transition to a new epoch,
+		// or continue waiting to build a block.
+		if bbd.getPChainHeight() != pChainHeight {
```
samliok, line 95: what if we need to build a block _and_ the pchain height has changed?  in this case we will continue in the for loop. then in the next iteration, either we should transition the epoch or we shouldn't. 

in the case that we shouldn't, we will call `waitForPChainChangeOrPendingBlock` again.  so we know we need to build a block, but  its possible the pchain height has changed _again_.  Therefore can't this theoretically never return if pchain heights are increasing fast enough. 

I'm not sure if this is realistic since I dont know if the pchain height will ever increase this fast, but its worth thinking about since an attacker can spam the pchain, thus slowing down blockbuilding on the l1.

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+		pChainHeight := bbd.getPChainHeight()
+
+		shouldTransitionEpoch, err := bbd.shouldTransitionEpoch(pChainHeight)
+		if err != nil {
+			return blockBuildingDecisionUndefined, 0, err
+		}
+
+		if shouldTransitionEpoch {
+			// If we should transition to a new epoch, maybe we can also build a block along the way.
+			return bbd.maybeBuildBlockWithEpochTransition(ctx), pChainHeight, nil
```
samliok: the `maybe` in `maybeBuildBlockWithEpochTransition` is only if `ctx` gets cancelled? If thats the case shouldn't we just call this `buildBlockWithEpochTransition`?

samliok: ex. `BuildBlock` or `WaitForPendingBlock` are not called `MaybeBuildBlock` or `MaybeWaitForPendingBlock` even though cancelling a context wont build the block or tell us a block is ready to be built

msm/build_decision.go
```diff
@@ -0,0 +1,158 @@
...
+	maxBlockBuildingWaitTime time.Duration
+	pChainlistener           PChainProgressListener
+	waitForPendingBlock      func(ctx context.Context)
+	shouldTransitionEpoch    func(pChainHeight uint64) (bool, error)
+	getPChainHeight          func() uint64
+}
+
+// shouldBuildBlock determines whether we should build a block at the current time,
+// based on the current P-chain height and whether we should transition to a new epoch.
+// It returns a blockBuildingDecision, the current P-chain height sampled at the time of deciding,
```
samliok:
```suggestion
// It returns a blockBuildingDecision, the P-chain height sampled at the time of deciding,
```

msm/build_decision_test.go
```diff
@@ -0,0 +1,203 @@
...
+				}
+				<-ctx.Done()
+			},
+		},
+		waitForPendingBlock: func(ctx context.Context) {
+			// First iteration: block on context (P-chain listener will cancel it).
+			// Second iteration: return immediately (VM has a block).
+			if iteration.Add(1) == 1 {
+				<-ctx.Done()
+				return
```
samliok: this return could be dropped

msm/encoding.go
yacovm: We also have ICMEpochInfo... which is re-introduced in a later commit after I removed it as per your request.

msm/encoding.go
yacovm:
```
type NilReceiver struct{}

func (nr *NilReceiver) IsNil() bool {
	return nr == nil
}

func main() {
	var nr *NilReceiver
	fmt.Println(nr.IsNil()) // prints true
}
```

msm/build_decision.go
yacovm: That's why we create a cancel-able context, so we cancel it asynchronously and it will stop blocking. That's common in Golang and is documented:

```
	// WaitForPendingBlock returns when either the given context is cancelled,
```

msm/build_decision.go
yacovm: That's a good point however in practice, if `WaitForProgress` returns an error, it is because `pChainAwareContext` was cancelled or its parent context was cancelled:

```
func (ps *ProgressSubscription[T]) WaitForProgress(ctx context.Context, pos T) error {
	ps.lock.Lock()
	defer ps.lock.Unlock()

	for pos >= ps.progress {
		if err := ps.signal.Wait(ctx); err != nil {
			return err
		}
	}
	return nil
}
```

Therefore, the `pChainAwareContext` was cancelled it's because the parent context was cancelled.

msm/misc.go
yacovm: yeah, it's an artifact of copy-paste from avalanchego

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+	"bytes"
+	"slices"
+)
+
+// go:generate go run github.com/StephenButtolph/canoto/canoto encoding.go
+
+// OuterBlock is the top-level encoding of a Simplex block.
+// It contains the inner block (the block built by the VM),
+// as well as metadata created by the StateMachine.
+type OuterBlock struct {
```
samliok: i'm a bit confused between `msm.OuterBlock` and `simplex.Block`. Why we wrap the `VM` block? and not the `simplex.Block`?

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+	return cloned
+}
+
+func (nbms NodeBLSMappings) TotalWeight() (uint64, error) {
+	return nbms.SumWeights(func(int, NodeBLSMapping) bool {
+		return true
+	})
+}
+
+func (nbms NodeBLSMappings) ForEach(selector func(int, NodeBLSMapping)) {
```
samliok: i think we should remove this method

msm/encoding.go
```diff
@@ -0,0 +1,307 @@
...
+	AuxInfoSeqDigest [32]byte `canoto:"fixed bytes,2"`
+	PChainHeight     uint64   `canoto:"uint,3"`
+	Signature        []byte   `canoto:"bytes,4"`
+
+	canotoData canotoData_ValidatorSetApproval
+}
+
+type ValidatorSetApprovals []ValidatorSetApproval
+
+func (vsa ValidatorSetApprovals) ForEach(f func(int, ValidatorSetApproval)) {
```
samliok: same with this

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/c77a88b1f1729585d30b0a07f4148a22b208d5df

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/366/changes/c77a88b1f1729585d30b0a07f4148a22b208d5df

msm/encoding.go
yacovm: It's just the way we encode a block made by Simplex: 

> OuterBlock is the top-level encoding of a Simplex block.

Since `simplex.Block` is an interface we cannot serialize it to bytes. 

We're not using this outer block yet, it's just there because it's part of the spec.

msm/build_decision.go
yacovm: > in the case that we shouldn't, we will call waitForPChainChangeOrPendingBlock again. so we know we need to build a block, but its possible the pchain height has changed again. Therefore can't this theoretically never return if pchain heights are increasing fast enough.

If our P-chain height is not up to date, then there is no point in building a block, because when we build a block it is verified against the latest ICM epoch we know about which is a function of P-chain height.
If we just build a block with a stale ICM epoch it will just get rejected by nodes that are up-to-date.

The rate of P-chain block creation in the steady state isn't expected to be high, so there is no real liveness problem for a node that is up-to-date with regarding its P-chain height.

yacovm: > I'm not sure if this is realistic since I dont know if the pchain height will ever increase this fast, but its worth thinking about since an attacker can spam the pchain, thus slowing down blockbuilding on the l1.

Even if you spam the P-chain, there is always a block that is currently being agreed upon by snowman. You need to go through 15 successive polls which are usually cast (at best) in 4 iterations of communication round trips, so we should be fine.

msm/encoding.go
yacovm: True, but I actually think that was a mistake and we should've used `IsZero` for the blacklist as well. 

I suggest we change it at a later time and just keep SimplexEpochInfo's IsZero as it's more more idiomatic and it's not a container (stack, list, slice) so it makes no sense to give it an "empty" adjective.

msm/encoding.go
yacovm: I think this implementation is more complex to read than mine 🤷

msm/build_decision.go
yacovm: No it's not. We might need to build a block but there is no "pending" transaction. 

Remember that `WaitForPendingBlock` is connected to the VM. 

When we return from `shouldBuildBlock` we might:

- Need to build a block because the VM wants us to
- Need to build a block to transition to a new epoch
- Not need to build a block (Simplex moves to a new round and cancels the context).

Therefore I named it differently to distinguish from the `WaitForPendingBlock`

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468R20-R24

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-be01cb21dff5bfcd6691f445fd82589735f72839add32c29f657ae8bdf0d8863R11

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468R105

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468R53

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468L152-L158

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468R139-R151

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468R85-R86

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468R131

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-fa14ce9307d84bc3663cb1d40827bd69e19658e033ef90021eab75d13109a468R61

msm/build_decision_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/changes/b7adda8741f908b052a4c0613822adbc238c3b09#diff-e2b05e87487a3395efde33020e635e36f46d4f72ab6c802678ae8d2b740d86cbL125

msm/encoding.go
samliok: sounds good, created an issue to remember this
https://github.com/ava-labs/Simplex/issues/375

msm/encoding.go
samliok: in which PR do we use the `OuterBlock`? I don't see it being used in the other reconfig prs, maybe we should hold off on creating this until its needed?

msm/encoding.go
samliok: are we adding these to a seperate pr? i thought we decided to update this pr with its comments

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/commits/83700c2026e3b0741ab1fa43b73f857fa3425e27

msm/encoding.go
samliok: fair, its a bit faster though going from O(nlogn) -> O(n) but if you think the speedup is not worth the readability i'm fine keeping it as is

msm/build_decision.go
```diff
@@ -0,0 +1,151 @@
...
+	// WaitForProgress should block until either the context is cancelled, or the P-chain height has increased from the provided pChainHeight.
+	WaitForProgress(ctx context.Context, pChainHeight uint64) error
+}
+
+type blockBuildingDecider struct {
+	logger                   simplex.Logger
+	maxBlockBuildingWaitTime time.Duration
+	pChainListener           PChainProgressListener
+	waitForPendingBlock      func(ctx context.Context)
+	shouldTransitionEpoch    func(pChainHeight uint64) (bool, error)
```
samliok: can we have a more detailed comment on what this callback expects? i feel like this should be named `hasValidatorSetChanged` to be more clear. and if we are doing `hasValidatorSetChanged` then the parameters of the callback should be `prevPChainHeight, currentPChainHeight`.

samliok (commented):
looks good to me, one comment but not needed. i think there are a couple unresolved comments from my previous review but when those are resolved i'll hit approve

msm/encoding.go
samliok: yea thats fair, but I would say `Simplex` is a weird prefix here since we have pchain references which are unrelated to simplex. Maybe `StateMachineEpochInfo`?

samliok: Also fine to keep this as is. Going to resolve the comment and not block on it

samliok (approved):

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/commits/9beddac0f9f71d8a1c453fed2cd426fc4febecb1

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/commits/9beddac0f9f71d8a1c453fed2cd426fc4febecb1

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/363/commits/c5098aab15662583015824e81334f266cc037f7b

PR #364, opened 2026-04-16
Simplex reconfiguration framework - Part II (more helpers and definitions)

yacovm:
- Add some structs and interface definitions
- Add some utility methods to be used in the next commit.


msm/msm_test.go
```diff
@@ -0,0 +1,498 @@
...
+func (f *fakeVMBlock) Verify(_ context.Context) error { return nil }
+
+type outerBlock struct {
+	finalization *simplex.Finalization
+	block        StateMachineBlock
+}
+
+type blockStore map[uint64]*outerBlock
+
+func (bs blockStore) clone() blockStore {
```
samliok: `clone` is unused

msm/msm_test.go
```diff
@@ -0,0 +1,498 @@
...
+	newStore := make(blockStore)
+	for k, v := range bs {
+		newStore[k] = v
+	}
+	return newStore
+}
+
+func (bs blockStore) getBlock(opts RetrievingOpts) (StateMachineBlock, *simplex.Finalization, error) {
```
samliok: same with `getBlock`

msm/msm_test.go
```diff
@@ -0,0 +1,498 @@
...
+func (f *fakeVMBlock) Height() uint64                 { return f.height }
+func (f *fakeVMBlock) Timestamp() time.Time           { return time.Time{} }
+func (f *fakeVMBlock) Verify(_ context.Context) error { return nil }
+
+type outerBlock struct {
+	finalization *simplex.Finalization
+	block        StateMachineBlock
+}
+
+type blockStore map[uint64]*outerBlock
```
samliok: `blockStore` is unused

msm/msm_test.go
```diff
@@ -0,0 +1,498 @@
...
+type fakeVMBlock struct {
+	height uint64
+}
+
+func (f *fakeVMBlock) Digest() [32]byte               { return [32]byte{} }
+func (f *fakeVMBlock) Height() uint64                 { return f.height }
+func (f *fakeVMBlock) Timestamp() time.Time           { return time.Time{} }
+func (f *fakeVMBlock) Verify(_ context.Context) error { return nil }
+
+type outerBlock struct {
```
samliok: `outerblock` is also unused

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	if prevBlockSimplexEpochInfo.SealingBlockSeq > 0 || prevBlockSimplexEpochInfo.BlockValidationDescriptor != nil {
+		return stateBuildBlockEpochSealed, nil
+	}
+
+	// In any other case, NextPChainReferenceHeight > 0 but the previous block is not a Telock or sealing block,
+	// it means we are in the process of collecting approvals for the next epoch.
+	return stateBuildCollectingApprovals, nil
+}
+
+func IdentifyBlockType(nextBlockMD StateMachineMetadata, prevBlockMD StateMachineMetadata, prevSeq uint64) BlockType {
```
samliok: we aren't using `IdentifyBlockType` in this PR, maybe we should introduce it later?

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	// Otherwise, we do not fall into any of these cases, so it's a block in the middle of the epoch,
+	// not in the edges.
+	return BlockTypeNormal
+}
+
+// computeNewApproverSignaturesAndSigners computes the signatures of the nodes that approve the next epoch including the previous aggregated signature,
+// and bitmask of nodes that correspond to those signatures, and aggregates all signatures together.
+func computeNewApproverSignaturesAndSigners(nextEpochApprovals *NextEpochApprovals, approvalsFromPeers ValidatorSetApprovals, oldApprovingNodes bitmask, nodeID2ValidatorIndex map[nodeID]int, aggregator SignatureAggregator) ([]byte, bitmask, error) {
```
samliok, line 94: we should check if `nextEpochApprovals` is `nil` before reading from it

samliok, line 94: sometimes we pass `bitmask` by value and othertimes by reference. Can we pass them by reference for consistency?

msm/msm.go
samliok: same could be said for the `state` type alias

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	case BlockTypeSealing:
+		return "Sealing"
+	case BlockTypeNewEpoch:
+		return "NewEpoch"
+	default:
+		return fmt.Sprintf("UnknownBlockType(%d)", state)
+	}
+}
+
+func identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) (state, error) {
```
samliok: should we drop the `error` return?

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	mdDigest := sha256.Sum256(smb.Metadata.MarshalCanoto())
+	combined := make([]byte, 64)
+	copy(combined[:32], blockDigest[:])
+	copy(combined[32:], mdDigest[:])
+	return sha256.Sum256(combined)
+}
+
+// ApprovalsRetriever retrieves the approvals from validators of the next epoch for the epoch change.
+type ApprovalsRetriever interface {
```
samliok: this is unused for this pr

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	RetrieveApprovals() ValidatorSetApprovals
+}
+
+// SignatureVerifier verifies a cryptographic signature against a message and public key.
+// Used to verify Approvals from validators for epoch transitions.
+type SignatureVerifier interface {
```
samliok: unused as well

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// SignatureAggregator combines multiple cryptographic signatures into a single aggregated signature.
+// Used to aggregate validator signatures for epoch transitions.
+type SignatureAggregator interface {
+	AggregateSignatures(signatures ...[]byte) ([]byte, error)
+}
+
+// KeyAggregator combines multiple public keys into a single aggregated public key.
+type KeyAggregator interface {
```
samliok: unused as well

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	AggregateKeys(keys ...[]byte) ([]byte, error)
+}
+
+// ValidatorSetRetriever retrieves the validator set at a given P-chain height.
+type ValidatorSetRetriever func(pChainHeight uint64) (NodeBLSMappings, error)
```
samliok, line 45: unused as well

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+type BlockType uint8
+
+const (
+	BlockTypeNormal BlockType = iota
+	BlockTypeTelock
+	BlockTypeSealing
+	BlockTypeNewEpoch
+)
+
+func (state BlockType) String() string {
```
```suggestion
func (bt BlockType) String() string {
```

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+
+	if totalWeight == 0 {
+		return 0, fmt.Errorf("total weight of validators is 0")
+	}
+
+	if totalWeight > math.MaxInt64 {
+		return 0, fmt.Errorf("total weight of validators is too big, overflows int64: %d", totalWeight)
+	}
+	return int64(totalWeight), nil
+}
```
```suggestion
}

```

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	// If we don't have a next P-chain preference height, it means we are not transitioning to a new epoch just yet.
+	if prevBlockSimplexEpochInfo.NextPChainReferenceHeight == 0 {
+		return stateBuildBlockNormalOp, nil
+	}
+
+	// If the previous block has a sealing block sequence, it's a Telock.
+	// If it has a block validation descriptor, it's a sealing block.
+	// Eithe way, the epoch has been sealed.
```
```suggestion
	// Either way, the epoch has been sealed.
```

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+
+type approvals struct {
+	canSeal   bool
+	nodeIDs   []byte
+	signature []byte
+}
+
+func ensureNextEpochApprovalsSignersSupersetOfApprovalsOfPrevBlock(prev SimplexEpochInfo, next SimplexEpochInfo) error {
+	if prev.NextEpochApprovals == nil {
+		// Condition satisifed vacously.
```
```suggestion
		// Condition satisfied vacuously.
```

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+func computePrevVMBlockSeq(parentBlock StateMachineBlock, prevBlockSeq uint64) uint64 {
+	// Either our parent block has no inner block, in which case we just inherit its previous VM block sequence,
+	if parentBlock.InnerBlock == nil {
+		return parentBlock.Metadata.SimplexEpochInfo.PrevVMBlockSeq
+	}
+	// or it has an inner block, in which case it is the previous block sequence.
+	return prevBlockSeq
+}
+
+type approvals struct {
```
samliok: approvals is never used

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	canSeal   bool
+	nodeIDs   []byte
+	signature []byte
+}
+
+func ensureNextEpochApprovalsSignersSupersetOfApprovalsOfPrevBlock(prev SimplexEpochInfo, next SimplexEpochInfo) error {
```
samliok, line 210: this might be a record for our longest method name 😅

msm/msm_test.go
```diff
@@ -0,0 +1,498 @@
...
+			name: "prev has approvals but next has nil approvals",
+			prev: SimplexEpochInfo{NextEpochApprovals: &NextEpochApprovals{NodeIDs: []byte{1}}},
+			next: SimplexEpochInfo{},
+			err:  "previous block has next epoch approvals but proposed block doesn't have next epoch approvals",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			err := ensureNextEpochApprovalsSignersSupersetOfApprovalsOfPrevBlock(tc.prev, tc.next)
+			if tc.err != "" {
+				require.ErrorContains(t, err, tc.err)
```
samliok: instead of using error contains, can we create actual error variables for these and use `ErrorIs`?

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Else, if the previous block has a sealing block sequence and is in the same epoch as this block,
+	// then this block has to be a Telock, as the sealing block sequence indicates that the sealing block has been created.
+	// [ Sealing Block ] <-- [ Prev block ] <-- [ New Block ]
+	if simplexEpochInfo.EpochNumber == prevSimplexEpochInfo.EpochNumber && prevSimplexEpochInfo.SealingBlockSeq != 0 {
+		return BlockTypeTelock
+	}
+
+	// This block is the first block of its epoch if the epoch number is the sealing block sequence of the previous epoch
+	if simplexEpochInfo.EpochNumber == prevSimplexEpochInfo.SealingBlockSeq {
```
samliok: the zeroth epoch + 0 as the SealingBlockSeq would return `NewEpoch` prematurely?

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+// computeNewApproverSignaturesAndSigners computes the signatures of the nodes that approve the next epoch including the previous aggregated signature,
+// and bitmask of nodes that correspond to those signatures, and aggregates all signatures together.
+func computeNewApproverSignaturesAndSigners(nextEpochApprovals *NextEpochApprovals, approvalsFromPeers ValidatorSetApprovals, oldApprovingNodes bitmask, nodeID2ValidatorIndex map[nodeID]int, aggregator SignatureAggregator) ([]byte, bitmask, error) {
+	// Prepare the new signatures from the new approvals that haven't approved yet and that agree with our candidate auxiliary info digest and P-Chain height.
+	newSignatures := make([][]byte, 0, len(approvalsFromPeers)+1)
+
+	// We will overwrite the old approving nodes with the new approving nodes, by turning on the bits for the new approvers.
+	newApprovingNodes := oldApprovingNodes.Clone()
+
+	for _, approval := range approvalsFromPeers {
```
samliok, line 104: should we ensure the `approvalsFromPeers` are unique before using them here. otherwise we could fail

see this test that fails(Claude wrote) in `TestComputeNewApproverSignaturesAndSigners`

```go

	t.Run("duplicate peer with already-approved node does not double-aggregate", func(t *testing.T) {
		// node0 is already in the previous approvals (bit 0 set). A duplicate peer
		// entry for node0 must not append node0's signature to the new aggregate
		// (the prior aggregate already covers it via prevApprovals.Signature).
		prevApprovals := &NextEpochApprovals{
			NodeIDs:   []byte{1}, // bit 0
			Signature: []byte("existing"),
		}
		oldApproving := bitmaskFromBytes([]byte{1})

		peers := ValidatorSetApprovals{
			{NodeID: node0, Signature: []byte("sig0")},
			{NodeID: node0, Signature: []byte("sig0")},
		}

		aggSig, newApproving, err := computeNewApproverSignaturesAndSigners(prevApprovals, peers, oldApproving, nodeID2Index, concatAggregator{})
		require.NoError(t, err)
		require.True(t, newApproving.Contains(0))
		require.Equal(t, 1, newApproving.Len())
		// Only the existing aggregate should remain; node0's sig is already covered by it.
		require.Equal(t, []byte("existing"), aggSig)
	})
```

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+}
+func findFirstSimplexBlock(getBlock BlockRetriever, endHeight uint64) (uint64, error) {
+	var haltError error
+
+	// Make sure the bound passed to sort.Search is valid, to avoid illegal input caused by overflow.
+	if endHeight >= math.MaxInt {
+		endHeight = math.MaxInt - 1
+	}
+
+	firstSimplexBlock := sort.Search(int(endHeight+1), func(i int) bool {
```
samliok: it seems weird to clamp `endHeight` to an int, when it can very realistically be a unit64

samliok: if we are going to use this, i think we should just fail if the endHeight is > max int

msm/msm.go
yacovm: Doesn't matter, we only use this function in case it's a non-zero block.

For the zero block we have a special handling.

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	if uint64(firstSimplexBlock) > endHeight {
+		return 0, fmt.Errorf("no simplex blocks found in range [%d, %d]", 0, endHeight)
+	}
+
+	return uint64(firstSimplexBlock), nil
+}
+
+func computePrevVMBlockSeq(parentBlock StateMachineBlock, prevBlockSeq uint64) uint64 {
```
samliok, line 196: i'm a bit confused, when we say `VMBlock` do we mean the block built by Avalanchego?

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	if approvingWeight > math.MaxInt64 {
+		return 0, fmt.Errorf("approving weight of validators is too big, overflows int64: %d", approvingWeight)
+	}
+
+	return int64(approvingWeight), nil
+}
+
+func computeTotalWeight(validators NodeBLSMappings) (int64, error) {
```
samliok: not sure where this function will be used, but would it be better to make this a method of `NodeBLSMappings`? i.e `validators.computeTotalWeight`

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+		if !exists {
+			// If the approving node is not in the validator set, we ignore this approval.
+			return false
+		}
+		// Only pick approvals from nodes that haven't already approved
+		return !oldApprovingNodes.Contains(approvingNodeIndexOfNewApprover)
+	}
+}
+
+func computeApprovingWeight(validators NodeBLSMappings, approvingNodes *bitmask) (int64, error) {
```
samliok: same comment as computeTotalWeight, should this be a method of `NodeBLSMappings`?

msm/msm.go
```diff
@@ -0,0 +1,353 @@
...
+func sanitizeApprovals(approvals ValidatorSetApprovals, pChainHeight uint64, nodeID2ValidatorIndex map[nodeID]int, oldApprovingNodes bitmask) ValidatorSetApprovals {
+	filter1 := approvalsThatAgreeWithAuxInfoAndPChainHeight(pChainHeight)
+	filter2 := approvalsThatAreInValidatorSetAndHaveNotAlreadyApproved(oldApprovingNodes.Clone(), nodeID2ValidatorIndex)
+	return approvals.Filter(filter1).Filter(filter2).UniqueByNodeID()
+}
+
+func approvalsThatAgreeWithAuxInfoAndPChainHeight(pChainHeight uint64) func(i int, approval ValidatorSetApproval) bool {
+	return func(i int, approval ValidatorSetApproval) bool {
+		// Pick only approvals that agree with our candidate auxiliary info digest and P-Chain height
+		return approval.PChainHeight == pChainHeight
```
samliok, line 147: where are we checking for the auxillary info?

msm/msm.go
samliok: I assume if we are not using PoS, all of these weights would be equal.  Also, in the simplex `epoch` code we've avoided using `weight` in favor for `quorum` terminology. Should we keep this distinction and abstract away these calculations someplace else? 

We abstract away the quorum stuff in the `SignatureAggregator`. We also use the signature aggregator in the MSM code so I'm concerned about this scenario:

the signature aggregator `IsQuorum` method returns true if 2/3 of the nodes are part of the signature(say these nodes could have different stake). However, in this code we go strictly off the node weight. Therefore isn't it a problem if `IsQuorum` and `canSealBlock` have different quorum criterias?

samliok (approved):
It's a bit hard to review this code because I can't see how many of these functions are called. I know this is a product of splitting up the MSM code into smaller prs so I guess I will more ruber-stampy approve this PR and re-review the entire MSM file on the next PR🥲 

feel free to ignore any irrelevant comments

msm/msm.go
yacovm: Exactly

msm/msm.go
yacovm: yeah I should've used `IsQuorum` instead computing the ratio myself. I'll change it accordingly in the next PRs where it's introduced.

yacovm: https://github.com/ava-labs/Simplex/pull/364/changes/6b1bfbac6418acbaeba3acfc82e04e15f9454543

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/364/changes/0d0017d88645d7bf8ed00f10b003d4b087658578

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/364/commits/098a17754af3c9e68d88e7ba4140593aec2de252

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/364/commits/2c09a67108c80d7b91ee3c538e1b89e4245986eb

msm/msm_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/364/changes/bc968095254feab77211cc8da54ab09c2e9efd90

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/364/changes/445a85156cc7b222a33f6963a296ab6d33148b35

msm/msm.go
samliok: ok so another follow up, is this the `AvalancheGo` block we build in `block.go` or is this the inner `VMBlock` within that avalanchego block?

msm/msm.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// Make sure that previous signers are still there.
+	prevSigners := bitmaskFromBytes(prev.NextEpochApprovals.NodeIDs)
+	nextSigners := bitmaskFromBytes(next.NextEpochApprovals.NodeIDs)
+	// Remove all bits in nextSigners from prevSigners
+	prevSigners.Difference(&nextSigners)
+	// If we have some bits left, it means there was a bit in prevSigners that wasn't in nextSigners
+	if prevSigners.Len() > 0 {
+		return errSignerSetShrinked
```
samliok, line 228: signer set shrunk or shrank 😅 i don't think shrinked is a word but i'm not confident on which of the two we should use. i think Shrunk is correct but not 100%

msm/msm.go
yacovm: well, it's an interface so it can be anything we want it to be. 

The important thing is that when we call `Verify()` it will execute the VM's verify method and not our own,

and the digest will only include the VM block without our own stuff.

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/377

PR #365, opened 2026-04-16
Simplex reconfiguration framework - Part III (MSM implementation)

yacovm:
- Add block building to msm.go
- Add verification.go which contains logic for block verification
- Add tests that mimic Simplex flow (fake_node_test.go)


msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+
+type blockStore map[uint64]*outerBlock
+
+func (bs blockStore) clone() blockStore {
+	newStore := make(blockStore)
+	for k, v := range bs {
+		newStore[k] = v
+	}
+	return newStore
+}
```
samliok:
```suggestion
func (bs blockStore) clone() blockStore {
	newStore := make(blockStore)
	maps.Copy(newStore, bs)
	return newStore
}
```

msm/msm.go
```diff
@@ -39,23 +42,85 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+}
+
+// ValidatorSetRetriever retrieves the validator set at a given P-chain height.
+type ValidatorSetRetriever func(pChainHeight uint64) (NodeBLSMappings, error)
+
+// BlockRetriever retrieves a block and its finalization status given the block's sequence number and expected digest.
 // If the block cannot be found it returns ErrBlockNotFound.
 // If an error occurs during retrieval, it returns a non-nil error.
-type BlockRetriever func(RetrievingOpts) (StateMachineBlock, *simplex.Finalization, error)
+type BlockRetriever func(seq uint64, digest [32]byte) (StateMachineBlock, *simplex.Finalization, error)
```
samliok: do we ever pass in a non-empty digest?

msm/msm.go
```diff
@@ -89,6 +331,455 @@ func identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) state {
...
+
+		firstSimplexBlock, err := findFirstSimplexBlock(sm.GetBlock, sm.LatestPersistedHeight+1)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find first simplex block: %w", err)
+		}
+		firstSimplexBlockRetrieved, _, err := sm.GetBlock(firstSimplexBlock, [32]byte{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to retrieve first simplex block at height %d: %w", firstSimplexBlock, err)
+		}
+		simplexEpochInfo.PrevSealingBlockHash = firstSimplexBlockRetrieved.Digest()
```
samliok: would it be better to just store the genesis/first simplex block in the `StateMachine` struct itself?

msm/msm.go
```diff
@@ -39,23 +42,85 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
-type BlockRetriever func(RetrievingOpts) (StateMachineBlock, *simplex.Finalization, error)
+type BlockRetriever func(seq uint64, digest [32]byte) (StateMachineBlock, *simplex.Finalization, error)
+
+// BlockBuilder builds a new VM block with the given observed P-chain height.
+type BlockBuilder interface {
+	BuildBlock(ctx context.Context, pChainHeight uint64) (VMBlock, error)
+
+	// WaitForPendingBlock returns when either the given context is cancelled,
+	// or when the VM signals that a block should be built.
+	WaitForPendingBlock(ctx context.Context)
```
samliok, line 69: `WaitForPendingBlock` is supposed to be the same as `WaitForPendingBlock` in the `BlockBuilder` interface inside api.go? should we create a common interface like `BlockWaiter` and use it in both `BlockBuilder` interface definitions?

msm/msm.go
samliok: can we remove this method? it seems weird to have a method that does binary search everytime we want to find the first simplex block. i really think this can be simplified if we just store the `firstSimplexBlock` in the `StateMachine` struct.

msm/msm.go
```diff
@@ -66,6 +131,183 @@ const (
 	stateBuildBlockEpochSealed
 )
 
+// BuildBlock constructs the next block on top of the given parent block, and passes in the provided simplex metadata and blacklist.
+func (sm *StateMachine) BuildBlock(ctx context.Context, simplexMetadata simplex.ProtocolMetadata, simplexBlacklist *simplex.Blacklist) (*StateMachineBlock, error) {
+	sm.maybeInit()
```
samliok: we were going to make a constructor and remove this?

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+			require.NoError(t, err)
+			require.NotNil(t, block)
+
+			if testCase.mutateBlock != nil {
+				testCase.mutateBlock(block)
+			}
+
+			err = sm2.VerifyBlock(context.Background(), block)
+			if testCase.err != "" {
+				require.ErrorContains(t, err, testCase.err)
```
samliok, line 135: i think we should create errors, rather than using `ErrorContains`

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+
+func (bs blockStore) getBlock(seq uint64, _ [32]byte) (StateMachineBlock, *simplex.Finalization, error) {
+	blk, exits := bs[seq]
+	if !exits {
+		return StateMachineBlock{}, nil, fmt.Errorf("%w: block %d not found", simplex.ErrBlockNotFound, seq)
+	}
+	return blk.block, blk.finalization, nil
+}
+
+type approvalsRetriever struct {
```
samliok: how do you envision us extending the msm tests? I know you mentioned making a test that has the full validator change flow, would that be using a lot of these components? if so, I think we should separate them like we did with the test node utils so that our helpers are not scattered across testing files

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+
+func (sv *signatureAggregator) IsQuorum(approverWeights []uint64, totalWeights uint64) bool {
+	var sum uint64
+	for _, w := range approverWeights {
+		sum += w
+	}
+	return sum*3 > totalWeights*2
+}
+
+type noOpPChainListener struct{}
```
samliok:
```suggestion
type noopPChainListener struct{}
```

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+
+func (n *noOpPChainListener) WaitForProgress(ctx context.Context, _ uint64) error {
```
samliok: this isn't really a `noop`, it blocks until the context is cancelled. Maybe we should rename it something more clear?

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+type blockBuilder struct {
+	block VMBlock
+	err   error
+}
+
+func (bb *blockBuilder) WaitForPendingBlock(_ context.Context) {
+	// Block is always ready in tests.
+}
+
+func (bb *blockBuilder) BuildBlock(_ context.Context, _ uint64) (VMBlock, error) {
```
samliok: should we fail if the block & err are both nil? I can see us forgetting to set these in later tests and then having to debug weirdness that is caused by our(my) forgetfulness

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+	validatorSetRetriever validatorSetRetriever
+}
+
+func newStateMachine(t *testing.T) (StateMachine, *testConfig) {
+	bs := make(blockStore)
+	bs[0] = &outerBlock{block: genesisBlock}
+
+	var testConfig testConfig
+	testConfig.blockStore = bs
+	testConfig.validatorSetRetriever.result = NodeBLSMappings{
```
samliok: maybe we should have constructors for helpers? seems like it would set them up to be more re-usable + maintainable in the future?

msm/msm.go
```diff
@@ -39,23 +42,85 @@ func (smb *StateMachineBlock) Digest() [32]byte {
 // Used to aggregate validator signatures for epoch transitions.
 type SignatureAggregator interface {
 	AggregateSignatures(signatures ...[]byte) ([]byte, error)
+
+	IsQuorum(approverWeights []uint64, totalWeight uint64) bool
```
samliok: i don't think we should redefine this interface in the msm. Would it be better to use the interface defined in `api.go`? 

I think it's better for 2 reasons, the first being we would currently need to implement these `SignatureAggregator` interfaces using two different structs because function names cannot be overloaded. More importantly, I think we should just have one implementation of `IsQuorum`. I think its reasonable to assume the `SignatureAggregator` would have the weights of each nodes in a PoS setting, so we can just use the `IsQuorum` method defined in `api.go:SignatureAggregator`

msm/msm.go
```diff
@@ -133,6 +824,17 @@ func computeNewApproverSignaturesAndSigners(nextEpochApprovals *NextEpochApprova
 	return aggregatedSignature, newApprovingNodes, nil
 }
 
+func canSealBlock(validators NodeBLSMappings, newApprovingNodes bitmask, sigAggr SignatureAggregator) (bool, error) {
```
samliok: following up on my `IsQuorum` comment, I think `canSealBlock` can just be removed with something like 

```go
nodes := validators.applyBitmask(newApprovingNodes)
canSeal := sigAggr.IsQuorum(nodes)
```

msm/msm.go
yacovm: oh yeah, I like this idea 👍

msm/msm.go
yacovm: The problem is that the signature aggregator cannot know which P-chain block we're referring to, and therefore if it gets an invocation from an epoch or from an MSM it has no clue which P-chain to lookup against.

I agree that it's best that the IsQuorum will be a unified and single API.

What do you think about doing the complete opposite - to have the Epoch use `IsQuorum(approverWeights []uint64, totalWeight uint64) bool` and also add weights to the membership of the communication? This will simplify the implementation of IsQuorum in the avalanchego side.

msm/msm.go
samliok: https://github.com/ava-labs/Simplex/pull/360/changes#r3028812427
https://github.com/ava-labs/Simplex/pull/360/changes#r3024116683

msm/msm.go
```diff
@@ -66,6 +131,183 @@ const (
 	stateBuildBlockEpochSealed
 )
 
+// BuildBlock constructs the next block on top of the given parent block, and passes in the provided simplex metadata and blacklist.
+func (sm *StateMachine) BuildBlock(ctx context.Context, simplexMetadata simplex.ProtocolMetadata, simplexBlacklist *simplex.Blacklist) (*StateMachineBlock, error) {
```
```suggestion
func (sm *StateMachine) BuildBlock(ctx context.Context, metadata simplex.ProtocolMetadata, blacklist *simplex.Blacklist) (*StateMachineBlock, error) {
```

msm/msm.go
yacovm: We can do that, but I am not sure what's the point of having an interface for a single method. 

Usually we define an interface to embed inside another interface to avoid repetition, but this is only one method 🤷 

I guess you can argue that if we change the signature of the method we only need to change a single place, but it seems to me unimportant. I don't mind doing that if you feel strongly about it.

The reason I'm commenting at such length is that in Golang, we have structural equivalence vs. name equivalence (like Java, etc.) , therefore the important parts of an interface are its methods, not its name.

This interface is supposed to be implemented by the VM.

The BlockBuilder interface in `api.go` used by the epoch is supposed to be implemented by the MSM (I suppose we will just override `WaitForPendingBlock(ctx context.Context)` to proxy it over to the epoch.

So, the block builder interface for the MSM is this one, while the block builder interface for the epoch is defined in `api.go` (you can argue it should be placed in closer proximity to the epoch, and I guess we can move it there, but let's defer the code moving and arrangement to a later time)

msm/msm.go
yacovm: yeah that's one of the things I still need to cherry pick from https://github.com/ava-labs/Simplex/pull/366

msm/msm.go
yacovm: yeah that's a good idea, we'll just inject it all the way from avalanchego 👍

msm/msm_test.go
yacovm: In golang, all methods that get a context should return once it's cancelled.

msm/msm_test.go
yacovm: yeah makes sense

msm/msm_test.go
yacovm: the approvals retriever is supposed to be retrieving the approvals from a kind of mempool. I think in a test that has multiple nodes, we'll implement (for now) a fake single-per-network approvals retriever to disseminate approvals. 

Then later once we have the implementation of the real mempool of approvals + its dissemination logic, we'll just use that.

Does that make sense?

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+			sm1, testConfig1 := newStateMachine(t)
+			sm2, testConfig2 := newStateMachine(t)
+
+			testConfig1.blockStore[0] = &outerBlock{
+				block: genesisBlock,
+			}
+
+			testConfig2.blockStore[0] = &outerBlock{
+				block: genesisBlock,
+			}
```
samliok: these are unnecessary as we set the genesis already in `newStateMachine(t)`

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+	genesisBlock = StateMachineBlock{
+		// Genesis block metadata has all zero values
+		InnerBlock: &InnerBlock{
+			TS:    time.Now(),
+			Bytes: []byte{1, 2, 3},
+		},
+	}
+)
+
+func TestMSMFirstBlockAfterGenesis(t *testing.T) {
```
samliok:
```suggestion
func TestMSMBuildAndVerifyBlocksAfterGenesis(t *testing.T) {
```
this slightly more descriptive imo

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+		err         string
+		configure   func(*StateMachine, *testConfig)
+		mutateBlock func(*StateMachineBlock)
+	}{
+		{
+			name: "correct information",
+			md:   validMD,
+		},
+		{
+			name: "trying to build a genesis block",
```
samliok: but we are not trying to build a genesis block, since the `MD` is not the genesis MD?

msm/msm.go
```diff
@@ -66,6 +131,183 @@ const (
...
+
+	pmd, err := simplex.ProtocolMetadataFromBytes(block.Metadata.SimplexProtocolMetadata)
+	if err != nil {
+		return fmt.Errorf("failed to parse ProtocolMetadata: %w", err)
+	}
+
+	seq := pmd.Seq
+
+	if seq == 0 {
+		return fmt.Errorf("attempted to build a genesis inner block")
```
samliok, line 228: this is in verification though, not block building

msm/msm.go
samliok: to recap our quick sync offline: let's initialize `SignatureAggregators` with a set of node <-> weight mappings, and then use the newly initialized signature aggregator when calling `IsQuorum` in the MSM approval logic.  This allows us to not worry about passing in weights and keep the abstraction in AvalancheGo as is.

msm/msm_test.go
yacovm: but we're mutating the metadata: `md.Seq = 0`

msm/msm.go
yacovm: yeah but the idea is "hey, someone tried to build a genesis block". 

Also it shouldn't be "inner" - it's a copy-paste error I thought I fixed.

Will fix it to something more reasonable like "received a genesis block"

msm/msm_test.go
```diff
@@ -25,6 +28,1039 @@ func (f *fakeVMBlock) Height() uint64                 { return f.height }
...
+			TS:    time.Now(),
+			Bytes: []byte{1, 2, 3},
+		},
+	}
+)
+
+func TestMSMFirstBlockAfterGenesis(t *testing.T) {
+	validMD := simplex.ProtocolMetadata{
+		Round: 0,
+		Seq:   1,
```
samliok: we shouldn't be able to have a round < sequence?

msm/msm.go
```suggestion
	stateFirstSimplexBlock state = iota + 1
```
in case we pass in an uninitialized state somewhere we don't want to accidentally build the first block

msm/msm.go
```diff
@@ -66,6 +131,183 @@ const (
...
+	}
+
+	if block.InnerBlock == nil {
+		return nil
+	}
+
+	return block.InnerBlock.Verify(ctx)
+}
+
 func identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) state {
```
samliok: what are your thoughts on making this a function of `SimplexEpochInfo`? i.e. we can just call `prevBlockSimplexEpochInfo.NextBlockState()`. i personally think i'd be cleaner

msm/msm.go
```diff
@@ -66,6 +131,183 @@ const (
...
+	}
+
+	prevMD := prevBlock.Metadata
+	currentState := identifyCurrentState(prevMD.SimplexEpochInfo)
+
+	switch currentState {
+	case stateFirstSimplexBlock:
+		err = sm.verifyBlockZero(ctx, block, prevBlock)
+	default:
+		err = sm.verifyNonZeroBlock(ctx, block, prevBlock.Metadata, currentState, seq-1)
```
samliok: i think the verification stuff would be good to separate out for a different PR?

msm/msm.go
```diff
@@ -89,6 +331,455 @@ func identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) state {
...
+
+	return sm.wrapBlock(parentBlock, childBlock, newSimplexEpochInfo, pChainHeight, simplexMetadata, simplexBlacklist), nil
+}
+
+// buildBlockZero builds the first ever block for Simplex,
+// which is a special block that introduces the first validator set and starts the first epoch.
+func (sm *StateMachine) buildBlockZero(ctx context.Context, parentBlock StateMachineBlock, simplexMetadata, simplexBlacklist []byte) (*StateMachineBlock, error) {
+	pChainHeight := sm.GetPChainHeight()
+
+	newValidatorSet, err := sm.GetValidatorSet(pChainHeight)
```
samliok: aren't the initial validators encoded in the subnet config?

msm/msm.go
samliok, line 820:
```suggestion
	// Either our parent block has no inner block, in which case we just inherit its previous VM block sequence.
```

```suggestion
		// Condition satisfied vacuously.
```

msm/msm.go
```diff
@@ -89,6 +331,455 @@ func identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) state {
...
+		simplexEpochInfo.PrevSealingBlockHash = firstSimplexBlockRetrieved.Digest()
+	}
+
+	return sm.buildBlockImpatiently(ctx, parentBlock, simplexMetadata, simplexBlacklist, simplexEpochInfo, pChainHeight)
+}
+
+// wrapBlock creates a new StateMachineBlock by wrapping the VM block (if applicable) and adding the appropriate metadata.
+func (sm *StateMachine) wrapBlock(parentBlock StateMachineBlock, childBlock VMBlock, newSimplexEpochInfo SimplexEpochInfo, pChainHeight uint64, simplexMetadata, simplexBlacklist []byte) *StateMachineBlock {
+	parentMetadata := parentBlock.Metadata
+	timestamp := parentMetadata.Timestamp
```
```suggestion
	timestamp := parentBlock.Metadata.Timestamp
```

msm/msm.go
```diff
@@ -89,6 +331,455 @@ func identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) state {
...
+
+	hasChildBlock := childBlock != nil
+
+	var newTimestamp time.Time
+	if hasChildBlock {
+		newTimestamp = childBlock.Timestamp()
+		timestamp = uint64(newTimestamp.UnixMilli())
+	}
```
```suggestion
    timestamp := parentBlock.Metadata.Timestamp
	if childBlock != nil {
		timestamp = uint64(childBlock.Timestamp().UnixMilli())
	}
```

msm/msm.go
samliok, line 17: `StateMachineBlock` is supposed to implement the `simplex.VerifiedBlock` interface right? if so, can we add a type assertion?

msm/msm.go
```diff
@@ -89,6 +331,455 @@ func identifyCurrentState(prevBlockSimplexEpochInfo SimplexEpochInfo) state {
...
+		prevVMBlockSeq = parentBlock.InnerBlock.Height()
+	} else {
+		// We can only have blocks without inner blocks in Simplex blocks, but this is the first Simplex block.
+		// Therefore, the parent block must have an inner block.
+		sm.Logger.Error("Parent block has no inner block, cannot determine previous VM block sequence for zero block")
+		return nil, fmt.Errorf("failed constructing zero block: parent block has no inner block")
+	}
+	simplexEpochInfo := constructSimplexZeroBlock(pChainHeight, newValidatorSet, prevVMBlockSeq)
+
+	return sm.buildBlockImpatiently(ctx, parentBlock, simplexMetadata, simplexBlacklist, simplexEpochInfo, pChainHeight)
```
samliok: 🤔

the caller [`BuildBlock`](https://github.com/ava-labs/Simplex/blob/84c769656b60c2101ea4cf390b6a046bc4851194/msm/msm.go#L135) is only called when we are selected to build a block. We also do the wait impatiently thingy here, but more often than not we will actually _not_ have an actual block to build. I say this because when we start a chain we will probably not begin issuing transactions right away. 

Therefore, more often than not the first Simplex block will just be an empty one that is produced right after the simplex chain is created. This also means that most block "zero" will be the same. 

Because most zero blocks will be the same I'm thinking we could either hardcode the zeroth block and remove a lot of the branching that is caused by treating zeroth blocks so specially(no need for the build block zero method, the verify method becomes simple and a lot of branching is removed). 

Another possibility -> why rush to build the zeroth block? can't we just wait like the normal blocks?

Maybe my logic is wrong somewhere, but I generally have a feeling there is a much simpler solution with handling the zeroth case.

msm/msm_test.go
samliok: so this test is more-so verifying rather than building the genesis block, could we rename?

msm/msm.go
samliok: i would say that if the methods are supposed to do the same thing then the same interface should be re-used rather than creating a new one. i don't feel too strongly about it, just thought it would be a slight improvement but wont really effect code correctness

> but let's defer the code moving and arrangement to a later time

im concerned that a later time will end up being never 😅

> So, the block builder interface for the MSM is this one, while the block builder interface for the epoch is defined in api.go

this is a fair point, i'd say we can keep it as for now

msm/msm.go
yacovm: yeah. this is a note to myself to open an issue that discusses this at length tomorrow

msm/msm.go
yacovm: > Because most zero blocks will be the same I'm thinking we could either hardcode the zeroth block and remove a lot of the branching that is caused by treating zeroth blocks so specially(no need for the build block zero method, the verify method becomes simple and a lot of branching is removed).

we can't hardcode the zero block because it has metadata fields we cannot precompute, such as ICM epoch and timestamp, previous VM block seq, etc.

> Another possibility -> why rush to build the zeroth block? can't we just wait like the normal blocks?

so we can easily expand the validator set. Essentially, if we spin up a single node network, we can then add a node, add another node, add another node, all by interacting with the P-chain, and we will never need any user traffic. 

If we need user traffic to build the zero epoch, we're stuck.

msm/msm_test.go
samliok: yea makes sense for the approvalsRetriever. The comment was also asking more generally about the collection of helpers, and whether it made sense to keep them together in their own test files?

msm/msm.go
```diff
@@ -66,6 +131,183 @@ const (
...
+	simplexMetadataBytes := simplexMetadata.Bytes()
+	prevBlockSeq := simplexMetadata.Seq - 1
+
+	switch currentState {
+	case stateFirstSimplexBlock:
+		return sm.buildBlockZero(ctx, parentBlock, simplexMetadataBytes, simplexBlacklistBytes)
+	case stateBuildBlockNormalOp:
+		return sm.buildBlockNormalOp(ctx, parentBlock, simplexMetadataBytes, simplexBlacklistBytes, prevBlockSeq)
+	case stateBuildCollectingApprovals:
+		return sm.buildBlockCollectingApprovals(ctx, parentBlock, simplexMetadataBytes, simplexBlacklistBytes, prevBlockSeq)
```
samliok, line 205: what happens if we are collecting approvals for a validator set change `v1`, but the pchain height has advanced to a new validator set `v2`. am i correct to say the current code would first finish collecting approvals for `v1` and then transition to `v2` after all approvals are collected? 

would it be better to stop the transition of `v1` when we notice `v2`? curious about your thoughts

samliok (commented):
w

msm/msm.go
yacovm: Once we have the next P-chain reference height (we moved from normal block to collecting approvals) we are essentially "locked" on that P-chain height for the epoch transition and cannot change it. 

> am i correct to say the current code would first finish collecting approvals for v1 and then transition to v2 after all approvals are collected?

Yes, in such a case we'll have two epoch changes one after the other.

> would it be better to stop the transition of v1 when we notice v2? curious about your thoughts

So, I did think about this "problem" as well when I designed the MSM, but I don't think this is a use case we should optimize for. 

We might just be able to solve this problem passively by using P-chain from ICM epochs, because they change their P-chain height slowly every time frame (I think 30 seconds or so) and therefore it makes implicit batching of P-chain changes.

I'll look at the code once I re-introduce ICM epoching and see if I can make it more robust. 

Thanks for asking this 👍

msm/msm.go
yacovm: You asked me to remove the OuterBlock in a preceding review. To do that I will need to re-introduce it 🫠 

Let's just tackle this when we need it.

msm/msm_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes#diff-40d3ba649d92af619cbe7ad938b6df3a625bd7ae5ac31c03ed92ae667248d16dR39-R42

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R129-R136

msm/msm.go
yacovm: Removed it

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R148-R150

msm/msm_test.go
yacovm: I'll do it once I re-introduce verification as I removed the code for this PR

msm/msm_test.go
yacovm: tomato, tomato https://github.com/ava-labs/Simplex/pull/365/changes#diff-40d3ba649d92af619cbe7ad938b6df3a625bd7ae5ac31c03ed92ae667248d16dR110

msm/msm_test.go
yacovm: how can we have both of them nil though? I don't think such a thing is possible in the code.

msm/msm_test.go
yacovm: maybe we can do that after i re-introduce verification to do that in bulk.

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R752

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/dff27ead6d0098387cd1395544675f4b6d507166#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R154

msm/msm_test.go
yacovm: removed them

msm/msm_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/dff27ead6d0098387cd1395544675f4b6d507166#diff-40d3ba649d92af619cbe7ad938b6df3a625bd7ae5ac31c03ed92ae667248d16dL23-L887

msm/msm_test.go
yacovm: changed

msm/msm_test.go
yacovm: changed

msm/msm_test.go
yacovm: fixed

msm/msm.go
yacovm: fixed

msm/msm.go
yacovm: changed

msm/msm.go
yacovm: removed

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65L212

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R620

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R620

api.go
```diff
@@ -130,8 +130,33 @@ type SignatureAggregator interface {
...
+		nodes[i] = nw.Node
+	}
+	return nodes
+}
+
+// NodeWeight is a struct that pairs a node with its weight in the signature aggregator.
+type NodeWeight struct {
+	Node   NodeID
+	Weight uint64
+}
```
samliok: what are your thoughts on calling this `Node`? and having this type of struct


```go
type Node struct {
	ID   NodeID
	Weight uint64
}
```

this keeps it more generic, and i think `Node.ID` & `Node.Weight` is more clear than `NodeWeight.Weight` and `NodeWeight.Node`. Plus if we ever want to extend `Node` with more fields we wouldn't need a rename

api.go
```diff
@@ -130,8 +130,33 @@ type SignatureAggregator interface {
...
 	// In the case of PoA, this means at least a quorum of the nodes are given.
 	// In the case of PoS, this means at least two thirds of the st.
 	IsQuorum([]NodeID) bool
 }
+
+// NodeWeights is a list of NodeWeight elements.
+type NodeWeights []NodeWeight
+
+// NodesIDs returns the NodeIDs of the nodes in the NodeWeights.
+func (nws NodeWeights) NodesIDs() []NodeID {
```
samliok:
```suggestion
// NodeIDs returns the NodeIDs of the nodes in the NodeWeights.
func (nws NodeWeights) NodeIDs() []NodeID {
```

epoch.go
```diff
@@ -3420,10 +3424,10 @@ func (e *Epoch) nextSeqToCommit() uint64 {
 	return e.Storage.NumBlocks()
 }
 
-// SortNodes sorts the nodes in place by their byte representations.
-func SortNodes(nodes []NodeID) {
-	slices.SortFunc(nodes, func(a, b NodeID) int {
-		return bytes.Compare(a[:], b[:])
+// SortNodesWeights sorts the nodes in place by their byte representations.
+func SortNodesWeights(nodes NodeWeights) {
```
samliok:
```suggestion
func SortNodes(nodes Nodes) {
```

pos_test.go
samliok, line 16: should we create a PoA test where the nodes do not have equal weight?

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+		return nil, err
+	}
+	if simplexEpochInfo.BlockValidationDescriptor == nil {
+		simplexEpochInfo.BlockValidationDescriptor = &BlockValidationDescriptor{}
+	}
+	simplexEpochInfo.BlockValidationDescriptor.AggregatedMembership.Members = validators
+
+	// If this is not the first epoch, and this is the sealing block, we set the hash of the previous sealing block.
+	if simplexEpochInfo.EpochNumber > 1 {
+		prevSealingBlock, finalization, err := sm.GetBlock(simplexEpochInfo.EpochNumber, [32]byte{})
```
samliok, line 570: why do we have an empty digest?

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+	Verify(in verificationInput) error
+}
+
+// StateMachine manages block building and verification across epoch transitions.
+type StateMachine struct {
+	// verifiers is the list of verifiers used to verify proposed blocks.
+	// Each verifier is responsible for verifying a specific aspect of the block's metadata.
+	verifiers []verifier
+
+	*Config
```
samliok, line 92: should we make this a value rather than pointer? similar to how we inlined  `EpochConfig` in the `Epoch` struct

api.go
yacovm: we can rename, sure

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+	return sm.wrapBlock(parentBlock, childBlock, newSimplexEpochInfo, pChainHeight, simplexMetadata, simplexBlacklist), nil
+}
+
+// buildBlockZero builds the first ever block for Simplex,
+// which is a special block that introduces the first validator set and starts the first epoch.
+func (sm *StateMachine) buildBlockZero(parentBlock StateMachineBlock, simplexMetadata, simplexBlacklist []byte) (*StateMachineBlock, error) {
+	if sm.LastNonSimplexInnerBlock == nil {
+		sm.Logger.Error("Last non-Simplex inner block is nil, cannot build zero block with correct metadata")
+		return nil, fmt.Errorf("failed constructing zero block: last non-Simplex inner block is nil")
+	}
```
samliok: should we just check for this value to be present when we create the StateMachine in `NewStateMachine`? I think it makes sense to require this value and might as well error earlier rather than later. It would allow us to remove this check in `buildBlockZero` and `verifyBlockZero`.

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
-	// If the previous block has a sealing block sequence, it's a Telock.
-	// If it has a block validation descriptor, it's a sealing block.
-	// Either way, the epoch has been sealed.
-	if prevBlockSimplexEpochInfo.SealingBlockSeq > 0 || prevBlockSimplexEpochInfo.BlockValidationDescriptor != nil {
-		return stateBuildBlockEpochSealed
+func (sm *StateMachine) buildBlockAndMaybeTransitionEpoch(ctx context.Context,
+	parentBlock StateMachineBlock,
+	simplexMetadata []byte,
+	simplexBlacklist []byte,
+	childBlock VMBlock,
```
samliok: this variable is never used

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+	}
+
+	sm.Logger.Debug("Block building decision", zap.Stringer("decision", decisionToBuildBlock))
+
+	var childBlock VMBlock
+
+	switch decisionToBuildBlock {
+	case decisionBuild, decisionBuildAndTransitionEpoch:
+		// If we reached here, we need to build a new block, and maybe also transition to a new epoch.
+		return sm.buildBlockAndMaybeTransitionEpoch(ctx, parentBlock, simplexMetadata, simplexBlacklist, childBlock, decisionToBuildBlock, newSimplexEpochInfo, pChainHeight)
```
samliok: i feel like we can avoid creating `buildBlockAndMaybeTransitionEpoch` altogether

samliok: https://github.com/ava-labs/Simplex/pull/382

had claude create this to visualize my thoughts

yacovm: so basically collapse enums into a struct? sure I can do that

msm/msm.go
yacovm: fair

msm/msm.go
yacovm: I wish we could, that would break all tests. Look at [init](https://github.com/ava-labs/Simplex/pull/381/changes/d1e6ee72a0d1187f376ccb220c7473075b607f5d#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R244-R296) and see why.

msm/msm.go
yacovm: an empty digest means "we don't care what is the block digest". 

Since GetBlock may look into the finalized chain as well as the not-yet finalized chain suffix, we need to pass a digest if we need to address a specific block.

But the previous sealing block has to be finalized, so we don't need to pass a digest in this case.

pos_test.go
yacovm: I think we have different weights in `TestFakeNode`, no?

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+}
+
+// BuildBlock constructs the next block on top of the given parent block, and passes in the provided simplex metadata and blacklist.
+func (sm *StateMachine) BuildBlock(ctx context.Context, metadata simplex.ProtocolMetadata, blacklist *simplex.Blacklist) (*StateMachineBlock, error) {
+	// The zero sequence number is reserved for the genesis block, which should never be built.
+	if metadata.Seq == 0 {
+		return nil, fmt.Errorf("invalid ProtocolMetadata sequence number: should be > 0, got %d", metadata.Seq)
+	}
+
+	parentBlock, _, err := sm.GetBlock(metadata.Seq-1, metadata.Prev)
```
samliok: nit: we can define prevBlockSeq before this line and re-use it the logs, as well as here

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+	// We prepare information that is needed to compute the approvals for the new epoch,
+	// such as the validator set for the next epoch, and the approvals from peers.
+	validators, err := sm.GetValidatorSet(parentBlock.Metadata.SimplexEpochInfo.NextPChainReferenceHeight)
+	if err != nil {
+		return nil, err
+	}
+
+	// We retrieve approvals that validators have sent us for the next epoch.
+	// These approvals are signed by validators of the next epoch.
+	approvalsFromPeers := sm.ApprovalsRetriever.RetrieveApprovals()
```
samliok: at a high level, what kind of struct do you think will be holding these approvals and where do you think it will live? I'm assuming this struct will just be intercepting approval messages coming in from the eninge and collecting them?

msm/msm.go
yacovm: which one?

msm/msm.go
samliok: also, this approval retriever seems like it should catch any approvals that are not valid. Instead, we do this verification & filtering after. But I would assume we would also need some verification in the `ApprovalRetriever` itself so is there a way to make the `ApprovalRetriever` interface return already verified approvals? 

hard for me to say because i'm not sure how you envision implementing this interface

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
 	return sha256.Sum256(combined)
 }
 
-// Used to aggregate validator signatures for epoch transitions.
-type SignatureAggregator interface {
-	AggregateSignatures(signatures ...[]byte) ([]byte, error)
+// ApprovalsRetriever retrieves the approvals from validators of the next epoch for the epoch change.
+type ApprovalsRetriever interface {
+	RetrieveApprovals() ValidatorSetApprovals
```
```suggestion
	Approvals() ValidatorSetApprovals
```

this would make it the call site cleaner `ApprovalsRetriever.Approvals()`

msm/msm.go
samliok: also we would need a way to clean up old approvals once the epoch advances

msm/encoding.go
```diff
@@ -282,10 +289,10 @@ type ValidatorSetApproval struct {
 
 type ValidatorSetApprovals []ValidatorSetApproval
 
-func (vsa ValidatorSetApprovals) Filter(f func(int, ValidatorSetApproval) bool) ValidatorSetApprovals {
+func (vsa ValidatorSetApprovals) Filter(f func(int, ValidatorSetApproval, simplex.Logger) bool, logger simplex.Logger) ValidatorSetApprovals {
```
samliok: i don't think we ever use the `int` in the function `Filter`

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+	for i, nbm := range validators {
+		nodeID2ValidatorIndex[nbm.NodeID] = i
+	}
+
+	oldApprovalFromPeersCount := len(approvalsFromPeers)
+	// We have the approvals obtained from peers, but we need to sanitize them by filtering out approvals that are not valid,
+	// such as approvals that do not agree with our candidate auxiliary info digest and P-Chain height,
+	// and approvals that are from nodes that are not in the validator set or have already approved in prior blocks.
+	approvalsFromPeers = sanitizeApprovals(approvalsFromPeers, pChainHeight, nodeID2ValidatorIndex, oldApprovingNodes, logger)
+	logger.Debug("Santizied approvals after filtering out invalid approvals", zap.Int("numApprovalsBefore", oldApprovalFromPeersCount), zap.Int("numApprovalsAfter", len(approvalsFromPeers)))
```
```suggestion
	logger.Debug("Sanitized approvals after filtering out invalid approvals", zap.Int("numApprovalsBefore", oldApprovalFromPeersCount), zap.Int("numApprovalsAfter", len(approvalsFromPeers)))
```

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+		PrevVMBlockSeq:            prevVMBlockSeq,
+		SealingBlockSeq:           0,          // We don't have a sealing block in the zero block.
+		PrevSealingBlockHash:      [32]byte{}, // The zero block has no previous sealing block.
+		NextPChainReferenceHeight: 0,
+	}
+	return newSimplexEpochInfo
+}
+
+func computeNewApprovals(
+	nextEpochApprovals *NextEpochApprovals,
```
samliok: we pass in `prevNextEpochApprovals` to this function though?

samliok: i think `prevApprovals` makes more sense

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+		sm.Logger.Error("Error building block, building block without inner block instead", zap.Error(err))
+	}
+	if impatientContext.Err() != nil {
+		sm.Logger.Debug("Timed out waiting for block to be built, building block without inner block instead",
+			zap.Duration("elapsed", time.Since(start)), zap.Duration("maxBlockBuildingWaitTime", sm.MaxBlockBuildingWaitTime))
+	}
+	return sm.wrapBlock(parentBlock, childBlock, simplexEpochInfo, pChainHeight, simplexMetadata, simplexBlacklist), nil
+}
+
+func (sm *StateMachine) createSealingBlock(ctx context.Context, parentBlock StateMachineBlock, simplexMetadata []byte, simplexBlacklist []byte, simplexEpochInfo SimplexEpochInfo, newApprovals *approvals, pChainHeight uint64) (*StateMachineBlock, error) {
```
samliok: newApprovals is unused

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+		NextEpochApprovals:        nil, // We don't need to collect approvals to seal the first ever epoch.
+		PrevVMBlockSeq:            prevVMBlockSeq,
+		SealingBlockSeq:           0,          // We don't have a sealing block in the zero block.
+		PrevSealingBlockHash:      [32]byte{}, // The zero block has no previous sealing block.
+		NextPChainReferenceHeight: 0,
+	}
+	return newSimplexEpochInfo
+}
+
+func computeNewApprovals(
```
samliok, line 690: this function was a bit confusing at first, I think mainly the amount of `xxxApproval` names we have here. We have 3 different `ValidatorSetApprovals`, `NextEpochApprovals` and `approvals` structs that make it hard to follow.

Maybe we can just return the individual values rather than the `approvals` struct? it seems we aren't using the approvals struct anywhere else. If we plan on using it later, than its fine to keep it. 

Secondly, i think renaming `NextEpochApprovals` to `AggregatedApprovals` could be better. 

still a lot of different approval terms but its probably fine

msm/msm.go
```diff
@@ -36,62 +36,738 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+		nextEpochApprovals = &NextEpochApprovals{}
+	}
+
+	oldApprovingNodes := bitmaskFromBytes(nextEpochApprovals.NodeIDs)
+
+	// We map each validator to its relative index in the validator set.
+	nodeID2ValidatorIndex := make(map[nodeID]int)
+	for i, nbm := range validators {
+		nodeID2ValidatorIndex[nbm.NodeID] = i
+	}
```
samliok: would it make sense for this to be a helper of `NodeBLSMappings`? 

```go
// IndexByNodeID returns a lookup from NodeID to the validator's index in the set,
  // which is the position used by approval bitmasks.
  func (nbms NodeBLSMappings) IndexByNodeID() map[nodeID]int {
      result := make(map[nodeID]int, len(nbms))
      for i, nbm := range nbms {
          result[nbm.NodeID] = i
      }
      return result
  }
```

msm/msm.go
samliok: childBlock(we override it right away)

msm/msm.go
samliok: yea it also removes some of the enum cases like `contextCancelled` and then the requirement to decode the enum in `buildBlockAndMaybeTransitionEpoch`

msm/msm.go
samliok: makes sense, thats what i thought just wanted to confirm

msm/msm.go
yacovm: > at a high level, what kind of struct do you think will be holding these approvals and where do you think it will live? I'm assuming this struct will just be intercepting approval messages coming in from the eninge and collecting them?

Something like a specialized mempool that maps each nodeID to a list of up to `n` approvals, where `n` is the validator set size of the epoch. It will intercept approvals and check their signatures. 

Why up to `n`? Because each node can claim its P-chain height is something else and build the first block that moves us from normal block mode to collecting approvals mode. For each such P-chain height option, any validator should be able to vote for that P-chain height.


>  also, this approval retriever seems like it should catch any approvals that are not valid. Instead, we do this verification & filtering after. But I would assume we would also need some verification in the ApprovalRetriever itself so is there a way to make the ApprovalRetriever interface return already verified approvals?

Well, the only verification that we need to do before the approval hits the MSM is checking the signature. However we don't know what P-chain height the MSM would want to pick, right? So we can't filter out approvals incompatible with the P-chain height we want outside of the MSM. That's why we do it inside the MSM but we only do signature checks at the verification side.

> also we would need a way to clean up old approvals once the epoch advances

True, but that's irrelevant to this PR. But even if we just keep the approvals that correspond to the top P-chain height for each validator, it'll be fine as it is finite and small. Even if we have 100 validators it amounts to less than 1MB of data.

msm/msm.go
yacovm: whoops

msm/msm.go
yacovm: > Maybe we can just return the individual values rather than the approvals struct? it seems we aren't using the approvals struct anywhere else. If we plan on using it later, than its fine to keep it.

I'm not sure I understand. You're saying a function has a lot of input parameters and the solution is to increase the return parameters?

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/19001401796daaaa77904bb3dd5fedcdefa93606#diff-be01cb21dff5bfcd6691f445fd82589735f72839add32c29f657ae8bdf0d8863R237-R244

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/19001401796daaaa77904bb3dd5fedcdefa93606#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65L531

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/19001401796daaaa77904bb3dd5fedcdefa93606#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R691

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/19001401796daaaa77904bb3dd5fedcdefa93606#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R711

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/19001401796daaaa77904bb3dd5fedcdefa93606#diff-be01cb21dff5bfcd6691f445fd82589735f72839add32c29f657ae8bdf0d8863R302

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/19001401796daaaa77904bb3dd5fedcdefa93606#diff-195b56382a45f394194c255adc7097624af8fdfbfe8c0277186e598948046970R215

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/changes/19001401796daaaa77904bb3dd5fedcdefa93606#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R164-R168

msm/msm.go
samliok: eh, i think my comment doesnt make sense, going to resolve it

msm/encoding.go
samliok, line 66: is this only set when we are building telocks?

msm/msm.go
```diff
@@ -36,62 +36,707 @@ func (smb *StateMachineBlock) Digest() [32]byte {
...
+	newSimplexEpochInfo = SimplexEpochInfo{
+		// P-chain reference height is previous block's NextPChainReferenceHeight.
+		PChainReferenceHeight: parentBlock.Metadata.SimplexEpochInfo.NextPChainReferenceHeight,
+		// The epoch number is the sequence of the sealing block.
+		EpochNumber:    sealingBlockSeq,
+		PrevVMBlockSeq: computePrevVMBlockSeq(parentBlock, prevBlockSeq),
+	}
+
+	// TODO: This P-chain height should be taken from the ICM epoch
+	childBlock, err := sm.BlockBuilder.BuildBlock(ctx, sm.GetPChainHeight())
```
samliok, line 661: 🤔 we are building a new block for the epoch, but a couple things can change between epochs

1. validator set ordering -> this means we might have been the leader for seq `x` in epoch `1`, but now that the epoch has changed we may not be the leader for seq `x` in epoch `2`.
2. Our node may be kicked out of validator set, so we shouldn't be building a block anyways

msm/msm_test.go
```diff
@@ -4,26 +4,700 @@
...
+		{
+			name: "verifying a genesis block",
+			md:   validMD,
+			mutateBlock: func(block *StateMachineBlock) {
+				md, err := simplex.ProtocolMetadataFromBytes(block.Metadata.SimplexProtocolMetadata)
+				require.NoError(t, err)
+				md.Seq = 0
+				block.Metadata.SimplexProtocolMetadata = md.Bytes()
+			},
+			err: "attempted to build a genesis inner block",
```
samliok, line 46: i still think we should create error variables and use `ErrorIs` instead of `ErrorContains`

msm/msm.go
yacovm: > we are building a new block for the epoch, but a couple things can change between epochs

To be precise, we're building a block for a new epoch.



> 
> 1. validator set ordering -> this means we might have been the leader for seq `x` in epoch `1`, but now that the epoch has changed we may not be the leader for seq `x` in epoch `2`.
> 2. Our node may be kicked out of validator set, so we shouldn't be building a block anyways

A new epoch, a new Epoch instance, right? 

Why does it matter which node is the leader at a current round? The MSM assumes nothing about the relative position of the node within the validator set. Eventually there would be a non faulty node that will be a leader and will execute this line. When a node executes this line, it will already have the new epoch instance with the new validator set, right?

Keep in mind that the API of the MSM is very primitive - we can either BuildBlock or VerifyBlock. There is no assumption of any kind of continuation. We can only build a single block at any given round. 

So, if in the first time we execute this function we haven't sealed (finalized the sealing block) our epoch, we are running under the context of the previous epoch. Otherwise, If the epoch has been sealed then this is executed under the context of the new epoch, and therefore a node that has been evicted isn't expected to execute it (and even if it has, no one will accept such a block).

Does that make sense?

msm/encoding.go
yacovm: Yes. It's set once we build the first Telock and then we copy over the previous Telock's sealing block seq in induction.

msm/msm_test.go
yacovm: I agree but it makes sense to do that after I re-introduce verification, because most errors are there, and I prefer to do that in bulk

msm/encoding.go
yacovm: https://github.com/ava-labs/Simplex/pull/365/commits/5ef3c853985d3a115e01de5f87bd8b112c32b80d

msm/msm.go
samliok: ah makes sense, thanks for the explaination 👍

samliok (approved):

PR #368, opened 2026-04-26
Simplex reconfiguration framework - Part V (ICM epoch)

yacovm:
Re-introduce ICM epoch info
Persist ICM epoch information in block metadata and maintain it on the
StateMachine the same way the proposerVM does, so blocks encode which
P-chain height the ICM protocol should use.

Metadata / encoding:
  - Add ICMEpochInfo (canoto field 7) and AuxiliaryInfo (field 6) to the
    block metadata, with Equal/IsZero helpers.

ICM epoch computation:
  - Add ICMEpoch, ICMEpochInput and the ICMEpochTransition function type,
    exposed via Config.ComputeICMEpoch (replacing GetUpgrades).
  - Add computeICMEpochInfo, which derives the next epoch from the parent
    block's epoch, P-chain height and timestamps.
  - Build/verify of normal, collecting-approvals and sealing blocks now
    compute and carry the ICM epoch, and feed its P-chain height into
    BlockBuilder.BuildBlock and InnerBlock.Verify (removing the prior
    TODOs and the hard-coded pChainHeight).

  Zero block:
  - The zero block carries over its parent's ICM epoch unchanged (empty
    when the parent predates ICM); the first epoch then begins on the
    block built on top of it.
  - verifyBlockZero now validates the whole block by digest instead of
    comparing SimplexEpochInfo alone (bug fix!)

  Timestamps:
  - Add verifyTimestamp (rejects non-monotonic timestamps, timestamps too
    far in the future beyond maxSkew, and values exceeding int64) and
    apply it when verifying both zero and non-zero blocks.

  Fuzzing:
 - Adds a fuzz test for MSM verification


msm/encoding.go
```diff
@@ -30,10 +30,69 @@ type StateMachineMetadata struct {
 	PChainHeight uint64 `canoto:"uint,4"`
 	// Timestamp is the time when the block is being built, in milliseconds since Unix epoch.
 	Timestamp uint64 `canoto:"uint,5"`
+	// AuxiliaryInfo is application-specific information that the StateMachine doesn't need to understand,
+	// but can be used by applications that care about epoch changes, such as threshold distributed public key generation.
+	AuxiliaryInfo *AuxiliaryInfo `canoto:"pointer,6"`
```
samliok: do you think it makes sense to split up `AuxillaryInfo` and `ICMEpoch` info separate prs?

msm/encoding.go
```diff
@@ -30,10 +30,69 @@ type StateMachineMetadata struct {
...
+
+// AuxiliaryInfo defines application-specific information for applications that might care about epoch change,
+// such as threshold distributed public key generation.
+type AuxiliaryInfo struct {
+	// Info is opaque bytes that can be used by applications to encode any information that describes
+	// the current state for the application.
+	Info []byte `canoto:"bytes,1"`
+	// PrevAuxInfoSeq is a sequence number that applications can use to find previous AuxiliaryInfo in the chain.
+	// It is zero if this is the first AuxiliaryInfo for this epoch.
+	PrevAuxInfoSeq uint64 `canoto:"uint,2"`
```
samliok: Is there a reason to reset this to 0 once an epoch change is completed? what if the app info does not care about epoch changes?

msm/encoding.go
```diff
@@ -30,10 +30,69 @@ type StateMachineMetadata struct {
...
+
+	canotoData canotoData_AuxiliaryInfo
+}
+
+func (ai *AuxiliaryInfo) IsZero() bool {
+	var zero AuxiliaryInfo
+	return ai.Equal(&zero)
+}
+
+func (ai *AuxiliaryInfo) Equal(a *AuxiliaryInfo) bool {
```
samliok: we are only using `Equal` in `IsZero`. should we just create a simple zero check like 
`return ai != nil & ApplicationID == 0 && PrevAuxInfoSeq == 0 && len(info) == 0` and remove the `Equal` function? i'm down to keep `Equal` if we plan on using it later, but if not I dont think we should expose?

msm/msm.go
```diff
@@ -10,12 +10,17 @@ import (
...
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/ava-labs/simplex"
 	"go.uber.org/zap"
 )
 
+const (
+	maxSkew = 10 * time.Second
```
samliok, line 23: nit: slight comment on what maxSkew is?

msm/encoding.go
```diff
@@ -30,10 +30,69 @@ type StateMachineMetadata struct {
...
 
 	canotoData canotoData_StateMachineMetadata
 }
 
+// ICMEpochInfo is metadata used for the ICM protocol.
+// The StateMachine maintains this metadata in a similar fashion to proposerVM.
+type ICMEpochInfo struct {
+	EpochStartTime    uint64 `canoto:"uint,1"`
+	EpochNumber       uint64 `canoto:"uint,2"`
+	PChainEpochHeight uint64 `canoto:"uint,3"`
```
samliok, line 48: what is the `PChainEpochHeight`?

msm/msm.go
```diff
@@ -105,6 +112,41 @@ func (smb *StateMachineBlock) Digest() [32]byte {
 	return sha256.Sum256(combined)
 }
 
+// ICMEpoch defines the ICM epoch information that is maintained by the StateMachine and used for the ICM protocol.
+// The Statemachine maintains this information identically to how the proposerVM maintains this information,
+// and it does so by building the ICMEpochInput and then passing it into the StateMachine's ComputeICMEpoch function.
+type ICMEpoch struct {
```
samliok: why do we have this struct and the `ICMEEpochInfo`, why can't we just re-use the other struct?

msm/encoding.go
samliok: is there some helpful docs on how ICM epoching works? is there some sort of index between `pChainEpochHeight` and `icm` epochs?

msm/encoding.go
yacovm: Updated accordingly.

msm/msm.go
samliok: is there a variable defined for this in avalanchego? Just curious how this is handled there

msm/msm.go
```diff
@@ -314,6 +356,10 @@ func (sm *StateMachine) verifyNonZeroBlock(ctx context.Context, block, prevBlock
 	prevBlockMD := prevBlock.Metadata
 	currentState := prevBlockMD.SimplexEpochInfo.NextState()
 
+	if err := verifyTimestamp(block, prevBlock, sm.GetTime()); err != nil {
```
samliok: why do we use `GetTime` some places and `time.Now()` in others? also the epoch already has an `AdvanceTime` function that may now differ from time.Now and GetTime. Do you think there is clean way to have one source of truth for time?

msm/msm.go
```diff
@@ -813,16 +927,18 @@ func (sm *StateMachine) verifyCollectingApprovalsBlock(ctx context.Context, pare
 	approvals := bitmaskFromBytes(newApprovals.NodeIDs)
 	canSeal := sigAggr.IsQuorum(validators.SelectSubset(approvals))
 
-	// TODO: P-chain height should be taken from the ICM epoch. For now we pass the block proposer's P-chain height.
```
samliok, line 816: the comment says to take the pchain height from the ICM epoch but we are still taking it from the metadata?

msm/encoding.go
yacovm: https://build.avax.network/docs/acps/181-p-chain-epoched-views 

Long story short - the goal of the ICM epoching is to slow down P-chain height progression for ICM.
The way it works is that it splits the time continuum into chunks (epochs) where in each epoch, the P-chain height is the same, and is the P-chain height of the last block of the previous epoch.

msm/msm.go
yacovm: Because this is public API (used in `type ICMEpochTransition func(ICMEpochInput) ICMEpoch`) and isn't part of the wire format. It's a good principle to not tie together how we encode data in the block and the API to our package.

Having said that, I don't have a strong opinion about it and can change to just use the `ICMEEpochInfo` everywhere.

msm/msm.go
yacovm: we don't use `time.Now()` anywhere except from once for logging elapsed time: 

```
➜  msm git:(reconfig-6) grep -r "time.Now" *
fake_node_test.go:                      TS:          time.Now(),
fuzz_test.go:   startTime := time.Now().Truncate(time.Second)
msm.go: start := time.Now()
msm_test.go:                    TS:          time.Now(),
msm_test.go:            TS:          time.Now(),
msm_test.go:    startTime := time.Now().Truncate(time.Second)
msm_test.go:                    // we need the timestamps to be deterministic. So instead of using time.Now(), we use a fixed
msm_test.go:    now := time.Now()
msm_test.go:            InnerBlock: &InnerBlock{TS: time.Now(), BlockHeight: 1, Bytes: []byte{0xAA}},
msm_test.go:    tc.blockBuilder.block = &InnerBlock{TS: time.Now(), BlockHeight: 2, Bytes: []byte{0x01}}
msm_test.go:    tc.blockBuilder.block = &InnerBlock{TS: time.Now(), BlockHeight: 3, Bytes: []byte{0x02}}
msm_test.go:                    InnerBlock: &InnerBlock{TS: time.Now(), BlockHeight: 1, Bytes: []byte{0xAA}},
msm_test.go:            tc.blockBuilder.block = &InnerBlock{TS: time.Now(), BlockHeight: 2, Bytes: []byte{0x01}}
msm_test.go:            InnerBlock: &InnerBlock{TS: time.Now(), BlockHeight: 1, Bytes: []byte{0xAA}},
msm_test.go:            tc.blockBuilder.block = &InnerBlock{TS: time.Now(), BlockHeight: seq, Bytes: []byte{byte(seq)}}
util_test.go:                   TS:    time.Now(),
util_test.go:   startTime := time.Now().Add(-time.Duration(endHeight+2) * time.Second)
util_test.go:           GetTime:                         time.Now,
```

> also the epoch already has an AdvanceTime function that may now differ from time.Now and GetTime. Do you think there is clean way to have one source of truth for time?

The `GetTime()` is expected to be tied to the `AdvanceTime`.

msm/msm.go
yacovm: Why? we do the right thing: We call `verifyAgainstExpected` down below

and do: 

```
	if innerBlock != nil {
		if err := innerBlock.Verify(ctx, expectedIcmEpochInfo.PChainEpochHeight); err != nil {
			return err
		}
	}
```

msm/msm.go
yacovm: > nit: slight comment on what maxSkew is?

Ah, yeah I should comment what it means. I just took it as-is from proposerVM and forgot to comment.

> is there a variable defined for this in avalanchego? Just curious how this is handled there

Yes: https://github.com/ava-labs/avalanchego/blob/master/vms/proposervm/block.go#L29

msm/encoding.go
yacovm: We use it in the next PR in unit tests.

msm/encoding.go
yacovm: The app info is data that is only generated during epoch changes, so by definition it cares about epoch changes.

If the application doesn't care about epoch changes, then it can fold the app info stuff into the VM.

The motivation to reset this to 0 upon an epoch change, is that zero values are skipped when encoded to bytes, and we save space this way.

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/368/changes/bfb8ae87572f98fc06a8764692f497fdb968decf

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/368/changes/a92a0dc409489e30c5730c89ad4b3da700ff0c08

msm/msm.go
samliok: how are we going to keep this variable in sync with the other? what if a couple months down the line someone who doesn't know simplex changes that variable and not the simplex one?

yacovm: I don't understand why we need to keep them in sync with proposerVM. 

We need some kind of value, so I just used the value from proposerVM. 

why would they be in sync?

By the way, I made this configurable [here](https://github.com/ava-labs/Simplex/pull/368/commits/6bbe4c8d87af8d0ab99037f64af34c2b5c89b05b).

msm/fuzz_test.go
```diff
@@ -0,0 +1,299 @@
...
+//
+// The mutation is applied at the field level (rather than by flipping serialized bytes)
+// so the fuzzed block is always well-formed: byte-level mutations of the Canoto encoding
+// overwhelmingly corrupt the structure and merely exercise the decoder. Each fuzzed field
+// is reconstructed by the verifier from the parent block (or checked against an exact
+// value), so any real change is guaranteed to be rejected.
+func FuzzVerifyBlock(f *testing.F) {
+	for blockIdx := 0; blockIdx < numBuiltBlocks; blockIdx++ {
+		for fieldIdx := range authoritativeFields {
+			f.Add(blockIdx, fieldIdx, uint64(0))
```
samliok, line 83: i remember there was a pr of mine where you said we shouldn't be seeding the fuzzer. Is there a reason you're doing it here?

samliok (approved):

samliok (approved):
should we update the pr title though?

msm/fuzz_test.go
yacovm: Just giving it lower and upper bound values, I can remove it, it's not that interesting.

PR #381, opened 2026-05-12
Simplex reconfiguration framework - Part IV (MSM implementation - verification)

yacovm:


msm/msm.go
```diff
@@ -6,13 +6,37 @@ package metadata
...
 	"github.com/ava-labs/simplex"
 	"go.uber.org/zap"
 )
 
+var (
+	errLastNonSimplexInnerBlockNil       = errors.New("failed constructing zero block: last non-Simplex inner block is nil")
+	errInvalidProtocolMetadataSeq        = errors.New("invalid ProtocolMetadata sequence number: should be > 0")
+	errUnknownState                      = errors.New("unknown state")
+	errNilInnerBlock                     = errors.New("InnerBlock is nil")
+	errBuiltGenesisInnerBlock            = errors.New("attempted to build a genesis inner block")
```
samliok: bumping this comment again
https://github.com/ava-labs/Simplex/pull/365#discussion_r3202752668

msm/msm.go
```diff
@@ -245,6 +270,62 @@ func (sm *StateMachine) VerifyBlock(ctx context.Context, block *StateMachineBloc
 	return err
 }
 
+func (sm *StateMachine) init() {
```
samliok: potentially call `initVerifiers`?

msm/msm.go
samliok: should we just use `prevVMBlockSeq` defined on line 508

msm/msm.go
```diff
@@ -443,33 +524,33 @@ func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock Stat
...
 
 	if simplexEpochInfo.BlockValidationDescriptor == nil {
-		return fmt.Errorf("invalid BlockValidationDescriptor: should not be nil")
+		return errNilBlockValidationDescriptor
 	}
 
 	membership := simplexEpochInfo.BlockValidationDescriptor.AggregatedMembership.Members
 	if !NodeBLSMappings(membership).Equal(expectedValidatorSet) {
-		return fmt.Errorf("invalid BlockValidationDescriptor: should match validator set at P-chain height %d", pChainHeight)
+		return fmt.Errorf("%w: %d", errBlockValidationDescriptorMismatch, pChainHeight)
```
samliok: is this redundant? I think we check for this when we compare the simplexEpochInfo?

msm/msm.go
```diff
@@ -410,25 +491,25 @@ func (sm *StateMachine) buildBlockZero(parentBlock StateMachineBlock, simplexMet
 
 func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock StateMachineBlock) error {
 	if block == nil {
-		return fmt.Errorf("block is nil")
+		return errNilBlock
 	}
 
 	simplexEpochInfo := block.Metadata.SimplexEpochInfo
 
 	if simplexEpochInfo.EpochNumber != 1 {
```
samliok: i think this check is also unnecessary, since we are constructing the expected `simplexEpochInfo` later. Shouldn't we just have that one check on line 538 be where we validate `simplexEpochInfo`?

msm/msm.go
samliok: should we just use one variable, and set innerBlockTimestamp = 0 when there is no inner block?

msm/verification.go
```diff
@@ -0,0 +1,529 @@
...
+		return errInsufficientApprovals
+	}
+
+	return nil
+}
+
+func (nv *nextEpochApprovalsVerifier) verifyNormal(prev SimplexEpochInfo, next SimplexEpochInfo) error {
+	if prev.NextPChainReferenceHeight == 0 {
+		return nil
+	}
```
samliok: I still think we need to do some verification, otherwise a malicious node can make us finalize a block without us ever checking `NextEpochApprovals`.

See this test I asked claude to write.

```go
// TestNextEpochApprovalsVerifierHaltViaGarbageApprovals demonstrates the liveness halt
// that the missing nil-check in verifyNormal enables.
//
// Scenario:
//
//	A: Normal block, prev.NextPChainReferenceHeight == 0.
//	B: Normal block proposed by a malicious leader. B legitimately advances
//	   NextPChainReferenceHeight (validator set changed), and also injects an
//	   all-1s bitmask into NextEpochApprovals. With prev (A).NextPChainReferenceHeight
//	   == 0 the verifier early-returns and B is accepted.
//	B+1: Honest leader. prev (B).NextPChainReferenceHeight > 0 so verifyNormal runs
//	     the full path. areNextEpochApprovalsSignersSupersetOfApprovalsOfPrevBlock
//	     requires B+1 to be a superset of B's all-1s bitmask. No honest proposer
//	     can produce that, so no valid B+1 exists -> chain halts at B.
//
// This test will start failing at step (B) once verifyNormal is fixed to reject
// garbage approvals when prev.NextPChainReferenceHeight == 0, which is the desired
// behavior.
func TestNextEpochApprovalsVerifierHaltViaGarbageApprovals(t *testing.T) {
	validators := NodeBLSMappings{
		{BLSKey: []byte{1}, Weight: 1},
		{BLSKey: []byte{2}, Weight: 1},
		{BLSKey: []byte{3}, Weight: 1},
	}
	getValidators := func(uint64) (NodeBLSMappings, error) { return validators, nil }

	v := &nextEpochApprovalsVerifier{
		sigVerifier:          &testSigVerifier{},
		getValidatorSet:      getValidators,
		keyAggregator:        &testKeyAggregator{},
		sigAggregatorCreator: newSignatureAggregatorCreator(),
	}

	// Malicious B: triggers transition AND injects all-1s garbage approvals.
	garbageApprovals := &NextEpochApprovals{NodeIDs: []byte{0xFF}, Signature: []byte("garbage")}
	mdA := SimplexEpochInfo{NextPChainReferenceHeight: 0, PChainReferenceHeight: 50}
	mdB := SimplexEpochInfo{
		NextPChainReferenceHeight: 100,
		PChainReferenceHeight:     50,
		NextEpochApprovals:        garbageApprovals,
	}

	// Step 1: verify B against A. Today the verifier accepts B. After the fix this
	// should fail with errUnexpectedApprovals and the halt below becomes unreachable.
	errB := v.Verify(verificationInput{
		nextBlockType:   BlockTypeNormal,
		prevMD:          StateMachineMetadata{SimplexEpochInfo: mdA},
		proposedBlockMD: StateMachineMetadata{SimplexEpochInfo: mdB},
	})
	require.NoError(t, errB, "malicious B accepted today; once fixed, expect errUnexpectedApprovals")

	// Step 2: honest B+1 tries to continue. It supplies its own approval (bit 0),
	// which is the realistic best-case for an honest proposer that just observed
	// the transition. The superset check against B's all-1s bitmask rejects it.
	honestApprovals := &NextEpochApprovals{NodeIDs: []byte{0x01}, Signature: []byte("sig")}
	mdBPlus1 := SimplexEpochInfo{
		NextPChainReferenceHeight: 100,
		PChainReferenceHeight:     50,
		NextEpochApprovals:        honestApprovals,
	}

	errBPlus1 := v.Verify(verificationInput{
		nextBlockType:   BlockTypeNormal,
		prevMD:          StateMachineMetadata{SimplexEpochInfo: mdB},
		proposedBlockMD: StateMachineMetadata{SimplexEpochInfo: mdBPlus1},
	})
	require.ErrorIs(t, errBPlus1, errSignerSetShrunk, "chain is wedged: no honest B+1 can satisfy the superset check against B's garbage")
}
```

samliok: so i think we need this check

```go
	if prev.NextPChainReferenceHeight == 0 {
		if next.NextEpochApprovals != nil {
			return fmt.Errorf("%w but got %v", errUnexpectedApprovals, next.NextEpochApprovals)
		}
		return nil
	}
```

msm/verification.go
```diff
@@ -0,0 +1,529 @@
...
+	return nil
+}
+
+type epochNumberVerifier struct{}
+
+func (e *epochNumberVerifier) Verify(in verificationInput) error {
+	prev, next := in.prevMD.SimplexEpochInfo, in.proposedBlockMD.SimplexEpochInfo
+
+	// An epoch number of 0 means this is not a Simplex block, so the next block should be the first Simplex block with epoch number 1.
+	if in.prevMD.SimplexEpochInfo.EpochNumber == 0 && in.proposedBlockMD.SimplexEpochInfo.EpochNumber != 1 {
```
samliok: how can we ever reach this code path via the msm? if we are in this verifier that means the EpochNumber must be > 0. Is this just a defensive check?

msm/verification.go
```diff
@@ -0,0 +1,529 @@
...
+	if err != nil {
+		return fmt.Errorf("failed parsing protocol metadata: %w", err)
+	}
+
+	// Else, if the previous block has an inner block, we point to it.
+	// Otherwise, we point to the parent block's previous VM block seq.
+	prevBlock, _, err := v.getBlock(in.prevBlockSeq, md.Prev)
+	if err != nil {
+		return fmt.Errorf("failed retrieving block: %w", err)
+	}
```
samliok: this additional lookup is just to check is prevBlock.InnerBlock != nil. We can remove this lookup if we just pass in the `prevBlock` we already queried in `VerifyBlock`

https://github.com/ava-labs/Simplex/blob/d5dd2de24ec920dcb7d80417c62f51a3dca646f1/msm/msm.go#L261

msm/msm.go
```diff
@@ -245,6 +270,62 @@ func (sm *StateMachine) VerifyBlock(ctx context.Context, block *StateMachineBloc
...
+			timeSkewLimit: sm.TimeSkewLimit,
+			getTime: func() time.Time {
+				return sm.Config.GetTime()
+			},
+		},
+		&pChainReferenceHeightVerifier{},
+		&epochNumberVerifier{},
+		&validationDescriptorVerifier{
+			getValidatorSet: func(pChainHeight uint64) (NodeBLSMappings, error) {
+				return sm.Config.GetValidatorSet(pChainHeight)
```
samliok: would it make sense to just pass in the validator set for the `pChainHeight` in the verification input rather than passing in the method to both `validationDescriptorVerifier` and `nextPChainReferenceHeightVerifier`? I guess it would also save a lookup too

msm/verification.go
```diff
@@ -0,0 +1,529 @@
...
+func (s *sealingBlockSeqVerifier) Verify(in verificationInput) error {
+	prev, next := in.prevMD.SimplexEpochInfo, in.proposedBlockMD.SimplexEpochInfo
+
+	// A block should only have a sealing block if it is a Telock.
+	switch in.nextBlockType {
+	case BlockTypeNewEpoch, BlockTypeNormal, BlockTypeSealing:
+		if next.SealingBlockSeq != 0 {
+			return fmt.Errorf("%w: expected 0 but got %d", errSealingBlockSeqMismatch, next.SealingBlockSeq)
+		}
+	case BlockTypeTelock:
```
samliok: should we add a defensive check to make sure the telock case falls under one of these ifs?

```go
switch {
		case prev.SealingBlockSeq > 0:
			if next.SealingBlockSeq != prev.SealingBlockSeq {
				return fmt.Errorf("%w: expected %d but got %d", errSealingBlockSeqMismatch, prev.SealingBlockSeq, next.SealingBlockSeq)
			}
		case prev.BlockValidationDescriptor != nil:
			md, err := simplex.ProtocolMetadataFromBytes(in.prevMD.SimplexProtocolMetadata)
			if err != nil {
				return fmt.Errorf("failed parsing protocol metadata: %w", err)
			}
			if next.SealingBlockSeq != md.Seq {
				return fmt.Errorf("%w: expected %d but got %d", errSealingBlockSeqMismatch, md.Seq, next.SealingBlockSeq)
			}
		default:
			return errInvalidTelockPredecessor
```

msm/msm.go
yacovm: indeed

msm/msm.go
yacovm: I think the code is easier to understand if we explicitly say `hasInnerBlock` instead of having to remember relations like "if timestamp is 0 then we have no inner block".

samliok: is it better to construct the `verificationInput` once before the loop?

msm/misc_test.go
```diff
@@ -487,10 +488,66 @@ func (failingAggregator) Aggregate([]simplex.Signature) (simplex.QuorumCertifica
...
 func (failingAggregator) AppendSignatures([]byte, ...[]byte) ([]byte, error) {
-	return nil, fmt.Errorf("aggregation failed")
+	return nil, errTestAggregationFailed
 }
 
 func (failingAggregator) IsQuorum([]simplex.NodeID) bool {
 	return false
 }
+
+type testBlockStore map[uint64]StateMachineBlock
```
samliok: i dont think we should be testing anything other structs in `misc.go` in this file. When we bridge over these avalanchego utilites defined in `misc.go` this file should just be deleted.

Unfortunately I think this means moving everything starting from line 181 and onwards.

msm/msm.go
```diff
@@ -6,13 +6,45 @@ package metadata
...
 )
 
+var (
+	errLastNonSimplexInnerBlockNil     = errors.New("failed constructing zero block: last non-Simplex inner block is nil")
+	errInvalidProtocolMetadataSeq      = errors.New("invalid ProtocolMetadata sequence number: should be > 0")
+	errUnknownState                    = errors.New("unknown state")
+	errNilInnerBlock                   = errors.New("InnerBlock is nil")
+	errBuiltGenesisInnerBlock          = errors.New("received a genesis block")
+	errZeroBlockParentNoInnerBlock     = errors.New("failed constructing zero block: parent block has no inner block")
+	errNilBlock                        = errors.New("block is nil")
```
samliok: thoughts on `errNilBlock` and `errNilInnerBlock` being one error?

msm/msm.go
```diff
@@ -6,13 +6,45 @@ package metadata
...
 
 	"github.com/ava-labs/simplex"
 	"go.uber.org/zap"
 )
 
+var (
+	errLastNonSimplexInnerBlockNil     = errors.New("failed constructing zero block: last non-Simplex inner block is nil")
+	errInvalidProtocolMetadataSeq      = errors.New("invalid ProtocolMetadata sequence number: should be > 0")
+	errUnknownState                    = errors.New("unknown state")
+	errNilInnerBlock                   = errors.New("InnerBlock is nil")
```
samliok: https://github.com/ava-labs/Simplex/blob/7d9705de4e676948613295a0f4fee57d86b00ce5/msm/msm.go#L247-L250

i think this is being used wrongly here. maybe we should just remove it and use `errNilBlock` instead

msm/msm.go
```diff
@@ -6,13 +6,45 @@ package metadata
...
+	errSealingBlockSeqUnset            = errors.New("cannot build epoch sealed block: sealing block sequence is 0 or undefined")
+	errNilNextEpochApprovals           = errors.New("next epoch approvals is nil")
+)
+
+var (
+	errPChainReferenceHeightMismatch  = errors.New("unexpected P-chain reference height")
+	errPChainReferenceHeightDecreased = errors.New("P-chain reference height is decreasing")
+	errValidatorSetUnchanged          = errors.New("validator set unchanged; next P-chain reference height should not have advanced")
+	errPChainHeightNotReached         = errors.New("haven't reached referenced P-chain height yet")
+	errUnknownBlockType               = errors.New("unknown block type")
```
samliok: we should remove this error and just use `errUnknownState`

msm/msm.go
```diff
@@ -6,13 +6,45 @@ package metadata
...
 	"go.uber.org/zap"
 )
 
+var (
+	errLastNonSimplexInnerBlockNil     = errors.New("failed constructing zero block: last non-Simplex inner block is nil")
+	errInvalidProtocolMetadataSeq      = errors.New("invalid ProtocolMetadata sequence number: should be > 0")
+	errUnknownState                    = errors.New("unknown state")
+	errNilInnerBlock                   = errors.New("InnerBlock is nil")
+	errBuiltGenesisInnerBlock          = errors.New("received a genesis block")
+	errZeroBlockParentNoInnerBlock     = errors.New("failed constructing zero block: parent block has no inner block")
```
samliok: this plus `errParentInnerBlockHasNoInnerBlock` can be merged into one

msm/msm.go
```diff
@@ -6,13 +6,45 @@ package metadata
...
+	errZeroBlockHasInnerBlock          = errors.New("zero block must not have an inner block")
+	errZeroBlockInnerDigestMismatch    = errors.New("zero block inner block digest does not match last non-Simplex inner block digest")
+	errZeroBlockTimestampMismatch      = errors.New("zero block timestamp does not match last non-Simplex inner block timestamp")
+	errPrevSealingBlockNotFinalized    = errors.New("previous sealing InnerBlock is not finalized")
+	errFirstEverSimplexBlockNotSet     = errors.New("first ever Simplex block is not set, but attempted to create a sealing block for the first epoch")
+	errSealingBlockSeqUnset            = errors.New("cannot build epoch sealed block: sealing block sequence is 0 or undefined")
+	errNilNextEpochApprovals           = errors.New("next epoch approvals is nil")
+)
+
+var (
```
samliok: maybe combine this into one block, is there a reason you defined another block here? also there is one on here which could be combined into one block
https://github.com/ava-labs/Simplex/blob/7d9705de4e676948613295a0f4fee57d86b00ce5/msm/msm.go#L1150

msm/msm.go
```diff
@@ -240,49 +272,25 @@ func (sm *StateMachine) VerifyBlock(ctx context.Context, block *StateMachineBloc
 	case stateFirstSimplexBlock:
 		err = sm.verifyBlockZero(block, prevBlock)
 	default:
-		err = sm.verifyNonZeroBlock(ctx, block, prevBlock.Metadata, currentState, seq-1)
+		err = sm.verifyNonZeroBlock(ctx, block, &prevBlock, seq-1)
```
samliok, line 257: can we just flatten this function instead of creating a helper + another switch for the next 3 cases?

msm/msm.go
samliok, line 1148: can we add a small function comment for this function?

msm/msm.go
```diff
@@ -329,6 +337,137 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
...
+	}
+
+	expectedBlock := sm.wrapBlock(parentBlock, nextBlock.InnerBlock, newSimplexEpochInfo, proposedPChainHeight, nextBlock.Metadata.SimplexProtocolMetadata, nextBlock.Metadata.SimplexBlacklist)
+	if expectedBlock.Digest() != nextBlock.Digest() {
+		return fmt.Errorf("expected block digest %s does not match proposed block digest %s", expectedBlock.Digest(), nextBlock.Digest())
+	}
+	return nil
+}
+
+func verifyPChainHeight(proposedPChainHeight uint64, currentPChainHeight uint64, prevPChainHeight uint64) error {
```
samliok, line 399: maybe ask claude to create some trivial unit test for this. just for sanity

msm/msm.go
```diff
@@ -329,6 +337,137 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
...
+	if proposedPChainHeight > currentPChainHeight {
```
samliok, line 400: in the validators state we have two functions `GetMinimumHeight` and `GetCurrentHeight`. 

https://github.com/ava-labs/avalanchego/blob/5040e8fbadebc8e37e52e04caa278756427a5904/snow/validators/state.go#L33-L36

I'm assuming we need to be using `GetMinimumHeight` when sampling for pChain heights. can we note this down somewhere if so/not? maybe where we define `GetPChainHeight func() uint64`

msm/msm.go
```diff
@@ -329,6 +337,137 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
 	return sm.wrapBlock(parentBlock, innerBlock, newSimplexEpochInfo, decisionToBuildBlock.pChainHeight, simplexMetadata, simplexBlacklist), nil
 }
 
+func (sm *StateMachine) verifyNormalBlock(ctx context.Context, parentBlock StateMachineBlock, nextBlock *StateMachineBlock, prevBlockSeq uint64) error {
```
samliok, line 382: we never verify `SimplexEpochInfo.Epoch` = `block.ProtocolMetadata.Epoch`. i doubt they can deviate, but i'm nervous that we have two separate fields. 

Secondly, where do we set/update `block.ProtocolMetadata.Epoch`? `buildBlock` is passed in protocol metadata, but if we are transitioning the epoch number in `computeSimplexEpochInfoForNewEpoch`[here](https://github.com/ava-labs/Simplex/blob/7d9705de4e676948613295a0f4fee57d86b00ce5/msm/msm.go#L917) don't we also need to update the epoch in the protocol metadata?

msm/msm.go
samliok: i dont think this function should be part of `StateMachine`

msm/msm.go
samliok, line 382: thoughts on separating out some of the common verification code?

https://github.com/ava-labs/Simplex/compare/reconfig-3c...reconfig-seperations?expand=1

msm/msm.go
samliok, line 1: unrelated but doesn't this comment in `SimplexEpochInfo` need to be updated

```
// PrevSealingBlockHash is the hash of the sealing block of the previous epoch.
	// It is empty for the first epoch, and the second epoch has the PrevSealingBlockHash set to be
	// the hash of the first ever block built by the StateMachine.
	PrevSealingBlockHash [32]byte `canoto:"fixed bytes,3"`
```

msm/msm.go
```diff
@@ -575,22 +827,22 @@ func (sm *StateMachine) createSealingBlock(ctx context.Context, parentBlock Stat
 		prevSealingBlock, finalization, err := sm.GetBlock(simplexEpochInfo.EpochNumber, [32]byte{})
 		if err != nil {
 			sm.Logger.Error("Error retrieving previous sealing block", zap.Uint64("seq", simplexEpochInfo.EpochNumber), zap.Error(err))
-			return nil, fmt.Errorf("failed to retrieve previous sealing InnerBlock at epoch %d: %w", simplexEpochInfo.EpochNumber-1, err)
+			return SimplexEpochInfo{}, fmt.Errorf("failed to retrieve previous sealing InnerBlock at epoch %d: %w", simplexEpochInfo.EpochNumber-1, err)
```
samliok: why do we do EpochNumber - 1 here?

msm/msm.go
```diff
@@ -447,40 +582,159 @@ func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock Stat
...
+	aggPK, err := sm.aggregatePubKeysForBitmask(nodeIDsBitmask, validators)
+	if err != nil {
+		return err
+	}
+
+	pChainHeight := prev.NextPChainReferenceHeight
+	pChainHeightBuff := make([]byte, 8)
+	binary.BigEndian.PutUint64(pChainHeightBuff, pChainHeight)
+
+	if err := sm.SignatureVerifier.VerifySignature(next.NextEpochApprovals.Signature, pChainHeightBuff, aggPK); err != nil {
```
samliok: should we add a domain separator to this signature? similarly to what we do for quorum certificates?

msm/msm.go
yacovm: whoops, damn auto-complete

msm/msm.go
yacovm: yeah I was aware of it but kept postponing it because we only introduce signing at the next PR so didn't felt like it's needed at this point, but might as well do it now

msm/msm.go
yacovm: I already fixed it [here](https://github.com/ava-labs/Simplex/pull/388/changes/c1ffb03ba56af1825867fb726c525a3fed4eef78#diff-be01cb21dff5bfcd6691f445fd82589735f72839add32c29f657ae8bdf0d8863R49-R52).

msm/msm.go
yacovm: > we never verify SimplexEpochInfo.Epoch = block.ProtocolMetadata.Epoch. i doubt they can deviate, but i'm nervous that we have two separate fields.

Makes sense, added a check for now. 

In the future when we flatten the protocol metadata and canoto-tize it, we can remove the epoch number and we'll stay with only a single field

yacovm: > Secondly, where do we set/update block.ProtocolMetadata.Epoch? buildBlock is passed in protocol metadata, but if we are transitioning the epoch number in computeSimplexEpochInfoForNewEpoch[here](https://github.com/ava-labs/Simplex/blob/7d9705de4e676948613295a0f4fee57d86b00ce5/msm/msm.go#L917) don't we also need to update the epoch in the protocol metadata?

The Simplex Epoch is doing that. We are still missing a component that orchestrates restarting it once we transition epochs.

msm/msm.go
yacovm: No, this function is big enough, I don't want to make it even bigger.

samliok (commented):
There was one thing confusing me but I couldn't put my finger on it. I think the question that I have been stumped on is "How do we find the first ever simplex block?"

I know we have the `FirstEverSimplexBlock` callback defined in `StateMachine` but I don't get how this function would be implemented in practice. We would need to initialize our engine to find the first ever simplex block which is either going to be `seq 1` or `seq lastSnowmanBlock + 1`. If we started from genesis this is easy, but I'm unsure of how we plan on storing `lastSnowmanBlock`? Iterating through all the blocks to find this seems slow.

Also if I'm a non-validator looking to replicate epoch 1, I must validate the first ever simplex block before any other blocks in that epoch. But the question again is "how can i query for the first ever simplex block"? If another epoch has been created then I can just use `PrevSealingBlockHash`, but if no other epoch has been created what do I do?

```go
// PrevSealingBlockHash is the hash of the sealing block of the previous epoch.
	// It is empty for the first epoch, and the second epoch has the PrevSealingBlockHash set to be
	// the hash of the first ever block built by the StateMachine.
	PrevSealingBlockHash [32]byte `canoto:"fixed bytes,3"`
```

samliok (commented):
another question regarding the MSM + non validators. The MSM is to  "hijack"  `BlockBuilding` and `Verification`. For non-validators we do not care about `BlockBuilding` but we probably want to use the `MSM` for the same verification logic. Does it make sense to re-use the `MSM` struct for both non-validators and validators, even though `non-validators` do not care about `BlockBuilding`

samliok:
> There was one thing confusing me but I couldn't put my finger on it. I think the question that I have been stumped on is "How do we find the first ever simplex block?"
> 
> I know we have the `FirstEverSimplexBlock` callback defined in `StateMachine` but I don't get how this function would be implemented in practice. We would need to initialize our engine to find the first ever simplex block which is either going to be `seq 1` or `seq lastSnowmanBlock + 1`. If we started from genesis this is easy, but I'm unsure of how we plan on storing `lastSnowmanBlock`? Iterating through all the blocks to find this seems slow.
> 
> Also if I'm a non-validator looking to replicate epoch 1, I must validate the first ever simplex block before any other blocks in that epoch. But the question again is "how can i query for the first ever simplex block"? If another epoch has been created then I can just use `PrevSealingBlockHash`, but if no other epoch has been created what do I do?
> 
> ```go
> // PrevSealingBlockHash is the hash of the sealing block of the previous epoch.
> 	// It is empty for the first epoch, and the second epoch has the PrevSealingBlockHash set to be
> 	// the hash of the first ever block built by the StateMachine.
> 	PrevSealingBlockHash [32]byte `canoto:"fixed bytes,3"`
> ```

It would be nice in some regards(and bad in others) if the first ever epoch was the sequence number of the first simplex block. In the `genesis` case the first epoch will be `epoch 1`. In the case of 100 snowman blocks -> simplex, the first epoch will be `epoch 101` because the sequence of the first simplex block is `101`. 

This would be nice because if we are in the first epoch we can now know which sequence number to fetch to get to the start of the epoch. However, because the epoch number is not `1`, we don't know for sure if we are in the first epoch. This is an easy problem because we can just check if that sequence is a sealing block, or check if `PrevSealingBlockHash` is empty & `BlockValidatorDescriptor` is present. In these cases we know we have hit the first ever simplex block. 

Another nice benefit of this change is the branching in `createSealingBlock` can go away PLUS the weird callback that I was confused about in my earlier comment. This is because the logic in the `if simplexEpochInfo.EpochNumber > 1` is applicable to both scenarios

```go
	// If this is not the first epoch, and this is the sealing block, we set the hash of the previous sealing block.
	if simplexEpochInfo.EpochNumber > 1 {
		prevSealingBlock, finalization, err := sm.GetBlock(simplexEpochInfo.EpochNumber, [32]byte{})
		if err != nil {
			sm.Logger.Error("Error retrieving previous sealing block", zap.Uint64("seq", simplexEpochInfo.EpochNumber), zap.Error(err))
			return nil, fmt.Errorf("failed to retrieve previous sealing InnerBlock at epoch %d: %w", simplexEpochInfo.EpochNumber-1, err)
		}
		if finalization == nil {
			sm.Logger.Error("Previous sealing block is not finalized", zap.Uint64("seq", simplexEpochInfo.EpochNumber))
			return nil, fmt.Errorf("previous sealing InnerBlock at epoch %d is not finalized", simplexEpochInfo.EpochNumber-1)
		}
		simplexEpochInfo.PrevSealingBlockHash = prevSealingBlock.Digest()
	} else { // Else, this is the first epoch, so we use the hash of the first ever Simplex block.
		firstSimplexBlock := sm.FirstEverSimplexBlock()
		if firstSimplexBlock == nil {
			return nil, fmt.Errorf("first ever Simplex block is not set, but attempted to create a sealing block for the first epoch")
		}
		simplexEpochInfo.PrevSealingBlockHash = firstSimplexBlock.Digest()
	}
```


These are just my thoughts, not a concrete suggestion but they could be worth thinking about.

yacovm:
> another question regarding the MSM + non validators. The MSM is to "hijack" `BlockBuilding` and `Verification`. For non-validators we do not care about `BlockBuilding` but we probably want to use the `MSM` for the same verification logic. Does it make sense to re-use the `MSM` struct for both non-validators and validators, even though `non-validators` do not care about `BlockBuilding`

I am not sure why a non-validator should use the verification function of the MSM. It checks the state transition is legal, but there is no need to do that because as a non-validator, you only care about finalized blocks, which anyway carry an aggregated signature.

yacovm:
> > There was one thing confusing me but I couldn't put my finger on it. I think the question that I have been stumped on is "How do we find the first ever simplex block?"
> > I know we have the `FirstEverSimplexBlock` callback defined in `StateMachine` but I don't get how this function would be implemented in practice. We would need to initialize our engine to find the first ever simplex block which is either going to be `seq 1` or `seq lastSnowmanBlock + 1`. If we started from genesis this is easy, but I'm unsure of how we plan on storing `lastSnowmanBlock`? Iterating through all the blocks to find this seems slow.
> > Also if I'm a non-validator looking to replicate epoch 1, I must validate the first ever simplex block before any other blocks in that epoch. But the question again is "how can i query for the first ever simplex block"? If another epoch has been created then I can just use `PrevSealingBlockHash`, but if no other epoch has been created what do I do?
> > ```go
> > // PrevSealingBlockHash is the hash of the sealing block of the previous epoch.
> > 	// It is empty for the first epoch, and the second epoch has the PrevSealingBlockHash set to be
> > 	// the hash of the first ever block built by the StateMachine.
> > 	PrevSealingBlockHash [32]byte `canoto:"fixed bytes,3"`
> > ```
> 
> It would be nice in some regards(and bad in others) if the first ever epoch was the sequence number of the first simplex block. In the `genesis` case the first epoch will be `epoch 1`. In the case of 100 snowman blocks -> simplex, the first epoch will be `epoch 101` because the sequence of the first simplex block is `101`.
> 
> This would be nice because if we are in the first epoch we can now know which sequence number to fetch to get to the start of the epoch. However, because the epoch number is not `1`, we don't know for sure if we are in the first epoch. This is an easy problem because we can just check if that sequence is a sealing block, or check if `PrevSealingBlockHash` is empty & `BlockValidatorDescriptor` is present. In these cases we know we have hit the first ever simplex block.
> 
> Another nice benefit of this change is the branching in `createSealingBlock` can go away PLUS the weird callback that I was confused about in my earlier comment. This is because the logic in the `if simplexEpochInfo.EpochNumber > 1` is applicable to both scenarios
> 
> ```go
> 	// If this is not the first epoch, and this is the sealing block, we set the hash of the previous sealing block.
> 	if simplexEpochInfo.EpochNumber > 1 {
> 		prevSealingBlock, finalization, err := sm.GetBlock(simplexEpochInfo.EpochNumber, [32]byte{})
> 		if err != nil {
> 			sm.Logger.Error("Error retrieving previous sealing block", zap.Uint64("seq", simplexEpochInfo.EpochNumber), zap.Error(err))
> 			return nil, fmt.Errorf("failed to retrieve previous sealing InnerBlock at epoch %d: %w", simplexEpochInfo.EpochNumber-1, err)
> 		}
> 		if finalization == nil {
> 			sm.Logger.Error("Previous sealing block is not finalized", zap.Uint64("seq", simplexEpochInfo.EpochNumber))
> 			return nil, fmt.Errorf("previous sealing InnerBlock at epoch %d is not finalized", simplexEpochInfo.EpochNumber-1)
> 		}
> 		simplexEpochInfo.PrevSealingBlockHash = prevSealingBlock.Digest()
> 	} else { // Else, this is the first epoch, so we use the hash of the first ever Simplex block.
> 		firstSimplexBlock := sm.FirstEverSimplexBlock()
> 		if firstSimplexBlock == nil {
> 			return nil, fmt.Errorf("first ever Simplex block is not set, but attempted to create a sealing block for the first epoch")
> 		}
> 		simplexEpochInfo.PrevSealingBlockHash = firstSimplexBlock.Digest()
> 	}
> ```
> 
> These are just my thoughts, not a concrete suggestion but they could be worth thinking about.

Yeah I like this idea. 👍

msm/msm.go
yacovm: I agree we should use`GetMinHeight` but why should we note this down here? we should not care in the implementation of Simplex where does an implementation of an interface / function pointer live.

msm/msm.go
```diff
@@ -415,29 +605,25 @@ func (sm *StateMachine) buildBlockZero(parentBlock StateMachineBlock, simplexMet
 
 func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock StateMachineBlock) error {
 	if block == nil {
-		return fmt.Errorf("block is nil")
+		return errNilBlock
```
samliok: we have this check already in `verifyBlock`

msm/msm.go
```diff
@@ -6,13 +6,45 @@ package metadata
...
 	"fmt"
 	"time"
 
 	"github.com/ava-labs/simplex"
 	"go.uber.org/zap"
 )
 
+var (
+	errLastNonSimplexInnerBlockNil    = errors.New("failed constructing zero block: last non-Simplex inner block is nil")
+	errInvalidProtocolMetadataSeq     = errors.New("invalid ProtocolMetadata sequence number: should be > 0")
```
samliok, line 21: we never check for this error. should we combine 

```go
	errInvalidProtocolMetadataSeq     = errors.New("invalid ProtocolMetadata sequence number: should be > 0")
	errInvalidSimplexEpochInfo        = errors.New("invalid SimplexEpochInfo")
```
into one error maybe `errSeqIsGenesis`?

msm/msm.go
```diff
@@ -300,6 +314,19 @@ func (sm *StateMachine) buildBlockNormalOp(ctx context.Context, parentBlock Stat
...
+	var isSealingBlockFinalized bool
+	sealingBlockSeq := parentBlock.Metadata.SimplexEpochInfo.EpochNumber
+	_, finalization, err := sm.GetBlock(sealingBlockSeq, [32]byte{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve sealing block for previous epoch (%d): %w", sealingBlockSeq, err)
+	}
+	if finalization != nil {
+		isSealingBlockFinalized = true
+	} else {
+		sm.Logger.Debug("Previous sealing block not finalized yet, "+
```
samliok: can we combine this log with the one on line 336? The `sealingBlockSeq` log is just the epoch number which we log before this call.

```go
	sm.Logger.Debug("Block building decision",
		zap.Bool("build inner block", decisionToBuildBlock.buildInnerBlock),
		zap.Bool("transition epoch", decisionToBuildBlock.transitionEpoch),
		zap.Uint64("P-chain height", decisionToBuildBlock.pChainHeight),
		zap.Bool("SealingBlockFinalized", isSealingBlockFinalized))
```

msm/msm.go
```diff
@@ -300,6 +314,19 @@ func (sm *StateMachine) buildBlockNormalOp(ctx context.Context, parentBlock Stat
 
 // buildBlockOrTransitionEpoch builds a block and decides whether to transition to a new epoch based on the P-chain height and validator set changes.
 func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentBlock StateMachineBlock, simplexMetadata, simplexBlacklist []byte, newSimplexEpochInfo SimplexEpochInfo) (*StateMachineBlock, error) {
+	var isSealingBlockFinalized bool
+	sealingBlockSeq := parentBlock.Metadata.SimplexEpochInfo.EpochNumber
+	_, finalization, err := sm.GetBlock(sealingBlockSeq, [32]byte{})
```
samliok, line 319: no action: BlockBuilding runs in its own go-routine, therefore it's possible the sealing block was not finalized when this was called but become shortly after. In this case, I think we will fail verifying the block later on which should be fine

msm/msm.go
```diff
@@ -311,7 +338,7 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
 		zap.Bool("transition epoch", decisionToBuildBlock.transitionEpoch),
 		zap.Uint64("P-chain height", decisionToBuildBlock.pChainHeight))
 
-	if decisionToBuildBlock.transitionEpoch {
+	if decisionToBuildBlock.transitionEpoch && isSealingBlockFinalized {
```
samliok, line 338: i'm not sure i get this change though. Don't we build telocks if the sealing block is not finalized?

samliok: regardless, should we use `isSealingBlockFinalizaed()` instead ?

msm/msm.go
```diff
@@ -240,49 +254,49 @@ func (sm *StateMachine) VerifyBlock(ctx context.Context, block *StateMachineBloc
...
+	case stateBuildBlockEpochSealed:
+		return sm.verifyBlockEpochSealed(ctx, *prevBlock, block, prevSeq)
+	default:
+		return fmt.Errorf("%w: %d", errUnknownState, currentState)
 	}
+}
 
-	return block.InnerBlock.Verify(ctx)
+func (sm *StateMachine) verifyEpochNumber(block *StateMachineBlock) error {
+	md, err := simplex.ProtocolMetadataFromBytes(block.Metadata.SimplexProtocolMetadata)
```
samliok, line 292: we already fetch the protocol metadata from bytes on line `232`. should we just pass that in here instead of re-parsing?

msm/msm.go
```diff
@@ -630,47 +909,103 @@ func (sm *StateMachine) buildBlockEpochSealed(ctx context.Context, parentBlock S
...
 	// Else, we build a block for the new epoch.
+	newSimplexEpochInfo = computeSimplexEpochInfoForNewEpoch(newSimplexEpochInfo, parentBlock, sealingBlockSeq, prevBlockSeq)
+
+	// TODO: This P-chain height should be taken from the ICM epoch
+
+	return sm.buildBlockOrTransitionEpoch(ctx, sealingBlock, simplexMetadata, simplexBlacklist, newSimplexEpochInfo)
+
+}
+
+func computeSimplexEpochInfoForNewEpoch(newSimplexEpochInfo SimplexEpochInfo, parentBlock StateMachineBlock, sealingBlockSeq uint64, prevBlockSeq uint64) SimplexEpochInfo {
```
samliok: why are we updating `newSimplexEpochInfo` and also returning it?

msm/msm.go
```diff
@@ -630,47 +909,103 @@ func (sm *StateMachine) buildBlockEpochSealed(ctx context.Context, parentBlock S
...
+	// We check if the sealing block has already been finalized.
+	// If not, we build a Telock block.
+	isSealingBlockFinalized, sealingBlockSeq, sealingBlock, err := sm.isSealingBlockFinalized(parentBlock, prevBlockSeq)
 	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve sealing block at sequence %d: %w", sealingBlockSeq, err)
+		return nil, err
 	}
 
-	isSealingBlockFinalized := finalization != nil
+	newSimplexEpochInfo := computeSimplexEpochInfoForTelock(parentBlock, sealingBlockSeq, prevBlockSeq)
```
samliok: should we just set this if the block is a tealock? i.e in the if statement right below?

samliok (commented):
mainly cosmetic things, except for the question about telocks. probably ready to be approved once resolved

msm/msm.go
yacovm: > i'm not sure i get this change though. Don't we build telocks if the sealing block is not finalized?

This is actually to make sure that the zero block is finalized before we start transitioning to a new epoch. 

But looking at this, I think we need to add a check for that in the verification code + a comment, because this is indeed unclear. 



> regardless, should we use isSealingBlockFinalizaed() instead ?

Not sure what you mean by `()`, can you elaborate?

msm/msm.go
yacovm: why would we fail verifying the block later on? If the sealing block wasn't finalized when this is called, then we will just not start the epoch transition. It's perfectly legal to not opt-in and wait for some other node to start it.

msm/msm.go
yacovm: yeah we anyway override it.

msm/msm.go
yacovm: artifact of refactoring, will fix.

msm/msm.go
yacovm: yeah but then I'll need to ripple that all over the place.

I think it's actually easier to keep it as is for now, because once we canoto-tize the protocol metadata, we won't store them as bytes anymore in the `block.Metadata.SimplexProtocolMetadata` but as regular fields.

msm/msm.go
samliok: i was referring to the function rather than the variable
```
func (sm *StateMachine) isSealingBlockFinalized(parentBlock StateMachineBlock, prevBlockSeq uint64) (bool, uint64, StateMachineBlock, error) {
```

yacovm: ah no, we can't use it. That function can only be used if we're sealing the epoch. Look at its implementation. 

I guess I should clarify it in the comments, otherwise it's confusing.

msm/msm.go
yacovm: I added a test. I don't think we should combine as they are errors that stem from different reasons - the first is illegal input from the upper layer and the second is illegal input from an incorrect node.

msm/msm.go
yacovm: I'll just remove it

msm/msm.go
yacovm: > But looking at this, I think we need to add a check for that in the verification code + a comment, because this is indeed unclear.

I'm wrong, we do verify it.

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/381/changes/bac903dc107734c4b2472e89f52ca4056976dfae#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R932

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/381/changes/bac903dc107734c4b2472e89f52ca4056976dfae#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R940

msm/msm.go
yacovm: Added a test: https://github.com/ava-labs/Simplex/pull/381/changes/bac903dc107734c4b2472e89f52ca4056976dfae#diff-40d3ba649d92af619cbe7ad938b6df3a625bd7ae5ac31c03ed92ae667248d16dR898-R975

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/381/changes/bac903dc107734c4b2472e89f52ca4056976dfae#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65L325-L327

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/381/changes/bac903dc107734c4b2472e89f52ca4056976dfae#diff-40d3ba649d92af619cbe7ad938b6df3a625bd7ae5ac31c03ed92ae667248d16dR206-R213

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/381/changes/bac903dc107734c4b2472e89f52ca4056976dfae#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65L607-L610

msm/msm.go
```diff
@@ -326,7 +350,170 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
 		}
 	}
 
-	return sm.wrapBlock(parentBlock, innerBlock, newSimplexEpochInfo, decisionToBuildBlock.pChainHeight, simplexMetadata, simplexBlacklist), nil
+	return wrapBlock(parentBlock, innerBlock, newSimplexEpochInfo, decisionToBuildBlock.pChainHeight, simplexMetadata, simplexBlacklist), nil
+}
+
+func verifyAgainstExpected(
```
samliok, line 356: is it worth having tests for this function?

msm/msm.go
samliok: yea i don't think we should note `getMinHeight` particularly, but it might be worth noting we need a `pChainHeight` all nodes agree on.

samliok (approved):

PR #384, opened 2026-05-18
Implement approval store

yacovm:
Also self-sign during aggregation

- Introduces ApprovalStore (msm/approvals.go) — an in-memory store of validator approvals for epoch transitions, keyed by (NodeID, PChainHeight). It verifies BLS signatures on
  ingest, deduplicates by timestamp (newer wins, older is dropped), and prunes per-node entries to at most len(validators) by evicting the oldest timestamp.
- Adds Timestamp to ValidatorSetApproval (msm/encoding.go) so the store can order/evict approvals deterministically.
- computeNewApprovals (msm/msm.go) now optimistically self-signs the next epoch's P-chain reference height each round and appends its own ValidatorSetApproval to the peer set since the store deduplicates it later.


msm/msm.go
samliok, line 83: i just realized we have this interface already defined in `simplex.api`. thoughts on removing this and just using the simplex interface(we already do this for the `Signer` interface)?

samliok, line 104: this is unused

msm/util_test.go
```diff
@@ -282,6 +282,10 @@ func newStateMachine(t *testing.T) (*StateMachine, *testConfig) {
 	bs := make(blockStore)
 	bs[0] = &outerBlock{block: genesisBlock}
 
+	var myNodeID nodeID
```
samliok, line 285: is there any benefit of making this random?

msm/msm_test.go
```diff
@@ -1175,3 +1175,87 @@ func TestComputeNewApproverSignaturesAndSigners(t *testing.T) {
...
+	sm.SignatureAggregatorCreator = func(_ []simplex.Node) simplex.SignatureAggregator {
+		return concatAggregator{}
+	}
+
+	// Place MyNodeID at index 0 of a 3-node validator set so quorum is not
+	// reachable from a single approval (canSeal stays false on both rounds).
+	var myID nodeID
+	copy(myID[:], sm.MyNodeID)
+	validators := NodeBLSMappings{
+		{NodeID: myID, BLSKey: []byte{1}, Weight: 1},
```
samliok: nit
```go
validators := NodeBLSMappings{
{NodeID: nodeID(sm.MyNodeID), BLSKey: []byte{1}, Weight: 1}
```

also did github remove the suggestions option on pr review? can't seem to find it

msm/approvals.go
```diff
@@ -0,0 +1,139 @@
...
+	"encoding/binary"
+	"fmt"
+
+	"github.com/ava-labs/simplex"
+	"go.uber.org/zap"
+)
+
+type approvalsByPChainHeight map[uint64]*ValidatorSetApproval
+
+type ApprovalStore struct {
```
samliok, line 22: Is this supposed to be passed into the MSM somewhere? I don't see it connected anywhere outside of tests

msm/approvals.go
```diff
@@ -0,0 +1,139 @@
...
+				zap.String("nodeID", fmt.Sprintf("%x", oldestApproval.NodeID)),
+				zap.String("oldestApprovalPChainHeight",
+					fmt.Sprintf("%d", oldestApproval.PChainHeight)), zap.Uint64("oldestApprovalTimestamp", oldestApproval.Timestamp))
+			delete(as.approvalsByNodes[approval.NodeID], oldestApproval.PChainHeight)
+			as.storedCount--
+		}
+	}
+}
+
+func (as *ApprovalStore) checkApprovalSignature(approval *ValidatorSetApproval, pk []byte) error {
```
samliok, line 125: hmm is it not a problem that we verify the signature is valid over just the `pChainHeight` here, but in the MSM code we verify it with the context + asn1 encoding? shouldn't there be one canonical way of verifying/signing that is shared for both of them?

msm/approvals.go
```diff
@@ -0,0 +1,139 @@
...
+	approvals := make(ValidatorSetApprovals, 0, as.storedCount)
+	for _, approvalsByHeight := range as.approvalsByNodes {
+		for _, approval := range approvalsByHeight {
+			approvals = append(approvals, *approval)
+		}
+	}
+	return approvals
+}
+
+func (as *ApprovalStore) HandleApproval(approval *ValidatorSetApproval) error {
```
samliok: where is `HandleApproval` going to be called? Do we need to make this handling method thread safe?

msm/approvals.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+func (as *ApprovalStore) checkApprovalSignature(approval *ValidatorSetApproval, pk []byte) error {
+	pChainHeight := approval.PChainHeight
+	pChainHeightBuff := make([]byte, 8)
+	binary.BigEndian.PutUint64(pChainHeightBuff, pChainHeight)
+
+	// We check if the signature is valid before we store the approval.
+	return as.signatureVerifier.VerifySignature(approval.Signature, pChainHeightBuff, pk)
```
samliok: we dont we also check the signature is correct over the `AuxInfoSeqDigest`? I guess what is stopping nodes from agreeing on the pchain signature, but having different hashes in `AuxInfoSeqDigest`

msm/approvals_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"errors"
+	"math"
+	"testing"
+
+	"github.com/ava-labs/simplex/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func makeNodeID(seed byte) nodeID {
+	var n nodeID
```
samliok, line 16: i'm just realizing this now but `nodeID` will get very annoying imo. we already have `simplex.NodeID` and now we are adding `nodeID` which is different. Maybe in a separate pr can we change `nodeID` to something like `avagoNodeID` and merge that pr in before this one?

msm/approvals_test.go
```diff
@@ -0,0 +1,283 @@
...
+	require.NoError(t, as.HandleApproval(&ValidatorSetApproval{
+		NodeID:       makeNodeID(99),
+		PChainHeight: 1,
+		Timestamp:    1,
+	}))
+	require.Empty(t, as.Approvals())
+	require.Equal(t, 0, as.storedCount)
+}
+
+func TestApprovalStoreHandleApprovalInvalidSignature(t *testing.T) {
```
samliok: thoughts on https://github.com/ava-labs/Simplex/pull/392

feel free to close/merge/implement

msm/msm.go
yacovm: But the interface we have there takes in a `signer []NodeID` where here we take a public key instead.

yacovm: Yeah, I fixed it here: https://github.com/ava-labs/Simplex/pull/381/changes#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65L72-L84

msm/approvals.go
yacovm: Yeah, this is what implements:

```
// ApprovalsRetriever retrieves the approvals from validators of the next epoch for the epoch change.
type ApprovalsRetriever interface {
	Approvals() ValidatorSetApprovals
}
```

msm/approvals.go
yacovm: I will connect it somehow once I build the orchestration layer. 

For now, I just implemented this logic without paying attention to thread safety as it's not connected to anything yet

msm/approvals.go
yacovm: Yeah, because when I implemented this, I still haven't done that domain separation in the MSM. 

Will fix 👍

msm/approvals.go
yacovm: That's a good question.

So, it's true that in the spec I wrote that we check the signature on both the P-chain height and the AuxInfoDigest. 

However, since you asked to remove the code that handles the AuxInfo from the MSM PR and I have yet to introduce it back, I am signing for now only on the P-chain height.

msm/approvals_test.go
yacovm: Sure we can align the NodeID definitions. 

But why do that before merging this PR? I think it'll be easier just doing it afterwards, no?

We anyway will need to do this before the repo is swallowed by avalanchego

msm/util_test.go
yacovm: No.

msm/approvals_test.go
yacovm: I think my version is a bit more clear on what's going on https://github.com/ava-labs/Simplex/pull/384/changes/6718a90a20bd2561227271b64ff11315c43d1271#diff-bee276fee4ff263e239cceb097391399585c4218b30ab371dfc6e08e18e19956R4-R255

msm/approvals.go
yacovm: https://github.com/ava-labs/Simplex/pull/384/changes/6718a90a20bd2561227271b64ff11315c43d1271#diff-5e383ab850834861309c7e77b4123dbe73b6b4d3e2b96f96832869f5847b48fdR117-R125

msm/msm_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/384/changes/6718a90a20bd2561227271b64ff11315c43d1271#diff-40d3ba649d92af619cbe7ad938b6df3a625bd7ae5ac31c03ed92ae667248d16dR1287

msm/approvals.go
samliok: should we include an MSM test that integrates the approvals retriever in this pr?

yacovm: I think this should be made once we have the production code that connects the MSM with the approval retriever.

msm/approvals.go
samliok: sg, can we create an issue or a todo so we remember this?

msm/approvals_test.go
samliok: i was hoping to do it before to make it .5% easier to review the pr. but i think either is fine

msm/approvals.go
yacovm: https://github.com/ava-labs/Simplex/issues/395

msm/approvals.go
samliok: is it still not a problem though that we don't check the approval signature against the timestamp? we also don't pass in who sent the approval so someone could send an approval with a timestamp = MaxUint64 and say it was from another node. This would remove all other approvals from that node in `maybePruneOldApprovals`.

msm/approvals.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// We only store the last |as.validators| of approvals for each node,
+	// so we need to delete old approvals if we have more than |as.validators| approvals stored for this node.
+	as.maybePruneOldApprovals(approval)
+
+	return nil
+}
+
+func (as *ApprovalStore) maybePruneOldApprovals(approval *ValidatorSetApproval) {
+	for len(as.approvalsByNodes[approval.NodeID]) > len(as.validators) {
```
samliok: should we just do something like instead? i think the for loop is redundant(see comment)

```go

func (as *ApprovalStore) maybePruneOldApprovals(approval *ValidatorSetApproval) {
	approvals := as.approvalsByNodes[approval.NodeID]

	// We add at most one approval per call, so the node can exceed the cap by
	// at most one entry. When it does, evict the one with the oldest timestamp.
	if len(approvals) <= len(as.validators) {
		return
	}

	var oldest *ValidatorSetApproval
	for _, a := range approvals {
		if oldest == nil || a.Timestamp < oldest.Timestamp {
			oldest = a
		}
	}

	as.logger.Debug("Deleting old approval from node",
		zap.String("nodeID", fmt.Sprintf("%x", oldest.NodeID)),
		zap.Uint64("oldestApprovalPChainHeight", oldest.PChainHeight),
		zap.Uint64("oldestApprovalTimestamp", oldest.Timestamp))
	delete(approvals, oldest.PChainHeight)
	as.storedCount--
}

```

msm/approvals.go
yacovm: I think you have answered your own question. 

This only works if the dissemination of the approvals is point to point, over authenticated channels, and not via gossip.

yacovm: Also recall that:

```
type NextEpochApprovals struct {
	NodeIDs   []byte `canoto:"bytes,1"`
	Signature []byte `canoto:"bytes,2"`

	canotoData canotoData_NextEpochApprovals
}
```

This doesn't include timestamps, and we naturally cannot expect all signatures to match a single timestamp.

The timestamp is just there to give us liveness when a node crashes and restarts or picks a different P-chain height to sign over. We need some way of disposing of "old" in-memory values. 

We could have the timestamp be even computed locally inside of `HandleApproval` and not sent over the wire at all.

samliok: > We could have the timestamp be even computed locally inside of HandleApproval and not sent over the wire at all

might be a nice change, why send it if we don't need to.

msm/approvals.go
yacovm: yeah we could do that, that was probably too much of defensive programming

msm/approvals.go
yacovm: https://github.com/ava-labs/Simplex/pull/384/changes/da16638b8b0ee69b071de2f78c2fa5f5986d890b#diff-5e383ab850834861309c7e77b4123dbe73b6b4d3e2b96f96832869f5847b48fdR17-R19

msm/approvals.go
yacovm: https://github.com/ava-labs/Simplex/pull/384/changes/da16638b8b0ee69b071de2f78c2fa5f5986d890b#diff-5e383ab850834861309c7e77b4123dbe73b6b4d3e2b96f96832869f5847b48fdR104-R121

samliok (approved):

PR #385, opened 2026-05-18
Transition epochs if needed even if no transaction occurs

yacovm:
This commit makes the MSM monitor the P-chain height when building the first block of a new epoch, right after an epoch transition.

This is needed so that if we have no transaction traffic, but a need to transition to a new epoch, we won't get stuck in the old epoch.

### DIFF
--- a/msm/fake_node_test.go
+++ b/msm/fake_node_test.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
-	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -44,13 +43,7 @@ func TestFakeNodeEpochChangesDespiteEmptyMempool(t *testing.T) {
 		node.tryFinalizeNextBlock()
 	}
 
-	var wg sync.WaitGroup
-	wg.Add(2)
-
-	defer wg.Wait()
-
 	go func() {
-		defer wg.Done()
 		pChainHeight.Store(200)
 	}()
 
@@ -71,7 +64,6 @@ func TestFakeNodeEpochChangesDespiteEmptyMempool(t *testing.T) {
 
 		if node.isLastBlockSealing() {
 			go func() {
-				defer wg.Done()
 				pChainHeight.Store(300)
 			}()
 		}
--- a/msm/fake_node_test.go
+++ b/msm/fake_node_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


msm/fake_node_test.go
```diff
@@ -337,6 +400,14 @@ func (fn *fakeNode) isNextBlockTelock(nextIndex int) bool {
 	return fn.blocks[nextIndex].block.Metadata.SimplexEpochInfo.SealingBlockSeq > 0
 }
 
+func (fn *fakeNode) isLastBlockSealing() bool {
+	if len(fn.blocks) == 0 {
+		return false
+	}
+	last := fn.blocks[len(fn.blocks)-1]
+	return last.block.Metadata.SimplexEpochInfo.BlockValidationDescriptor != nil
```
samliok, line 396: unrelated but why do we have `BlockValidationDescriptor` again? can't we just grab the information encoded in `BlockValidatorDescriptor` from the pChain + pChainReferenceHeight?

msm/fake_node_test.go
```diff
@@ -15,6 +16,68 @@ import (
...
+				result: []ValidatorSetApproval{{NodeID: [20]byte{1}, PChainHeight: 200, Signature: []byte{1}, AuxInfoSeqDigest: [32]byte{}}},
+			}
+		} else {
+			node.sm.ApprovalsRetriever = &approvalsRetriever{
+				result: []ValidatorSetApproval{{NodeID: [20]byte{2}, PChainHeight: 200, Signature: []byte{2}, AuxInfoSeqDigest: [32]byte{}}},
+			}
+		}
+
+		if node.isLastBlockSealing() {
+			go func() {
```
samliok: why do we have this go func()

msm/fake_node_test.go
```diff
@@ -15,6 +16,68 @@ import (
...
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	defer wg.Wait()
+
+	go func() {
+		defer wg.Done()
+		pChainHeight.Store(200)
+	}()
```
samliok: what do these wait groups add? the tests pass if i remove all waitGroup stuff

msm/msm.go
```diff
@@ -295,7 +295,12 @@ func (sm *StateMachine) buildBlockNormalOp(ctx context.Context, parentBlock Stat
 		PrevVMBlockSeq:        computePrevVMBlockSeq(parentBlock, prevBlockSeq),
 	}
 
-	blockBuildingDecider := sm.createBlockBuildingDecider(parentBlock)
+	return sm.buildBlockOrTransitionEpoch(ctx, parentBlock, simplexMetadata, simplexBlacklist, newSimplexEpochInfo)
+}
+
+// buildBlockOrTransitionEpoch builds a block and decides whether to transition to a new epoch based on the P-chain height and validator set changes.
+func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentBlock StateMachineBlock, simplexMetadata, simplexBlacklist []byte, newSimplexEpochInfo SimplexEpochInfo) (*StateMachineBlock, error) {
```
samliok, line 302: this makes it seem like it is one _or_ the other. but they are not mutually exclusive, we can both build a block and decide to transition to the next epoch.

samliok (commented):
The bug is that if we are building the next block after an epoch transition we would call `BuildBlock` directly rather than the wrapper causing it to hang under no transactions?  a small pr description of the bug would also be a bit helpful.

msm/fake_node_test.go
yacovm: Because we need to move to the new epoch, so we need to progress the P-chain. 

The test has two epoch transitions.

msm/fake_node_test.go
yacovm: removed them

msm/msm.go
yacovm: Or isn't mutually exclusiveEither 🤷

msm/fake_node_test.go
yacovm: I'm not sure I understand. This just checks if the last block notarized is a sealing block. We don't care about the content, just the existence.

samliok: i wasn't asking in the context in the code. I was just asking why `BlockValidationDescriptor` is even in `SimplexEpochInfo`. Its presence just denotes that the block is a sealing block, so I was wondering if we could instead just query the pChain + pChainHeight to get this information instead of encoding the entire validator set into the block. 

But we probably want to encode the validator set so the nodes explicitly agree on exactly who the next nodes are. 

TLDR: ignore my comment

msm/msm.go
samliok: fair point

samliok (approved):

msm/fake_node_test.go
samliok: but why do we call it inside a new `goroutine`?

msm/fake_node_test.go
yacovm: Querying the P-chain height is not for free. You need to do a database query for that, so that involves disk I/O. 
We don't have an index of P-chain height --> validator set for L1s, but we just have diffs stored in the DB, so we need to traverse diffs between our current P-chain height and the target P-chain height and apply their inverse. This is linear in the "depth" of the P-chain height, which can be quite far away from the P-chain since we replicate the P-chain first and the chains later.

msm/fake_node_test.go
yacovm: yeah we don't need it, removed.

msm/fake_node_test.go
```diff
@@ -15,6 +15,57 @@ import (
...
+	}
+
+	pChainHeight.Store(200)
+
+	for node.Epoch() == 1 {
+		node.buildAndNotarizeBlock()
+		if node.canFinalize() {
+			node.tryFinalizeNextBlock()
+		}
+		if flipCoin() {
```
samliok, line 53: would it be better to make this test deterministic? what is the benefit for the non-determinism here?

yacovm: I merged before seeing your message 🙈 

So, flipping a coin here makes us collect approvals across several blocks. It tests the approval collection logic works across several blocks, as it is expected to happen in practice

PR #387, opened 2026-05-20
Improve NextState Comments

samliok:
I don't think we should use `prevBlockSimplexEpochInfo := sei` in this function. The caller is a previous simplex block, so I had to double take looking at this function because it read to me like we were doing `previous.previous`.  Plus we are getting the `NextState()` of `sei`, so we shouldn't be concerned that  `sei` is a `prevBlockSimplexEpochInfo`

### DIFF
--- a/msm/encoding.go
+++ b/msm/encoding.go
@@ -111,7 +111,7 @@ func (sei *SimplexEpochInfo) Equal(other *SimplexEpochInfo) bool {
 }
 
 // NextState returns the state used to build (or verify) the block that follows the one
-// described by sei.
+// described by the SimplexEpochInfo.
 func (sei *SimplexEpochInfo) NextState() state {
 	// No Simplex epoch has started yet: the next block is the first one built by Simplex.
 	if sei.EpochNumber == 0 {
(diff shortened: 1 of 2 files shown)


msm/encoding.go
```diff
@@ -110,27 +110,29 @@ func (sei *SimplexEpochInfo) Equal(other *SimplexEpochInfo) bool {
 	return true
 }
 
+// NextState returns the state used to build (or verify) the block that follows the one
+// described by sei.
```
yacovm: by this SimplexEpochInfo.

SEI is a blockchain protocol (https://www.sei.io/), let's not use it in our codebase 😉

msm/encoding.go
```diff
@@ -110,27 +110,29 @@ func (sei *SimplexEpochInfo) Equal(other *SimplexEpochInfo) bool {
...
 		return stateBuildBlockNormalOp
 	}
 
-	// If the previous block has a sealing block sequence, it's a Telock.
-	// If it has a block validation descriptor, it's a sealing block.
-	// Either way, the epoch has been sealed.
-	if prevBlockSimplexEpochInfo.SealingBlockSeq > 0 || prevBlockSimplexEpochInfo.BlockValidationDescriptor != nil {
+	// If NextPChainReferenceHeight > 0, then an epoch transition is in progress.
+	// An epoch is sealed if
+	//   - SealingBlockSeq > 0: sei describes a Telock.
```
yacovm: again, no `sei`.

yacovm:
looks good, modulo the sei

msm/encoding.go
samliok: updated

yacovm (approved):

PR #388, opened 2026-05-20
Add inlined documentation about state transitions

yacovm:


samliok (dismissed):
I realllly like the FSA defined above the `state` variable. It's clear, and easy to follow. 🎸

I'm concerned that larger comments above the `buildBlock...` functions will be hard to maintain / get out of sync. I dont think the logic will change much so I think its fine, but they are bulkier function comments that are probably hard to edit manually in the future.

Although claude code can be prompted to do so and it'll do a good job of keeping the formatting.

msm/encoding.go
```diff
@@ -44,18 +46,21 @@ type SimplexEpochInfo struct {
 	// of the sealing block of the previous epoch.
 	EpochNumber uint64 `canoto:"uint,2"`
 	// PrevSealingBlockHash is the hash of the sealing block of the previous epoch.
-	// It is empty for the first epoch, and the second epoch has the PrevSealingBlockHash set to be
-	// the hash of the first ever block built by the StateMachine.
+	// It is set to the hash of the zero block in the first epoch, and in subsequent epochs it is set to be
+	// the hash of the sealing block of the previous epoch.
+	// This is used to be able to quickly fetch and verify the sealing blocks without having to retrieve the interleaving blocks,
+	// which allows to bootstrap the BLS keys of the validator set for each epoch before fully syncing the interleaving blocks.
 	PrevSealingBlockHash [32]byte `canoto:"fixed bytes,3"`
```
samliok, line 51: should this be a simplex.Digest?

yacovm: yes but it's unrelated to this PR. I can change it in the PR where I will flatten the protocol metadata

msm/msm.go
```diff
@@ -569,6 +617,28 @@ func (sm *StateMachine) createBlockBuildingDecider(pChainReferenceHeight uint64)
...
+// the block at sequence n):
+//
+//	────────────────── Epoch 1 ────────────────────────────────────│─── Epoch s ────
+//	                                                               │
+//	Seq:     z          ...     s            s+1     ...    s+x    │ s+1  (Telocks get pruned) ...
+//	       ┌──────┐            ┌────────┐  ┌──────┐       ┌──────┐ │ ┌────────────┐
+//	       │ Zero │    ...     │Sealing │  │Telock│  ...  │Telock│ │ │first block │  ...
+//	       │ block│            │ block  │  │      │       │      │ │ │ of epoch s │
+//	       └──────┘            └────────┘  └──────┘       └──────┘ │ └────────────┘
+//	       EN  = 1             EN  = 1     EN  = 1        EN  = 1  │ EN  = s
```
samliok: i think we changed the epoch to not always be 1 for the zero block

yacovm: That's true, but it can still be 1 if we're building on top of genesis. 

Do you have an idea of what I should put here instead?

msm/msm.go
```diff
@@ -569,6 +617,28 @@ func (sm *StateMachine) createBlockBuildingDecider(pChainReferenceHeight uint64)
...
+//	       SBS = 0             SBS = 0     SBS = s        SBS = s  │ SBS = 0
+//	       PSH = 0             PSH = h(z)  PSH = 0        PSH = 0  │ PSH = 0
+//
+//	- EN  : copied within an epoch; on the first block of a new epoch, EN
+//	        equals the sequence number of the previous epoch's sealing block.
+//	- PSH : only set on a sealing block. In epoch 1 it points to the zero block;
```
samliok: this desc probably needs to be updated too?

msm/msm.go
samliok: i think we should note the first epoch is the height of the chain when starting. So if we are starting fresh from genesis it is 1, otherwise its the next sequence.

samliok: so maybe just a comment saying the zero block does not always have epoch 1

msm/msm.go
samliok: maybe instead of epoch 1, we just say in the first ever epoch

yacovm: https://github.com/ava-labs/Simplex/pull/388/commits/5583d08ebf3d4a5f9f2700d131384c5e80cb038c

msm/msm.go
yacovm: Added a comment here https://github.com/ava-labs/Simplex/pull/388/changes/5583d08ebf3d4a5f9f2700d131384c5e80cb038c#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R638-R640

samliok (approved):

PR #390, opened 2026-05-22
Remove Logger from NewFinalization

samliok:
i don't think we should be passing in a logger into `NewFinalization` just for one log.

### DIFF
--- a/epoch.go
+++ b/epoch.go
@@ -1178,7 +1178,11 @@ func (e *Epoch) maybeCollectFinalization(round *Round) error {
 }
 
 func (e *Epoch) assembleFinalization(round *Round, finalizationVotes []*FinalizeVote) error {
-	finalization, err := NewFinalization(e.Logger, e.signatureAggregator, finalizationVotes)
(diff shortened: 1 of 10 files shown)


epoch.go
```diff
@@ -1178,7 +1178,11 @@ func (e *Epoch) maybeCollectFinalization(round *Round) error {
 }
 
 func (e *Epoch) assembleFinalization(round *Round, finalizationVotes []*FinalizeVote) error {
-	finalization, err := NewFinalization(e.Logger, e.signatureAggregator, finalizationVotes)
+	for _, vote := range finalizationVotes {
+		e.Logger.Debug("Collected a finalize vote from node", zap.Stringer("NodeID", vote.Signature.Signer), zap.Uint64("round", vote.Finalization.Round))
```
yacovm: can we add the sequence number too?

yacovm (approved):
yeah it was bothering me too, because in tests we were using the testutil constructor which was invoking this method and it was giving a misleading logging entry.

PR #393, opened 2026-06-02
Adding Support For Non-Validators

samliok:
# Non-Validator Node

Adds a `nonvalidator` package implementing a node that **follows** a Simplex chain,
verifying and indexing finalized blocks, without participating in consensus (it never
votes, proposes, or signs). It tracks the chain by collecting blocks + finalizations from
validators and replicating any sequences it's missing.

## `NonValidator` (`non_validator.go`)
Similarly to the epoch we have a `HandleMessage` function. However we only listen to these message types:

- **Block messages** (`handleBlock`) — accepts a block only if it's from the round's
  leader, in a known epoch, and within the sequence window. Blocks are buffered in
  `incompleteSequences` until the matching finalization arrives.
- **Finalizations** (`handleFinalization`) — verifies the QC against the epoch's validator
  set. Matches against any buffered block; finalizations too far ahead or for unknown
  epochs are forwarded to the replicator / trigger a sealing-block request.
- **Replication responses** (`handleReplicationResponse`) — validates each fetched quorum
  round, stores it in the replication state, then schedules the next in-order finalized
  block for verification and indexing.

Once a block **and** a verified finalization are held for a sequence, a verification task is
scheduled (`scheduleNewFinalizedBlockTask`) through a `BlockDependencyManager`, which
enforces that sequences are verified/indexed in order. 

## Epoch validation (`epochs.go`, `epoch_replicator.go`)

A non-validator must know each epoch's validator set to verify finalizations. Validator
sets are carried in **sealing blocks** (new `SealingBlockInfo`, see below) which form a
backward-linked hash chain across epochs.

- `epochs` map holds only epochs whose validator set has been validated. Bootstrapped from
  storage in `newEpochs` (handles genesis / pre-Simplex `Epoch 0`).
- A new epoch is validated when (`maybeValidateNextEpoch`):
  1. we receive/verify the sealing block normally, **or**
  2. `canValidate` — we already trust the *next* epoch and this block matches its
     `prevSealingBlockHash` (validate backward down the chain), **or**
  3. `epochReplicator.collectedQuorumRound` — we've collected `F+1` matching sealing-block
     responses from distinct nodes for an epoch we can't otherwise verify


nonvalidator/non_validator.go
```diff
@@ -0,0 +1,571 @@
...
+		return nil
+	}
+
+	if n.haltedError != nil {
+		return n.haltedError
+	}
+
+	switch {
+	case msg.BlockDigestRequest != nil:
+		// TODO: it seems reasonable for our non-validator to be able to process these messages and send out responses.
```
samliok: for another pr

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,571 @@
...
+		return fmt.Errorf("could not verify quorum round QC: %w", err)
+	}
+
+	// This block could be a sealing block, validate the next epoch if so.
+	n.maybeValidateNextEpoch(block)
+	n.sequenceReplicator.StoreQuorumRound(qr)
+	return nil
+}
+
+// TODO: add a re-broadcast timeout task until we have validated an epoch.
```
samliok, line 538: for another pr

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,571 @@
...
+	n.Comm.Broadcast(&simplex.Message{
+		ReplicationRequest: request,
+	})
+}
+
+// sendRequest sends a simplex.BlockDigestRequest for a given sequence to a node.
+func (n *NonValidator) sendRequest(seq uint64, to simplex.NodeID) {
+	digestRequest := simplex.BlockDigestRequest{
+		Seq:    seq,
+		Digest: simplex.Digest{}, // TODO: In the epoch code, update how we process digests to not drop the request given an empty digest.
```
samliok: for another pr

common/api.go
```diff
@@ -90,6 +90,22 @@ type Block interface {
 
 	// Verify verifies the block by speculatively executing it on top of its ancestor.
 	Verify(ctx context.Context) (VerifiedBlock, error)
+
+	// non nil only for sealing blocks & first ever simplex block
+	SealingBlockInfo() *SealingBlockInfo
+}
+
+type SealingBlockInfo struct {
```
yacovm, line 102: I think we don't need the `Epoch` here, and we can just do with the last two. 

The epoch can be inferred from the simplex protocol metadata by looking at the sequence.

Essentially, if this is a sealing block, it has a sequence and a validator set and a prev sealing block hash. 

The new epoch number is the seq of the protocol metadata.

simplex/block_scheduler.go
yacovm, line 13: Since this file only references `common`, and we need to use this file in the non-validator, it probably means it makes sense to move this file to `common`.

simplex/epoch.go
```diff
@@ -3436,12 +3436,16 @@ func LeaderForRound(nodes []common.NodeID, r uint64) common.NodeID {
...
 
 func Quorum(n int) int {
-	f := (n - 1) / 3
+	f := F(n)
 	// Obtained from the equation:
 	// Quorum * 2 = N + F + 1
 	return (n+f)/2 + 1
 }
 
+func F(n int) int {
```
yacovm: we can just move this to common

simplex/epoch.go
```diff
@@ -3436,12 +3436,16 @@ func LeaderForRound(nodes []common.NodeID, r uint64) common.NodeID {
 }
 
 func Quorum(n int) int {
```
yacovm: let's move this to common as well.

simplex/util.go
```diff
@@ -381,7 +381,7 @@ func (t *walRound) String() string {
 	return fmt.Sprintf("walRound{round: %d, hasEmptyNotarization: %t, hasEmptyVote: %t, hasNotarization: %t, hasFinalization: %t, hasBlock: %t}", t.round, hasEmptyNotarization, hasEmptyVote, hasNotarization, hasFinalization, hasBlock)
 }
 
-func newRandomSource() (*rand.Rand, error) {
+func NewRandomSource() (*rand.Rand, error) {
```
yacovm, line 371: shall we move this to common?

common/msg.go
```diff
@@ -320,17 +320,13 @@ func (q *QuorumRound) VerifyQCConsistentWithBlock() error {
 
 // String returns a string representation of the QuorumRound.
 // It is meant as a debugging aid for logs.
-func (q *QuorumRound) String() string {
-	if q != nil {
```
yacovm, line 324: why are we dropping this check?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+	n.cancelCtx()
+	n.sequenceReplicator.Close()
+	n.verifier.Close()
+}
+
+func (n *NonValidator) AdvanceTime(t time.Time) {
+	n.sequenceReplicator.AdvanceTime(t)
+}
+
+// this function should be ran under a lock?
```
yacovm: > // this function should be ran under a lock?

Yes, because we have a scheduler and therefore we have multiple goroutines that touch the state of `n *NonValidator)`.

nonvalidator/epochs.go
```diff
@@ -0,0 +1,119 @@
...
+
+	digest := block.BlockHeader().Digest
+	for _, md := range e {
+		if bytes.Equal(md.prevSealingBlockHash[:], digest[:]) {
+			// We have validated the next epoch, and the next epoch has a backward pointer to this one
+			return true
+		}
+	}
+	return false
+}
```
yacovm, line 129: this file is ~ 100 lines but it's used only for `non_validator.go`, do we really need a separate file for it?

nonvalidator/epochs.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	epochs := make(map[uint64]*epochMetadata)
+
+	// A zero Epoch means this is before the first ever Simplex block(ex. Genesis or Last Snowman Block)
+	if lastBlock.BlockHeader().Epoch == 0 {
+		return epochs, nil
+	}
+
+	sealingBlock := lastBlock
```
yacovm: > 	sealingBlock := lastBlock

I don't think we can make this assignment. It's not true that the last block is a sealing block. 

Below you override properly, but the assignment is premature.

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+		randomSource, err = simplex.NewRandomSource()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	scheduler := simplex.NewScheduler(config.Logger, simplex.DefaultProcessingBlocks)
+
+	lock := &sync.Mutex{}
```
yacovm, line 104: you can define the lock as a `sync.Mutex` and not a pointer and then you don't need to initialize it.

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+		return nil
+	}
+
+	// Duplicate block, or finalization not yet received.
+	if incomplete.block != nil || incomplete.finalization == nil {
+		return nil
+	}
+
+	if !bytes.Equal(incomplete.finalization.Finalization.Digest[:], bh.Digest[:]) {
+		n.Logger.Info(
```
yacovm: I think we can put this in Debug.

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+			"Received a block from the leader of a round whose digest mismatches the finalization",
```
yacovm, line 213: shouldn't we remove the block in such a case?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+	n.Comm.Broadcast(&common.Message{
+		ReplicationRequest: request,
+	})
+}
+
+// sendRequest sends a common.BlockDigestRequest for a given sequence to a node.
+func (n *NonValidator) sendRequest(seq uint64, to common.NodeID) {
+	digestRequest := common.BlockDigestRequest{
+		Seq:    seq,
+		Digest: common.Digest{}, // TODO: In the epoch code, update how we process digests to not drop the request given an empty digest.
```
yacovm: why not just send a replication request? why use a digest request if we don't know the digest?

nonvalidator/epochs.go
```diff
@@ -0,0 +1,119 @@
...
+	"errors"
+
+	"github.com/ava-labs/simplex/common"
+)
+
+var errNoGenesis = errors.New("No Genesis Found")
+
+type epochMetadata struct {
+	nodes                common.Nodes
+	nodeLookup           map[string]struct{}
```
yacovm: `nodeLookup` is a poor name for what it does. It's used in `VerifyQC` so let's call it `eligibleSigners` as it is called there?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+	if incomplete.finalization != nil {
+		// sanity check: should never happen.
+		if !bytes.Equal(incomplete.finalization.Finalization.Bytes(), finalization.Finalization.Bytes()) {
+			n.Logger.Warn(
+				"Mismatching finalizations",
+				zap.Uint64("Incoming Sequence", finalization.Finalization.Seq),
+				zap.Uint64("Stored sequence", incomplete.finalization.Finalization.Seq),
+			)
+			n.haltedError = fmt.Errorf("Conflicting finalizations")
+			return fmt.Errorf("Conflicting finalizations")
```
yacovm: errors should be lowercase prefixed

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+	// No block received yet for this sequence.
+	if incomplete.block == nil {
+		n.sequenceReplicator.ReceivedFutureFinalization(finalization, n.nextSeqToCommit())
+		return nil
+	}
+
+	digest := incomplete.block.BlockHeader().Digest
+	if !bytes.Equal(bh.Digest[:], digest[:]) {
+		n.Logger.Info(
+			"Received a block from the leader of a round whose digest mismatches the finalization",
```
yacovm, line 387: shouldn't we get rid of this block?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+
+	// No block received yet for this sequence.
+	if incomplete.block == nil {
+		n.sequenceReplicator.ReceivedFutureFinalization(finalization, n.nextSeqToCommit())
+		return nil
+	}
+
+	digest := incomplete.block.BlockHeader().Digest
+	if !bytes.Equal(bh.Digest[:], digest[:]) {
+		n.Logger.Info(
```
yacovm: can we log it as Debug?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+
+	// Don't store finalization in memory if it's too far ahead
+	if bh.Seq > n.MaxSequenceWindow+n.Storage.NumBlocks() {
+		n.Logger.Debug("Received a finalization from a sequence too far ahead", zap.Uint64("Num Blocks", n.Storage.NumBlocks()), zap.Uint64("Block Sequence", bh.Seq), zap.Stringer("From", from))
+		n.sequenceReplicator.ReceivedFutureFinalization(finalization, n.nextSeqToCommit())
+		return nil
+	}
+
+	incomplete, ok := n.incompleteSequences[bh.Seq]
+	if !ok {
```
yacovm, line 350: can we do the following to consolidate `n.sequenceReplicator.ReceivedFutureFinalization(finalization, n.nextSeqToCommit())` in all below cases? 

```
	if ! ok || (incomplete.finalization == nil || incomplete.block == nil) {
		n.sequenceReplicator.ReceivedFutureFinalization(finalization, n.nextSeqToCommit())
	}
```

yacovm (commented):
I am not done, but spilling what I have so far

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	e.logger.Debug("Collected a quorum round with a sealing block", zap.Stringer("QR", qr), zap.Stringer("From", from))
+
+	sealingInfo := qr.Block.SealingBlockInfo()
+	threshold := simplex.F(len(e.latestValidatorSetRetriever.Nodes())) + 1
+	epochResponses, ok := e.sealingBlockResponses[sealingInfo.Epoch]
+	digest := qr.Block.BlockHeader().Digest
+	if !ok {
+		digests := make(map[string]common.Digest)
+		digests[from.String()] = digest
```
yacovm: I don't think that `from.String()` is the same as `string(from)` below. 

This may lead to double counting, no?

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
...
+		return false
+	}
+	e.logger.Debug("Collected a quorum round with a sealing block", zap.Stringer("QR", qr), zap.Stringer("From", from))
+
+	sealingInfo := qr.Block.SealingBlockInfo()
+	threshold := simplex.F(len(e.latestValidatorSetRetriever.Nodes())) + 1
+	epochResponses, ok := e.sealingBlockResponses[sealingInfo.Epoch]
+	digest := qr.Block.BlockHeader().Digest
+	if !ok {
+		digests := make(map[string]common.Digest)
```
yacovm: why do we call this digests? this should be called `epochResponses` to be consistent. 

Also, can we make the flow more streamlined? 


```
	if !ok {
		epochResponses = make(map[string]common.Digest)
		epochResponses[string(from)] = digest
		e.sealingBlockResponses[sealingInfo.Epoch] = epochResponses
	}

	epochResponses[string(from)] = digest
```

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
...
+	} else {
+		epochResponses[string(from)] = digest
+	}
+
+	e.digests[digest.String()] = qr
+
+	// check if we have a threshold of responses
+	counts := make(map[string]uint64)
+	for _, digest := range epochResponses {
+		sDigest := digest.String()
```
yacovm: using `.String()` to count is not a good idea:

```
func (node NodeID) String() string {
	var nodePrint [8]byte
	copy(nodePrint[:], node)
	return hex.EncodeToString(nodePrint[:])
}
```

`nodePrint` only has 64 bits, so an adversarial node can brute force $2^{32}$ keys to find a node ID that collides with another existing node, and then to vote in its stead. 

Even without that, a probability of $\frac{1}{2^{64}}$ nodes colliding naively is not great, we can do better.

yacovm: oh actually I now see that this `.String()` is 10 bytes, not 8, so similar argument but with $40$ and $80$ respectively. 

This argument also holds for `from.String()` in line 55 with 8 bytes though.

nonvalidator/non_validator.go
yacovm: If that's for another PR can we open an issue and remove it in the meantime? 

And I don't understand how we are expected to receive a digest request when we don't even have a way to express a non-validator in the comm API?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+		n.Logger.Info(
+			"Received a block from the leader of a round whose digest mismatches the finalization",
+			zap.Stringer("Finalization Digest", incomplete.finalization.Finalization.Digest),
+			zap.Stringer("Block digest", bh.Digest),
+			zap.Stringer("From", from),
+		)
+		return nil
+	}
+
+	// add test that ensure this is here. otherwise i think an adversarial node can have us schedule many tasks
```
yacovm: what do you mean by "this is here" ?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+		start := time.Now()
+		defer func() {
+			elapsed := time.Since(start)
+			n.Logger.Debug("Block verification ended", zap.Uint64("sequence", md.Seq), zap.Duration("elapsed", elapsed))
+		}()
+
+		verifiedBlock, err := block.Verify(n.ctx)
+		// We have failed verifying a finalized block
+		if err != nil {
+			n.lock.Lock()
```
yacovm: we can just move `n.lock.Lock()` and `n.lock.Unlock()` to right after the `verifiedBlock, err := block.Verify(n.ctx)` and this saves us to write it twice here and in 267.

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+			n.Logger.Info("Failed verifying a block that has a finalization", zap.Uint64("Block Seq", md.Seq), zap.Stringer("Block Digest", md.Digest), zap.Error(err))
+
+			n.sequenceReplicator.ResendFinalizationRequest(md.Seq, finalization.QC.Signers())
+			return md.Digest
+		}
+
+		n.lock.Lock()
+		defer n.lock.Unlock()
+
+		// defensive check in case we scheduled multiple finalized block verification tasks
```
yacovm, line 248: can we move it to right after `verifiedBlock, err := block.Verify(n.ctx)` but inside the lock, so we only do it once? 


```
verifiedBlock, err := block.Verify(n.ctx)

		n.lock.Lock()
		defer n.lock.Unlock()

		// defensive check in case we scheduled multiple finalized block verification tasks
		if n.Storage.NumBlocks() != md.Seq {
			n.Logger.Debug("Received finalized block that is not the next sequence to commit",
				zap.Uint64("seq", md.Seq), zap.Uint64("height", n.Storage.NumBlocks()))
			return md.Digest
		}
		
		// We have failed verifying a finalized block
		if err != nil {
			n.Logger.Info("Failed verifying a block that has a finalization", zap.Uint64("Block Seq", md.Seq), zap.Stringer("Block Digest", md.Digest), zap.Error(err))

			n.sequenceReplicator.ResendFinalizationRequest(md.Seq, finalization.QC.Signers())
			return md.Digest
		}

		if err := n.Storage.Index(n.ctx, verifiedBlock, *finalization); err != nil {
			n.haltedError = err
			n.Logger.Info("Failed indexing a block and finalization", zap.Uint64("Block Seq", md.Seq), zap.Stringer("Block Digest", md.Digest), zap.Error(err))
			return md.Digest
		}

		n.Logger.Info("Verified and Indexed Block", zap.Uint64("Block Seq", md.Seq), zap.Stringer("Block Digest", md.Digest))

		n.removeOldSequencesAndEpochs(md.Seq, md.Epoch)

		// in case we need to queue up any more tasks
		if err := n.processReplicationState(); err != nil {
			n.haltedError = err
			n.Logger.Info("Failed calling process replication state", zap.Error(err))
			return md.Digest
		}

		return md.Digest
```

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+	if err := n.processQuorumRound(resp.LatestSeq, from); err != nil {
+		n.Logger.Debug("Failed processing latest seq", zap.Stringer("QR", resp.LatestSeq), zap.Error(err))
+	}
+
+	return n.processReplicationState()
+}
+
+func (n *NonValidator) processReplicationState() error {
+	nextSeqToCommit := n.nextSeqToCommit()
+	n.sequenceReplicator.MaybeAdvanceState(nextSeqToCommit, 0, 0)
```
yacovm, line 436: I get why we're putting 0 here but would it be really that bad to put the real values?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+		}()
+
+		verifiedBlock, err := block.Verify(n.ctx)
+		// We have failed verifying a finalized block
+		if err != nil {
+			n.lock.Lock()
+			defer n.lock.Unlock()
+
+			// defensive check in case we scheduled multiple finalized block verification tasks
+			if n.Storage.NumBlocks() != md.Seq {
```
yacovm: can we use `nextSeqToCommit` instead?

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+
+			n.sequenceReplicator.ResendFinalizationRequest(md.Seq, finalization.QC.Signers())
+			return md.Digest
+		}
+
+		n.lock.Lock()
+		defer n.lock.Unlock()
+
+		// defensive check in case we scheduled multiple finalized block verification tasks
+		if n.Storage.NumBlocks() != md.Seq {
```
yacovm: can we use `nextSeqToCommit` instead?

nonvalidator/epochs.go
```diff
@@ -0,0 +1,119 @@
...
+	epoch                uint64
+	signatureAggregator  common.SignatureAggregator
+	prevSealingBlockHash common.Digest
+}
+
+type epochs map[uint64]*epochMetadata
+
+func newEpochMetadata(sealingMetadata *common.SealingBlockInfo, sigCreator common.SignatureAggregatorCreator) *epochMetadata {
+	if sealingMetadata == nil {
```
yacovm, line 30: what's the semantic meaning of this?

nonvalidator/epochs.go
```diff
@@ -0,0 +1,119 @@
...
+
+	sealingBlock := lastBlock
+	if sealingBlock.SealingBlockInfo() == nil {
+		sealingBlock, _, err = storage.Retrieve(lastBlock.BlockHeader().Epoch)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	lastAcceptedEpoch := newEpochMetadata(sealingBlock.SealingBlockInfo(), sigAggCreator)
```
yacovm: what if `sealingBlock.SealingBlockInfo()` is nil? Shouldn't we return an error in such a case?

nonvalidator/epochs.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	lastBlock, _, err := storage.Retrieve(lastBlockHeight - 1)
+	if err != nil {
+		return nil, err
+	}
+
+	epochs := make(map[uint64]*epochMetadata)
+
+	// A zero Epoch means this is before the first ever Simplex block(ex. Genesis or Last Snowman Block)
```
yacovm, line 64: > // A zero Epoch means this is before the first ever Simplex block(ex. Genesis or Last Snowman Block)

But if it's the last snowman block, the first ever epoch number isn't numbered 0

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,572 @@
...
+	n.maybeValidateNextEpoch(block)
+	n.sequenceReplicator.StoreQuorumRound(qr)
+	return nil
+}
+
+// TODO: add a re-broadcast timeout task until we have validated an epoch.
+func (n *NonValidator) broadcastLatestEpoch() {
+	highestEpoch, _ := n.epochs.highestEpoch()
+
+	// Sending a LatestFinalizedSeq of 0 gets ignored by validators.
```
yacovm, line 542: do we really need to do this if it's 0? Maybe just skip it?

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// TODO: garbage collect map
+type epochReplicator struct {
+	logger common.Logger
+
+	// receivedSealingBlocks stores sealing blocks that have been received by our non-validator.
+	// It maps the epoch they are creating, to the NodeIds that have sent them to us(and which digest they sent).
+	// Once we have collected f+1 messages for a finalization, the sealing block for that epoch is validated.
+	sealingBlockResponses map[uint64]map[string]common.Digest
```
yacovm: when is this garbage collected?

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// digests stores the metadata associated with a sealing block.
+	digests map[string]*common.QuorumRound
```
yacovm: when is this garbage collected?

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	e.digests[digest.String()] = qr
+
+	// check if we have a threshold of responses
+	counts := make(map[string]uint64)
+	for _, digest := range epochResponses {
+		sDigest := digest.String()
+		_, ok := counts[sDigest]
+		if !ok {
```
yacovm: Isn't this simpler?

```
		count := counts[sDigest]
		count++
		counts[sDigest] = count
```

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
...
+// TODO: garbage collect map
+type epochReplicator struct {
+	logger common.Logger
+
+	// receivedSealingBlocks stores sealing blocks that have been received by our non-validator.
+	// It maps the epoch they are creating, to the NodeIds that have sent them to us(and which digest they sent).
+	// Once we have collected f+1 messages for a finalization, the sealing block for that epoch is validated.
+	sealingBlockResponses map[uint64]map[string]common.Digest
+
+	// digests stores the metadata associated with a sealing block.
```
yacovm: where do we use the value from the map? we only assign to it.

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,81 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
```
yacovm: This is called an epoch replicator but it doesn't really replicate.

I also feel like the amount of functionality we have here is minimal. 

I think it makes sense to just collapse this into `epochs.go`. 

Thoughts?

nonvalidator/non_validator_test.go
```diff
@@ -0,0 +1,643 @@
...
+// the "qc does not verify" code path without mutating message contents.
+type errQC struct {
+	common.QuorumCertificate
+}
+
+func (errQC) Verify(_ []byte, _ common.Nodes) error {
+	return errors.New("qc verification failed")
+}
+
+type TestConfig struct {
```
yacovm: I think this is unused

nonvalidator/epoch_replicator.go
yacovm: We should add a test for this

nonvalidator/non_validator_test.go
```diff
@@ -0,0 +1,643 @@
...
+	require.Never(t,
+		func() bool {
+			return nv.Storage.NumBlocks() > 3
+		},
+		2*time.Second,
+		50*time.Millisecond,
+	)
+}
+
+// TestHandleMessages_DuplicateBlock tests that when a duplicate block is received, the block is verify & indexed only once
```
yacovm: s/the block is verify & indexed only once/the block is verified & indexed only once

nonvalidator/chain_helpers_test.go
```diff
@@ -0,0 +1,223 @@
...
+	Epoch: 0,
+}, common.Blacklist{})
+
+type messageInfo struct {
+	msg  *common.Message
+	from common.NodeID
+}
+
+// send is a helper that calls HandleMessage on `nv` with `m`.
+func (m *messageInfo) send(nv *NonValidator) error {
```
yacovm: I don't see how this helps us... can't we just call `nv.HandleMessage(m.msg)` ?

nonvalidator/comm_test.go
```diff
@@ -0,0 +1,137 @@
...
+	// simulated responder.
+	ID common.NodeID
+
+	// responses is the queue of synthesized responses. Tests will pop messages
+	// and feed entries into NonValidator.HandleMessage.
+	responsesLock sync.Mutex
+	responses     []*messageInfo
+}
+
+func newTestResponder(t *testing.T, myNodeID common.NodeID, tc *testChain) *nonValidatorResponderComm {
```
yacovm, line 36: so, why do we need a test responder? Can't we just initialize several instances of `simplex.Epoch`? One for every epoch, all primed with a full blockchain up to their epoch, and this will prove that this works properly with real epoch instances.

simplex/block_scheduler.go
samliok: https://github.com/ava-labs/Simplex/pull/414

simplex/epoch.go
samliok: https://github.com/ava-labs/Simplex/pull/414

simplex/util.go
samliok: https://github.com/ava-labs/Simplex/pull/414

nonvalidator/non_validator.go
samliok: are you saying if our latest known epoch is 0 we shouldn't send this message? how else would we bootstrap from genesis/last snowman block:

nonvalidator/comm_test.go
samliok: our epoch isn't currently set up to handle receiving messages from nodes that are non-validators. We reject any message that is not from an eligible signer. I think it makes sense to create tests that use the epoch struct, but I think we should do it in a separate pr(to avoid bloating this one with changes to `epoch.go`) or we should wait to do it until the epoch coordinator changes are in. This way we could just route the messages to the epoch coordinator and it should be able to figure out a response from there

nonvalidator/non_validator.go
yacovm: well I thought that we might hear a new finalization and then replicate without it, but yeah let's keep it 👍

nonvalidator/comm_test.go
yacovm: > our epoch isn't currently set up to handle receiving messages from nodes that are non-validators. 

Well, this can be easily controlled by the epoch's config. 

> I think it makes sense to create tests that use the epoch struct, but I think we should do it in a separate pr(to avoid bloating this one with changes to epoch.go) or we should wait to do it until the epoch coordinator changes are in. 

But if we have real epochs in the test, wouldn't we be able to get rid of this artificial test responders and make this PR smaller?

common/msg.go
samliok: oops, i changed it for debugging forgot to change back

nonvalidator/epochs.go
samliok: yea i did this to avoid having and `if` + `else` clause but i think its more understandable if its set more explicitly
```
	var sealingBlock common.VerifiedBlock
	if lastBlock.SealingBlockInfo() == nil {
		sealingBlock, _, err = storage.Retrieve(lastBlock.BlockHeader().Epoch)
		if err != nil {
			return nil, err
		}
	} else {
		sealingBlock = lastBlock
	}
```

nonvalidator/non_validator.go
samliok: but i need to pass it into the `replicator` struct below. is there a preference for storing `lock` as a value or pointer in the `NonValidator` struct?

nonvalidator/non_validator.go
samliok: wdym by remove? if we are here this means we have never added a block to `incompleteSequences` and we return early so it's never added

nonvalidator/non_validator.go
samliok: yea good point

nonvalidator/non_validator.go
samliok: created this issue. https://github.com/ava-labs/Simplex/issues/417

> And I don't understand how we are expected to receive a digest request when we don't even have a way to express a non-validator in the comm API?

but this is the same problem as non-validators receiving `Block` and `Finalization` messages. We will probably need to update the comm interface or implementation regardless and when we do maybe we include an option to include non-validators in this request. Same can be said for replication requests.

nonvalidator/non_validator.go
samliok: i dont think it would matter because we only accept block messages from leaders of the round, so if we added a block message before and got a different finalization that means the leader sent us a wrong block. Deleting it would mean we expect the leader to send us another block, which we shouldn't expect because they previously acted adversarially. 

Instead, we should just let the replication path handle it properly.

nonvalidator/non_validator.go
samliok: this is a stale comment, i added the check below(in scheduleNewFinalizedBlockTask) to reduce over scheduling of tasks in the case that an adversary spams us with the same block message.
 
`if n.verifier.IsSequenceScheduled(bh.Seq) {`

nonvalidator/non_validator.go
samliok: we would need to get the round somehow. We could fetch the last block and get the most recent round from the metadata, but I think this storage lookup is unnecessary especially since we don't really care about rounds

nonvalidator/epochs.go
samliok: It was a defensive check in case we ever passed in a `nil` sealingMetadata, but we never do so I think its unnecessary.

nonvalidator/epochs.go
samliok: added this check for when we fetch the block + test

nonvalidator/epochs.go
samliok: i was saying that if we retrieve a block header and the epoch is numbered 0, that means we dont have the first ever epoch yet.

In the case of genesis, the block header returns an epoch 0.
In the case of last snowman block, I'm not sure. Currently our storage implementation in avalanchego will panic if we try and retrieve it. In my opinion we should be able to fetch the protocol metadata, but the epoch number should still be 0(equivalent to an uninitialized blockheader).

nonvalidator/comm_test.go
samliok: https://github.com/ava-labs/Simplex/issues/418

created an issue, will do in a separate pr as discussed offline

nonvalidator/epoch_replicator.go
samliok: i dont want to combine the structs, but i can move them to the same file. You have a comment about moving epochs.go into non_validator and I want to avoid making non-validator on big file. Mainly because I think it makes the abstractions easier to see and the tests easier to walk through(since they are in their own files).

nonvalidator/epoch_replicator.go
samliok: from.String() is not equal to string(from). I changed it so we use string(from) for all, does it make sense adding this test? 

```go
func TestNodeStringEquality(t *testing.T) {
	n := NodeID{1, 2, 3, 4, 45, 5, 56, 6, 6, 4, 123, 123, 90, 2, 1, 1, 1, 1, 3, 4, 5, 65, 1, 1}
	require.NotEqual(t, n.String(), string(n))
}
```

nonvalidator/non_validator.go
samliok: done
https://github.com/ava-labs/Simplex/pull/393/commits/676851b58ec0887614ca35af281301355491435d

common/api.go
samliok: updated https://github.com/ava-labs/Simplex/pull/393/commits/ef7a6d977124873857558a785d133bdb1fc129c6

nonvalidator/non_validator_test.go
samliok: not sure i understand this comment

yacovm: verify --> verified

nonvalidator/epoch_replicator.go
yacovm: That's not what I meant by saying we should add a test. I meant that we should add a test where a node votes twice (sends us two quorum rounds for the same sequence) and its votes are deduplicated.

common/api.go
```diff
@@ -92,6 +92,20 @@ type Block interface {
 
 	// Verify verifies the block by speculatively executing it on top of its ancestor.
 	Verify(ctx context.Context) (VerifiedBlock, error)
+
+	// non nil only for sealing blocks & first ever simplex block
```
yacovm, line 96: // SealingBlockInfo returns a non-nil value for a block that is not a sealing block and that is not the first ever simplex block.

common/api.go
```diff
@@ -92,6 +92,20 @@ type Block interface {
...
+	SealingBlockInfo() *SealingBlockInfo
+}
+
+type SealingBlockInfo struct {
```
yacovm, line 102: Please let's have a godoc comment. 

// SealingBlockInfo defines information that is derived from a sealing block, namely the validator set and the hash of the previous sealing block if applicable.

nonvalidator/chain_helpers_test.go
```diff
@@ -0,0 +1,215 @@
+package nonvalidator
```
yacovm, line 4: license header missing

nonvalidator/comm_test.go
```diff
@@ -0,0 +1,137 @@
+package nonvalidator
```
yacovm, line 4: license header missing on multiple files

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,77 @@
...
+
+package nonvalidator
+
+import (
+	"github.com/ava-labs/simplex/common"
+	"go.uber.org/zap"
+)
+
```
yacovm: // latestValidatorSetRetriever is an allows the epoch replicator to get the latest validator set. 
// This is used to calculate the threshold of votes needed to validate an epoch.

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,77 @@
...
+type latestValidatorSetRetriever interface {
+	Nodes() common.Nodes
+}
+
+type epochReplicator struct {
+	logger common.Logger
+
+	// receivedSealingBlocks stores sealing blocks that have been received by our non-validator.
```
yacovm: wrong name?

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,77 @@
...
+
+func newEpochReplicator(logger common.Logger, validatorSetRetriever latestValidatorSetRetriever) *epochReplicator {
+	return &epochReplicator{
+		sealingBlockResponses:       make(map[uint64]map[string]common.Digest),
+		logger:                      logger,
+		latestValidatorSetRetriever: validatorSetRetriever,
+	}
+}
+
+// collectedQuorumRound notes that an quorum round for an unknown epoch was received. If that quorum round
```
yacovm: shouldn't it be "a quorum round"?

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,77 @@
...
+	e.logger.Debug("Collected a quorum round with a sealing block", zap.Stringer("QR", qr), zap.Stringer("From", from))
+
+	sealingInfo := qr.Block.SealingBlockInfo()
+	threshold := common.F(len(e.latestValidatorSetRetriever.Nodes())) + 1
+	newEpoch := qr.Block.BlockHeader().Seq
+	epochResponses, ok := e.sealingBlockResponses[newEpoch]
+	digest := qr.Block.BlockHeader().Digest
+	if !ok {
+		epochResponses = make(map[string]common.Digest)
+		epochResponses[string(from)] = digest
```
yacovm: this line is redundant

nonvalidator/epoch_replicator.go
yacovm: can we call it something else? `epochDigestCounter` ?

nonvalidator/epoch_replicator.go
```diff
@@ -0,0 +1,77 @@
...
+	return &epochReplicator{
+		sealingBlockResponses:       make(map[uint64]map[string]common.Digest),
+		logger:                      logger,
+		latestValidatorSetRetriever: validatorSetRetriever,
+	}
+}
+
+// collectedQuorumRound notes that an quorum round for an unknown epoch was received. If that quorum round
+// was a sealing block we mark it and return if a threshold of responses was reached. otherwise, return false.
+func (e *epochReplicator) collectedQuorumRound(qr *common.QuorumRound, from common.NodeID) bool {
```
yacovm: we don't really need the entire QR, right? 

Maybe make it accept the sealing block info and the block header?

nonvalidator/epochs.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"bytes"
+	"errors"
+
+	"github.com/ava-labs/simplex/common"
+)
+
+var (
+	errNoGenesis          = errors.New("No Genesis Found")
```
yacovm: errors start with lowercase

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,562 @@
...
+	})
+}
+
+// sendRequest sends a common.BlockDigestRequest for a given sequence to a node.
+func (n *NonValidator) sendRequest(seq uint64, to common.NodeID) {
+	request := common.ReplicationRequest{
+		Seqs: []uint64{seq},
+	}
+
+	n.Logger.Debug("Broadcasting sealing block request", zap.Uint64("Requesting Seq", seq))
```
yacovm: We are sending this to a single node, so this isn't a broadcast, it's a unicast.

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,562 @@
...
+	block := qr.Block
+	finalization := qr.Finalization
+
+	// Non validators only process quorum rounds with finalizations
+	if finalization == nil {
+		return nil
+	}
+
+	if n.isAccepted(block.BlockHeader().Seq) {
+		return fmt.Errorf("processing quorum round for a block we already indexed")
```
yacovm, line 486: shouldn't we return nil here? It's not a bad thing.

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+
+	epoch, ok := n.epochs[block.BlockHeader().Epoch]
+	if !ok {
+		n.Logger.Debug("Received a QR from an Epoch that we have not validated", zap.Uint64("Epoch", block.BlockHeader().Epoch), zap.Uint64("Block Seq", block.BlockHeader().Seq), zap.Stringer("Block digest", block.BlockHeader().Digest))
```
yacovm: n.Logger.Debug("Received a QR from an Epoch that we have not validated", 
			zap.Uint64("Epoch", block.BlockHeader().Epoch), 
			zap.Uint64("Block Seq", block.BlockHeader().Seq), 
			zap.Stringer("Block digest", block.BlockHeader().Digest))

nonvalidator/non_validator.go
```diff
@@ -0,0 +1,562 @@
...
+			n.Logger.Debug("We can validate an epoch because we have received a threshold of messages of it.", zap.Stringer("Info", block.SealingBlockInfo()))
+			n.maybeValidateNextEpoch(block)
+
+			// We are storing a quorum round with a finalization we have not yet verified.
+			// We do this to tell the replicator a valid sequence exists and to begin replication if necessary.
+			// We will check the validity when we process this round.
+			n.sequenceReplicator.StoreQuorumRound(qr)
+		}
+
+		return nil
```
yacovm, line 492: can we extract this entire `! ok` block to a separate function?

nonvalidator/non_validator_test.go
```diff
@@ -0,0 +1,640 @@
...
+}
+
+func advanceUntil(nv *NonValidator, responder *nonValidatorResponderComm, seq uint64) {
+	startTime := nv.StartTime
+	for {
+		// Send any requests as responses back to the node
+		for msg, ok := responder.popResponse(); ok; {
+			// drop every other message
+			require.NoError(responder.t, nv.HandleMessage(msg.msg, msg.from))
+			msg, ok = responder.popResponse()
```
yacovm, line 628: why do we have this second pop?

nonvalidator/comm_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func (r *nonValidatorResponderComm) clearResponses() {
+	r.responsesLock.Lock()
+	defer r.responsesLock.Unlock()
+	r.responses = []*messageInfo{}
+}
+
+func (r *nonValidatorResponderComm) respondToDigestRequest(req *common.BlockDigestRequest, from common.NodeID) {
```
yacovm: this isn't a digest request though.

nonvalidator/non_validator.go
samliok: we return an error if the qr could not be processed in many places in this function. The callers of the function just log the error, but never propagate it up since as you said its not a bad thing.

nonvalidator/non_validator_test.go
samliok: we need to re-seed the msg variable. alternatively i could do this

for msg, ok := responder.popResponse(); ok; msg, ok = responder.popResponse() {

nonvalidator/comm_test.go
samliok: i think i can just delete this method since we dont send digest requests anymore.

nonvalidator/epochs.go
samliok: i combined the epoch replicator into the file. i'd like to keep this separated from the main non-validator flow for a little bit of organization

nonvalidator/non_validator.go
yacovm: it's fine let's keep as is

nonvalidator/non_validator.go
yacovm: ah right we're in `handleBlock`

yacovm (approved):

PR #402, opened 2026-06-09
Add handling for Auxiliary info

yacovm:
Introduce an AuxiliaryInfoApp interface abstraction that lets an application
piggyback application-specific data on Simplex epoch changes (e.g.
threshold distributed key generation), and gate the epoch transition on
that data being final. Final doesn't mean finalized as in Simplex,
but rather that the data is "good enough" to be used for the next epoch.

  Metadata & encoding:
  - Add AuxiliaryInfo (Info, PrevAuxInfoSeq, ApplicationID) to block
    metadata and an AppID type, with Canoto (de)serialization.
  - Each block's PrevAuxInfoSeq back-points to the most recent block with a
    non-empty Info, skipping empty-Info blocks. collectAuxiliaryInfo walks
    these pointers to rebuild the epoch's aux info history.

  The digest of the last AuxInfo in a final history is signed by the node,
  rather than the entire history of non-empty Aux Info.
  This is done for flexibility, as some applications like publicly verifiable DKG
  have commutative histories.

  State machine:
  - AuxiliaryInfoApp drives the flow: GenerateAuxInfo contributes to the
    history until IsFinalAuxInfoHistory reports it ready; only then are
    next-epoch approvals collected. IsLegalAuxInfoAppend validates a
    proposed append on the verify path. Each method also receives the
    next epoch's validator set (NodeBLSMappings).
  - The builder carries forward / extends aux info, and the verifier
    reconstructs the expected AuxiliaryInfo (ApplicationID, PrevAuxInfoSeq)
    and enforces it via the block-digest comparison.

  Approvals:
  - Approvals now bind to the auxiliary info: the signed payload is
    (NextPChainReferenceHeight, auxInfoDigest), where auxInfoDigest is the
    sha256 of the last aux info element in the history.
  - The ApprovalStore is now keyed by (NodeID, PChainHeight, AuxInfoSeqDigest)
    so approvals for different digests coexist, and sanitizeApprovals only
    aggregates approvals matching the candidate height and digest.

  Tests:
  - Verify approvals with real ECDSA signatures over the signed payload
    (the test signature verifier now actually checks signatures and
    (NextPChainReferenceHeight, auxInfoDigest), where auxInfoDigest is the
    sha256 of the final aux info history. assembleApprovalToBeSigned
    replaces the previous height-only payload.
    The switch to ECDSA is because this way we make sure we really sign
    the correct digest and not an empty one because of a bug.
  - The ApprovalStore is keyed by (NodeID, PChainHeight, AuxInfoSeqDigest)
    so approvals for different digests coexist, and sanitizeApprovals only
    aggregates approvals matching the candidate height and digest.

### DIFF
--- a/msm/approvals.go
+++ b/msm/approvals.go
@@ -4,15 +4,18 @@
 package metadata
 
 import (
-	"encoding/asn1"
-	"encoding/binary"
 	"fmt"
 
 	"github.com/ava-labs/simplex/common"
 	"go.uber.org/zap"
 )
 
-type approvalsByPChainHeight map[uint64]*approvalAndTimestamp
+type approvalKey struct {
+	pChainHeight  uint64
+	auxInfoDigest [32]byte
+}
+
+type approvalsByPChainHeightAndAuxInfoDigest map[approvalKey]*approvalAndTimestamp
 
 type approvalAndTimestamp struct {
 	ValidatorSetApproval
@@ -24,7 +27,7 @@ type ApprovalStore struct {
 	validators        NodeBLSMappings
 	logger            common.Logger
 	pkByNodeID        map[nodeID][]byte
-	approvalsByNodes  map[nodeID]approvalsByPChainHeight
+	approvalsByNodes  map[nodeID]approvalsByPChainHeightAndAuxInfoDigest
 	storedCount       int
 }
 
@@ -34,9 +37,9 @@ func NewApprovalStore(signatureVerifier SignatureVerifier, validators NodeBLSMap
 		pkByNodeID[vdr.NodeID] = vdr.BLSKey
 	}
 
-	approvalsByNodes := make(map[nodeID]approvalsByPChainHeight, len(validators))
+	approvalsByNodes := make(map[nodeID]approvalsByPChainHeightAndAuxInfoDigest, len(validators))
 	for _, vdr := range validators {
-		approvalsByNodes[vdr.NodeID] = make(approvalsByPChainHeight)
+		approvalsByNodes[vdr.NodeID] = make(approvalsByPChainHeightAndAuxInfoDigest)
 	}
 
 	return &ApprovalStore{
@@ -82,9 +85,14 @@ func (as *ApprovalStore) HandleApproval(approval *ValidatorSetApproval, timestam
 		return nil
 	}
 
+	key := approvalKey{
+		pChainHeight:  approval.PChainHeight,
+		auxInfoDigest: approval.AuxInfoDigest,
+	}
+
 	// Store the approval.
-	oldApproval := as.approvalsByNodes[approval.NodeID][approval.PChainHeight]
-	as.approvalsByNodes[approval.NodeID][approval.PChainHeight] = &approvalAndTimestamp{
+	oldApproval := as.approvalsByNodes[approval.NodeID][key]
+	as.approvalsByNodes[approval.NodeID][key] = &approvalAndTimestamp{
 		ValidatorSetApproval: *approval,
 		Timestamp:            timestamp,
 	}
@@ -113,22 +121,22 @@ func (as *ApprovalStore) maybePruneOldApprovals(approval *ValidatorSetApproval) 
 	}
 
 	if oldestApproval != nil {
+		key := approvalKey{
+			pChainHeight:  oldestApproval.PChainHeight,
+			auxInfoDigest: oldestApproval.AuxInfoDigest,
+		}
+
 		as.logger.Debug("Deleting old approval from node",
 			zap.String("nodeID", fmt.Sprintf("%x", oldestApproval.NodeID)),
 			zap.String("oldestApprovalPChainHeight",
 				fmt.Sprintf("%d", oldestApproval.PChainHeight)), zap.Uint64("oldestApprovalTimestamp", oldestApproval.Timestamp))
-		delete(as.approvalsByNodes[approval.NodeID], oldestApproval.PChainHeight)
+		delete(as.approvalsByNodes[approval.NodeID], key)
 		as.storedCount--
 	}
 }
 
 func (as *ApprovalStore) checkApprovalSignature(approval *ValidatorSetApproval, pk []byte) error {
-	pChainHeight := approval.PChainHeight
-	pChainHeightBuff := make([]byte, 8)
-	binary.BigEndian.PutUint64(pChainHeightBuff, pChainHeight)
-
-	signedMsg := common.SignedMessage{Payload: pChainHeightBuff, Context: signatureContext}
-	toBeSigned, err := asn1.Marshal(signedMsg)
+	toBeSigned, err := assembleApprovalToBeSigned(approval.PChainHeight, approval.AuxInfoDigest)
 	if err != nil {
 		return err
 	}
@@ -146,7 +154,13 @@ func (as *ApprovalStore) approvalExistsAndUpToDate(approval *ValidatorSetApprova
 	if as.approvalsByNodes[approval.NodeID] == nil {
 		return false
 	}
-	existingApproval := as.approvalsByNodes[approval.NodeID][approval.PChainHeight]
+
+	key := approvalKey{
+		pChainHeight:  approval.PChainHeight,
+		auxInfoDigest: approval.AuxInfoDigest,
+	}
+
+	existingApproval := as.approvalsByNodes[approval.NodeID][key]
 	if existingApproval == nil {
 		return false
 	}
--- a/msm/fuzz_test.go
+++ b/msm/fuzz_test.go
@@ -6,6 +6,7 @@ package metadata
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"testing"
 	"time"
 
@@ -114,6 +115,11 @@ func FuzzVerifyBlock(f *testing.F) {
 		// Mutating an authoritative field must make the block fail verification.
 		fuzzedMD := block.Metadata
 		field.set(&fuzzedMD, value)
+
+		if fieldIdx%2 == 1 && block.Metadata.AuxiliaryInfo == nil {
+			fuzzedMD.AuxiliaryInfo = &AuxiliaryInfo{PrevAuxInfoSeq: value}
+		}
+
 		if bytes.Equal(fuzzedMD.MarshalCanoto(), block.Metadata.MarshalCanoto()) {
 			t.Skip() // no-op mutation
 		}
@@ -184,6 +190,10 @@ func buildEpochChain(tb testing.TB, logger common.Logger) ([]*StateMachineBlock,
 	currentTime := startTime
 
 	sm, tc := newStateMachineWithLogger(tb, logger)
+	// This chain exercises the epoch lifecycle, not auxiliary info. Use an app whose
+	// history is always final so approvals are collected from the first collecting
+	// round (the builder only collects approvals once the aux info history is ready).
+	sm.AuxiliaryInfoApp = &noopTestAuxInfoApp{}
 	sm.GetValidatorSet = getValidatorSet
 	sm.GetPChainHeightForProposing = func() uint64 { return currentPChainHeight }
 	sm.GetPChainHeightForVerifying = func() uint64 { return currentPChainHeight }
@@ -244,21 +254,26 @@ func buildEpochChain(tb testing.TB, logger common.Logger) ([]*StateMachineBlock,
 	block3 := build(3, 2, 1, block2)
 	addBlock(3, block3, nil)
 
+	// The noopTestAuxInfoApp is always "ready" with an empty aux info history, so the candidate
+	// aux info digest the builder signs over is sha256 of the empty history. Peer approvals must
+	// carry the same digest to survive sanitizeApprovals' digest filter.
+	auxInfoDigest := sha256.Sum256(nil)
+
 	// block4 & block5: collecting-approvals blocks (1/3 then 2/3, not enough to seal).
-	approvalsResult = ValidatorSetApprovals{{NodeID: node1, PChainHeight: pChainHeight2, Signature: []byte("sig1")}}
+	approvalsResult = ValidatorSetApprovals{{NodeID: node1, PChainHeight: pChainHeight2, AuxInfoDigest: auxInfoDigest, Signature: signApproval(pChainHeight2, auxInfoDigest)}}
 	currentTime = startTime.Add(time.Second + 4*time.Millisecond)
 	tc.blockBuilder.block = nextInner(4)
 	block4 := build(4, 3, 1, block3)
 	addBlock(4, block4, nil)
 
-	approvalsResult = ValidatorSetApprovals{{NodeID: node2, PChainHeight: pChainHeight2, Signature: []byte("sig2")}}
+	approvalsResult = ValidatorSetApprovals{{NodeID: node2, PChainHeight: pChainHeight2, AuxInfoDigest: auxInfoDigest, Signature: signApproval(pChainHeight2, auxInfoDigest)}}
 	currentTime = startTime.Add(time.Second + 5*time.Millisecond)
 	tc.blockBuilder.block = nextInner(5)
 	block5 := build(5, 4, 1, block4)
 	addBlock(5, block5, nil)
 
 	// block6: the sealing block (3/3 approvals). Its successor is in stateBuildBlockEpochSealed.
-	approvalsResult = ValidatorSetApprovals{{NodeID: node3, PChainHeight: pChainHeight2, Signature: []byte("sig3")}}
+	approvalsResult = ValidatorSetApprovals{{NodeID: node3, PChainHeight: pChainHeight2, AuxInfoDigest: auxInfoDigest, Signature: signApproval(pChainHeight2, auxInfoDigest)}}
 	currentTime = startTime.Add(time.Second + 6*time.Millisecond)
 	tc.blockBuilder.block = nextInner(6)
 	block6 := build(6, 5, 1, block5)
@@ -286,6 +301,7 @@ func buildEpochChain(tb testing.TB, logger common.Logger) ([]*StateMachineBlock,
 
 	// Build a separate verifier MSM with its own copy of the fully populated store.
 	verifier, vtc := newStateMachineWithLogger(tb, logger)
+	verifier.AuxiliaryInfoApp = &noopTestAuxInfoApp{}
 	vtc.blockStore = tc.blockStore.clone()
 	verifier.GetBlock = vtc.blockStore.getBlock
 	verifier.GetValidatorSet = getValidatorSet
--- a/msm/msm.go
+++ b/msm/msm.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"slices"
 	"time"
 
 	"github.com/ava-labs/simplex/common"
@@ -87,6 +88,7 @@ var (
 	errTimestampTooBig                = errors.New("invalid timestamp: exceeds maximum int64 value")
 	errTimestampDecreasing            = errors.New("invalid timestamp: proposed timestamp is before parent block's timestamp")
 	errTimestampTooFarInFuture        = errors.New("invalid timestamp: proposed timestamp is too far in the future compared to current time")
+	errAuxInfoBlockRetrieval          = errors.New("failed to retrieve block while collecting auxiliary info")
 
 	signatureContext = "MSM approval"
 )
@@ -162,6 +164,30 @@ type BlockBuilder interface {
 	WaitForPendingBlock(ctx context.Context)
 }
 
+// AuxiliaryInfoGenVerifier abstracts the application-specific logic for generating and verifying auxiliary information
+// that is piggybacked on epoch transitions.
+type AuxiliaryInfoGenVerifier interface {
+	// IsLegalAppend checks whether the given auxiliary information byte slice [x]
+	// can be appended to the history of auxiliary information for the given versionID, according to the app's rules.
+	// Returns nil if the append is legal, or an error if the append is not legal or if any error occurs during the check.
+	IsLegalAppend(versionID VersionID, nodes NodeBLSMappings, history [][]byte, x []byte) error
+
+	// IsSufficient checks whether the given history of auxiliary information for the given versionID is sufficient
+	// to start the epoch transition process.
+	IsSufficient(versionID VersionID, nodes NodeBLSMappings, history [][]byte) (bool, error)
+
+	// Generate generates an auxiliary information encoded as a byte slice based on the history of auxiliary information
+	// for the given versionID in the current epoch so far.
+	// If this is the first invocation in the epoch, DefaultversionID() should be passed as the VersionID.
+	// Otherwise, the versionID from previous blocks in the epoch should be used.
+	// If the application deems the given history to be sufficient for the epoch change, it can return a nil byte slice,
+	// in which case it will not be appended to the history.
+	Generate(versionID VersionID, nodes NodeBLSMappings, history [][]byte) ([]byte, error)
+
+	// DefaultVersionID returns the default VersionID that should be used for epochs that don't have any any auxiliary information yet.
+	DefaultVersionID() VersionID
+}
+
 // StateMachine manages block building and verification across epoch transitions.
 type StateMachine struct {
 	*Config
@@ -213,6 +239,8 @@ type Config struct {
 	Signer common.Signer
 	// ComputeICMEpoch computes the ICM epoch information in order to know which P-chain height to encode.
 	ComputeICMEpoch ICMEpochTransition
+	// AuxiliaryInfoApp abstracts an application that piggybacks on epoch changes.
+	AuxiliaryInfoApp AuxiliaryInfoGenVerifier
 }
 
 type state uint8
@@ -454,7 +482,7 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
 		}
 	}
 
-	return wrapBlock(innerBlock, newSimplexEpochInfo, decisionToBuildBlock.pChainHeight, simplexMetadata, simplexBlacklist, now, icmEpochInfo), nil
+	return wrapBlock(innerBlock, newSimplexEpochInfo, decisionToBuildBlock.pChainHeight, simplexMetadata, simplexBlacklist, now, icmEpochInfo, nil), nil
 }
 
 func computeICMEpochInfo(parentBlock StateMachineBlock, computeICMEpoch ICMEpochTransition, childTimestamp time.Time) ICMEpochInfo {
@@ -481,6 +509,7 @@ func verifyAgainstExpected(
 	nextBlock *StateMachineBlock,
 	timestamp time.Time,
 	expectedIcmEpochInfo ICMEpochInfo,
+	auxInfo *AuxiliaryInfo,
 ) error {
 	if innerBlock != nil {
 		if err := innerBlock.Verify(ctx, expectedIcmEpochInfo.PChainEpochHeight); err != nil {
@@ -489,7 +518,7 @@ func verifyAgainstExpected(
 	}
 	expectedBlock := wrapBlock(
 		innerBlock, expectedSimplexEpochInfo, expectedPChainHeight,
-		nextBlock.Metadata.SimplexProtocolMetadata, nextBlock.Metadata.SimplexBlacklist, timestamp, expectedIcmEpochInfo)
+		nextBlock.Metadata.SimplexProtocolMetadata, nextBlock.Metadata.SimplexBlacklist, timestamp, expectedIcmEpochInfo, auxInfo)
 	if expectedBlock.Digest() != nextBlock.Digest() {
 		return fmt.Errorf("expected block digest %s does not match proposed block digest %s: %w",
 			expectedBlock.Digest(),
@@ -517,7 +546,7 @@ func (sm *StateMachine) verifyNormalBlock(ctx context.Context, parentBlock State
 	}
 	newSimplexEpochInfo.NextPChainReferenceHeight = nextBlock.Metadata.SimplexEpochInfo.NextPChainReferenceHeight
 
-	return verifyAgainstExpected(ctx, nextBlock.InnerBlock, newSimplexEpochInfo, proposedPChainHeight, nextBlock, timestamp, icmEpochInfo)
+	return verifyAgainstExpected(ctx, nextBlock.InnerBlock, newSimplexEpochInfo, proposedPChainHeight, nextBlock, timestamp, icmEpochInfo, nil)
 }
 
 func verifyPChainHeight(proposedPChainHeight uint64, currentPChainHeight uint64, prevPChainHeight uint64) error {
@@ -755,6 +784,7 @@ func (sm *StateMachine) buildBlockZero(parentBlock StateMachineBlock, simplexMet
 		simplexBlacklist,
 		time.UnixMilli(timestamp),
 		icmEpochInfo,
+		nil,
 	), nil
 }
 
@@ -799,6 +829,7 @@ func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock Stat
 		block.Metadata.SimplexBlacklist,
 		time.UnixMilli(int64(block.Metadata.Timestamp)),
 		expectedICMEpochInfo,
+		nil,
 	)
 
 	if expectedBlock.Digest() != block.Digest() {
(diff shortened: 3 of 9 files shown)


msm/msm.go
```diff
@@ -1008,33 +1070,116 @@ func (sm *StateMachine) computeNewApprovals(parentBlock StateMachineBlock) (*app
...
+	// if its parent block has a non-empty Info field, then the block's PrevAuxInfoSeq points to its parent block.
+	// Else, its parent block has an empty Info field, then the block's PrevAuxInfoSeq is inherited from its parent block's PrevAuxInfoSeq.
+
+	auxInfo := block.Metadata.AuxiliaryInfo
+	currentSeq := startSeq
+	for auxInfo != nil {
+		if len(auxInfo.Info) > 0 {
+			history = append(history, auxInfo.Info)
+			seqs = append(seqs, currentSeq)
+			appID = auxInfo.ApplicationID
```
samliok: im not sure I understand application id. I'm assuming it helps the user of `AuxInfo` to decode the info bytes. So for example DKG would have an application id of `1`, and then we would decode all []info bytes in the history message with application id `1` a certain way. 

Can't we have multiple application running though? Maybe a DKG and something else in the future so we would have multiple application ids corresponding within the same epoch?

samliok: or are you saying if we have multiple applications, they would both be serialized into a single info []byte? in this case the application id is just used for versioning of how to deserialize the bytes?

msm/msm.go
```diff
@@ -1008,33 +1070,116 @@ func (sm *StateMachine) computeNewApprovals(parentBlock StateMachineBlock) (*app
...
+}
+
+// collectAuxiliaryInfo traverses backwards starting from the given block and collects the AuxiliaryInfo of all blocks in the chain.
+// returns the collected AuxiliaryInfo, the corresponding sequences of the blocks they were collected from,
+// and the application ID of the most recent non-empty Info block (or defaultAppID if there was none).
+func collectAuxiliaryInfo(block StateMachineBlock, startSeq uint64, getBlock BlockRetriever, defaultAppID AppID) ([][]byte, []uint64, AppID, error) {
+	var history [][]byte
+	// seqs[i] is the block sequence that history[i] was collected from, so the caller can point a
+	// new block's PrevAuxInfoSeq at the most recent non-empty Info block (the last item, after reversing).
+	var seqs []uint64
```
samliok: do you think it would be easier to reason through if we created a struct 
{
[]byte
seq
}

rather than maintaining two separate arrays? it would also ensure the sorting is done correctly for the history since we can just sort by seq in this new struct

msm/msm.go
```diff
@@ -1008,33 +1070,116 @@ func (sm *StateMachine) computeNewApprovals(parentBlock StateMachineBlock) (*app
...
+	sig, err := sm.Signer.Sign(toBeSigned)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign approval: %w", err)
+	}
+	return sig, nil
+}
+
+// collectAuxiliaryInfo traverses backwards starting from the given block and collects the AuxiliaryInfo of all blocks in the chain.
+// returns the collected AuxiliaryInfo, the corresponding sequences of the blocks they were collected from,
+// and the application ID of the most recent non-empty Info block (or defaultAppID if there was none).
```
samliok: i think the applicationID returned is actually the oldest

msm/approvals.go
```diff
@@ -4,15 +4,25 @@
...
 	"fmt"
 
 	"github.com/ava-labs/simplex"
 	"go.uber.org/zap"
 )
 
-type approvalsByPChainHeight map[uint64]*approvalAndTimestamp
+// uint64DigestPair is a struct that contains the P-Chain height and the digest
+// of the auxiliary info for which a validator set approval is given.
+type uint64DigestPair [40]byte
```
samliok: what if we just had a type `approvalKey` 

```go
type approvalKey struct {
	pChainHeight     uint64
	auxInfoSeqDigest [32]byte
}
```

then we wouldn't need to have the byte logic `makeUint64DigestPair` and it would work in the map `approvalsByPChainHeightAndAuxInfoDigest` by default

msm/msm.go
```diff
@@ -848,11 +867,31 @@ func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock Stat
...
+	}
+
+	var newApprovals *approvals
+	if isAuxInfoReadyForEpochTransition {
+		newApprovals, err = sm.computeNewApprovals(parentBlock, validators, auxInfoDigest)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// We're not ready for epoch transition yet, so putting a zero-valure approvals here
```
```suggestion
		// We're not ready for epoch transition yet, so putting a zero-value approvals here
```

msm/msm.go
```diff
@@ -162,14 +164,26 @@ type BlockBuilder interface {
...
 
-type verificationInput struct {
-	prevMD              StateMachineMetadata
-	proposedBlockMD     StateMachineMetadata
-	hasInnerBlock       bool
-	innerBlockTimestamp time.Time // only set when hasInnerBlock is true
-	prevBlockSeq        uint64
-	nextBlockType       BlockType
-	state               state
+type AuxiliaryInfoApp interface {
```
samliok: im not sure what `App` is supposed to mean in the context of this interface, maybe we can think of a better sufix?

msm/msm.go
```diff
@@ -1233,7 +1378,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+			ApplicationID:  parentAuxInfo.ApplicationID,
+			PrevAuxInfoSeq: lastAuxInfoSeq(auxInfoSeqs),
+		}
+	}
+
+	if !isAuxInfoReadyForEpochTransition {
+		// If the auxiliary info isn't ready for epoch transition,
+		// we should focus on contributing to finalizing it before collecting approvals for the epoch transition,
+		// as without it being ready, we won't be able to transition epochs anyway.
+		auxInf, err := sm.AuxiliaryInfoApp.GenerateAuxInfo(appID, validators, auxInfoHistory)
```
samliok: so if we are not done with generating the `AuxInfo` we will add our own info to hopefully help finish it up. How does this work in the dkg? Does everyone need to contribute to the DKG setup? i assume this cant be the case because then an adversary can just block the epoch from progressing.

msm/msm.go
```diff
@@ -1233,7 +1378,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+	var digest [32]byte
+	if auxInfoReady {
+		digest = sha256.Sum256(lastHistory(auxInfoHistory))
+	}
+
+	return expectedAuxInfo, digest, auxInfoReady, nil
+}
+
+// computeAuxInfo computes the AuxiliaryInfo that should be included in the block being built, and whether the auxiliary info history is ready for epoch transition,
+func (sm *StateMachine) computeAuxInfo(parentBlock StateMachineBlock, prevBlockSeq uint64, validators NodeBLSMappings) (*AuxiliaryInfo, bool, [32]byte, error) {
```
samliok: should this be simplex.Digest?

msm/msm.go
```diff
@@ -1233,7 +1378,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+			auxInfo.Info = auxInf
+		}
+	}
+
+	var auxInfoDigest [32]byte
+	if isAuxInfoReadyForEpochTransition {
+		auxInfoDigest = sha256.Sum256(lastHistory(auxInfoHistory))
+	}
+
+	return auxInfo, isAuxInfoReadyForEpochTransition, auxInfoDigest, nil
```
samliok, line 1488: isn't `isAuxInfoReadyForEpochTransition` implied from auxInfoDigest != empty

msm/msm.go
```diff
@@ -848,11 +867,31 @@ func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock Stat
...
+	var newApprovals *approvals
+	if isAuxInfoReadyForEpochTransition {
+		newApprovals, err = sm.computeNewApprovals(parentBlock, validators, auxInfoDigest)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// We're not ready for epoch transition yet, so putting a zero-valure approvals here
+		// makes us stay in the collecting approvals state without contributing any approvals.
+		newApprovals = &approvals{}
```
samliok, line 896: one thing i don't get is the aux info is only contributed by the nodes in the current epoch? but for dkg, we want to collect info from nodes in the next epoch? I don't see where we are doing this, as the block builders are only nodes of the current epoch

msm/msm.go
```diff
@@ -1233,7 +1378,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+	return verifyAgainstExpected(ctx, nextBlock.InnerBlock, newSimplexEpochInfo, proposedPChainHeight, nextBlock, timestamp, icmEpochInfo, nil)
+}
+
+// computeExpectedAuxInfoForApprovalCollection computes the expected AuxiliaryInfo that should be included in the proposed block
+// for approval collection, and returns the auxiliary info digest, and whether the auxiliary info history is ready for epoch transition.
+func (sm *StateMachine) computeExpectedAuxInfoForApprovalCollection(parentBlock StateMachineBlock, nextBlock *StateMachineBlock, prevBlockSeq uint64, validators NodeBLSMappings) (*AuxiliaryInfo, [32]byte, bool, error) {
+	nextMD := nextBlock.Metadata
+	prevMD := parentBlock.Metadata
+
+	auxInfoHistory, auxInfoSeqs, appID, err := collectAuxiliaryInfo(parentBlock, prevBlockSeq, sm.GetBlock, sm.AuxiliaryInfoApp.DefaultAppID())
```
samliok: moreso a general thought, but I feel like this logic is a bit complex to test and reason about in scenarios where we fork due to an empty round timeout. 

What happens when we collect all the auxiliary information, but the next block is not built of our previous block and proposes something completely different? do we have tests for scenarios like this?

msm/msm.go
yacovm: Suppose we implement DKG in avalanchego version v1.16.1 and then in version v1.16.2 we improve the DKG implementation or fix a bug that requires changing the encoding.

A node that is mid epoch change and is restarted from v1.16.1 to v1.16.2 needs to continue using the old DKG implementation until the epoch change ends. It cannot use the new implementation mid epoch change. 

The application ID is there to ensure every epoch change uses a uniform version of DKG implementation.

msm/msm.go
yacovm: yeah we can do that. Initially I only had the history and then added the seqs. Will change

msm/msm.go
yacovm: yeah, you're right and that's by design. Will fix

msm/approvals.go
yacovm: makes sense

msm/msm.go
yacovm: We'll add our own, but it's a bit tricky because the actual implementation of `GenerateAuxInfo ` needs to make sure we don't add our own twice or more.

> How does this work in the dkg? Does everyone need to contribute to the DKG setup?

We need enough randomness such that we consider it non-biased and non-colluded. 

> i assume this cant be the case because then an adversary can just block the epoch from progressing.

Unless we have synchrony assumptions (which we don't)

msm/msm.go
yacovm: no, `auxInfoDigest` is just the last one, but it doesn't mean we are ready.

msm/msm.go
yacovm: Yeah that's a very good question. 

So, in the current implementation we cannot do that yet because this will require something similar to what we have with the approvals (an interface that lists approvals that reside in our approval memory pool).

However, when it comes to DKG then if the validator sets of the two adjacent epochs aren't too different, we can use the validator set of the previous epoch. 

It's not hard to expand in a later PR to add the Aux Info contribution of nodes of the next epoch, but one thing at a time...

msm/msm.go
yacovm: > moreso a general thought, but I feel like this logic is a bit complex to test and reason about in scenarios where we fork due to an empty round timeout.
> What happens when we collect all the auxiliary information, but the next block is not built of our previous block and proposes something completely different? do we have tests for scenarios like this?

Good question. 

I did think about this scenario when I implemented it, and I don't think we need to do anything special here. 

Essentially, as long as blocks cannot be re-ordered (have their parents changed) after they are built, there is no problem that we can have multiple competing histories. 

While a block may not end up being part of a chain, every block that is built on top of a parent block will never have its parent changed, so the checks we have that look at the entire history of auxiliary info are always valid in case the block becomes finalized, and if the block doesn't become finalized and ends up being dropped (rewinding the chain), that's fine.

Not every kind of protocol can support safe rewinding (for example, in a commit-reveal protocol, rewinding is fatal) but the kinds of protocol that we aim to support only involve putting encrypted random shares on-chain, so any item that doesn't end up part of the chain (rewinded) just doesn't end up being part of the DKG, and is anyway comprised of random data, so no information can be learned by an adversary that rewinds the chain.

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/981edb3e6909714041f7a26fc45233b1fa2a9406#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R1111-R1113

https://github.com/ava-labs/Simplex/pull/402/changes/981edb3e6909714041f7a26fc45233b1fa2a9406#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R1133

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/981edb3e6909714041f7a26fc45233b1fa2a9406#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R1132

msm/approvals.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/981edb3e6909714041f7a26fc45233b1fa2a9406#diff-5e383ab850834861309c7e77b4123dbe73b6b4d3e2b96f96832869f5847b48fdR13-R15

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/981edb3e6909714041f7a26fc45233b1fa2a9406#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R892-R893

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/981edb3e6909714041f7a26fc45233b1fa2a9406#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R169

https://martinfowler.com/bliki/TwoHardThings.html

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/981edb3e6909714041f7a26fc45233b1fa2a9406#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R1439

msm/msm.go
samliok: makes sense, would it be clearer to name this variable `VersionID` or even just `Version`?

msm/msm.go
samliok: cache coherence is easy, naming things is way harder

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/commits/2fbeae24f006b4c5cfec89f2423e9845ceebecb4

msm/msm.go
samliok: i dont see this change in the diff

msm/msm.go
```diff
@@ -1008,33 +1072,115 @@ func (sm *StateMachine) computeNewApprovals(parentBlock StateMachineBlock) (*app
...
+	if len(aih.data) == 0 {
+		return nil
+	}
+	return aih.data[len(aih.data)-1]
+}
+
+// collectAuxiliaryInfo traverses backwards starting from the given block and collects the AuxiliaryInfo of all blocks in the chain.
+// returns the collected AuxiliaryInfo, the corresponding sequences of the blocks they were collected from,
+// and the application ID of the oldest block that contains a non empty Info (or defaultversionID if there was none).
+func collectAuxiliaryInfo(block StateMachineBlock, startSeq uint64, getBlock BlockRetriever, defaultversionID VersionID) (auxInfoHistory, VersionID, error) {
```
```suggestion
func collectAuxiliaryInfo(block StateMachineBlock, startSeq uint64, getBlock BlockRetriever, defaultVersionID VersionID) (auxInfoHistory, VersionID, error) {
```

msm/msm.go
```diff
@@ -1233,7 +1379,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+	isAuxInfoReadyForEpochTransition, err := sm.AuxiliaryInfoApp.IsFinalAuxInfoHistory(versionID, validators, auxInfoHistory.data)
+	if err != nil {
+		return nil, false, common.Digest{}, fmt.Errorf("failed to check if auxiliary info history is final: %w", err)
+	}
+
+	var auxInfo *AuxiliaryInfo
+	parentAuxInfo := parentBlock.Metadata.AuxiliaryInfo
+	if parentAuxInfo != nil {
+		auxInfo = &AuxiliaryInfo{
+			VersionID:      parentAuxInfo.VersionID,
```
samliok, line 1456: so here we use the version ID from the parent, but in `collectAuxillaryInfo` we check against the oldest version id.

samliok: do we have any tests for upgrading the version in the middle of an epoch, or is this not testable until we implement `AuxiliaryInfoGenVerifier`

msm/msm.go
```diff
@@ -1233,7 +1379,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+			PrevAuxInfoSeq: auxInfoHistory.lastAuxInfoSeq(),
+		}
+	}
+
+	if !isAuxInfoReadyForEpochTransition {
+		// If the auxiliary info isn't ready for epoch transition,
+		// we should focus on contributing to finalizing it before collecting approvals for the epoch transition,
+		// as without it being ready, we won't be able to transition epochs anyway.
+		auxInf, err := sm.AuxiliaryInfoApp.GenerateAuxInfo(versionID, validators, auxInfoHistory.data)
```
samliok: should we call `IsLegalAuxInfoAppend` after generating the `AuxInfo`? not sure if this function should be called like similar to block.verify, where we need to call it after building a block.

msm/msm.go
```diff
@@ -162,14 +164,28 @@ type BlockBuilder interface {
...
-	prevMD              StateMachineMetadata
-	proposedBlockMD     StateMachineMetadata
-	hasInnerBlock       bool
-	innerBlockTimestamp time.Time // only set when hasInnerBlock is true
-	prevBlockSeq        uint64
-	nextBlockType       BlockType
-	state               state
+// AuxiliaryInfoGenVerifier abstracts the application-specific logic for generating and verifying auxiliary information
+// that is piggybacked on epoch transitions.
+type AuxiliaryInfoGenVerifier interface {
```
samliok, line 169: what about an interface with this api?

```go

  type AuxiliaryInfoCollector interface {
      Generate(versionID VersionID, nodes NodeBLSMappings, history [][]byte) ([]byte, error)
      IsSufficient(versionID VersionID, nodes NodeBLSMappings, history [][]byte) (bool, error)
      IsLegalAppend(versionID VersionID, nodes NodeBLSMappings, history [][]byte, auxInfoBytes []byte) error
      DefaultVersionID() VersionID
  }

```

msm/msm.go
```diff
@@ -1233,7 +1379,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+
+	if !isAuxInfoReadyForEpochTransition {
+		// If the auxiliary info isn't ready for epoch transition,
+		// we should focus on contributing to finalizing it before collecting approvals for the epoch transition,
+		// as without it being ready, we won't be able to transition epochs anyway.
+		auxInf, err := sm.AuxiliaryInfoApp.GenerateAuxInfo(versionID, validators, auxInfoHistory.data)
+		if err != nil {
+			return nil, false, common.Digest{}, fmt.Errorf("failed to generate auxiliary info: %w", err)
+		}
+		if auxInfo == nil {
```
samliok, line 1469: i don't get how we can have `isAuxInfoReadyForEpochTransition` and then get a `auxInfo == nil`. Is this edge case essentially saying "hey we need to build some aux info, but i am unable to build it".

msm/msm.go
```diff
@@ -1233,7 +1379,111 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
+	var digest [32]byte
+	if auxInfoReady {
+		digest = sha256.Sum256(auxInfoHistory.lastHistory())
+	}
+
+	return expectedAuxInfo, digest, auxInfoReady, nil
+}
+
+// computeAuxInfo computes the AuxiliaryInfo that should be included in the block being built, and whether the auxiliary info history is ready for epoch transition,
+func (sm *StateMachine) computeAuxInfo(parentBlock StateMachineBlock, prevBlockSeq uint64, validators NodeBLSMappings) (*AuxiliaryInfo, bool, common.Digest, error) {
```
samliok, line 1441: This method feels overly complex. It seems like some of this logic could be abstracted into Generate, although I'm not entirely sure what the ideal shape would be.

For example, why can't `Generate` simply return `auxInfo, error`? The implementation already knows whether `IsFinalAuxInfoHistory` returns, so it should be able to determine internally whether auxiliary info should be generated or not, rather than requiring the caller to handle that logic.

msm/approvals.go
```diff
@@ -4,15 +4,18 @@
...
 	"fmt"
 
 	"github.com/ava-labs/simplex/common"
 	"go.uber.org/zap"
 )
 
-type approvalsByPChainHeight map[uint64]*approvalAndTimestamp
+type approvalKey struct {
+	pChainHeight     uint64
+	auxInfoSeqDigest [32]byte
```
samliok: what is `auxInfoSeqDigest`? is it the combination of an auxInfo (simplex.Digest) + seq (uint64)? if so shouldn't it be 40 bytes?

msm/msm.go
```diff
@@ -1008,33 +1072,115 @@ func (sm *StateMachine) computeNewApprovals(parentBlock StateMachineBlock) (*app
...
+	sig, err := sm.Signer.Sign(toBeSigned)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign approval: %w", err)
+	}
+	return sig, nil
+}
+
+type auxInfoHistory struct {
+	data [][]byte
+	seq  []uint64
```
samliok: Should we store `lastSeq`instead of an array here, since we only ever use the last sequence?

msm/msm.go
```diff
@@ -1008,33 +1072,115 @@ func (sm *StateMachine) computeNewApprovals(parentBlock StateMachineBlock) (*app
...
+	// if its parent block has a non-empty Info field, then the block's PrevAuxInfoSeq points to its parent block.
+	// Else, its parent block has an empty Info field, then the block's PrevAuxInfoSeq is inherited from its parent block's PrevAuxInfoSeq.
+
+	auxInfo := block.Metadata.AuxiliaryInfo
+	currentSeq := startSeq
+	for auxInfo != nil {
+		if len(auxInfo.Info) > 0 {
+			history = append(history, auxInfo.Info)
+			seqs = append(seqs, currentSeq)
+			versionID = auxInfo.VersionID
```
samliok, line 1151: if the version changes between blocks in the same epoch, isn't that a problem?

msm/msm.go
samliok: also if the struct implementing the interface already has the `DefaultVersionID` function, why do we need to call it, then pass it back to the other functions?

we are basically doing 
```go
version := GetDefaultVersion()
version = oldestVersion ? oldestVersion : version
IsLegalAuxInfoAppend(version) 
```
It seems like we will need to associate the version & epoch in the struct implementing the interface anyways, so why pass it in everywhere?

msm/msm.go
```diff
@@ -489,6 +507,7 @@ func verifyAgainstExpected(
 	nextBlock *StateMachineBlock,
 	timestamp time.Time,
 	expectedIcmEpochInfo ICMEpochInfo,
+	auxInfo *AuxiliaryInfo,
```
samliok, line 512: this feels a bit weird to verify the `auxInfo` against expected mainly because aux info cannot be predicted before hand. This param seems unnecessary then because we already do the required checks for `auxInfo` in `expectedAuxInfoEntry`.

Maybe we can change `expectedAuxInfoEntry()` to `verifyExpectedAuxInfo() error` and simplify this `verifyAgainstExpected` function

msm/msm.go
samliok: i banged my head against a wall, then played around with some refactors I thought could make the code simpler / easy to reason through for me. 

heres a pr of a bunch of random changes(i have not reviewed the pr), but maybe a few of them could be useful ideas? the pr also incorporates a couple changes i mentioned in other comments, so its a random assortment of my artifacts via review. Feel free to disregard any if u think they are irrelevant/not useful. 

https://github.com/ava-labs/Simplex/pull/411/changes

msm/msm.go
yacovm: `collectAuxiliaryInfo` checks if we have auxiliary information in this epoch. 

it returns the default version ID if we don't have auxiliary information in the epoch.
In case we end up upgrading the auxiliary information app, we'll also change the default version ID (`.DefaultVersionID()`) and then it will return that. 

The idea is to fallback to the older version ID if we already have a block in the epoch with a version ID.

msm/msm.go
yacovm: I don't see why. It is the responsibility of the implementer of the interface to make sure it produces auxiliary info that passes verification checks.

msm/msm.go
yacovm: The vacuous case where we don't use / need the aux info, we are always ready despite not generating any aux info.

msm/msm.go
yacovm: > For example, why can't Generate simply return auxInfo, error? The implementation already knows whether IsFinalAuxInfoHistory returns, so it should be able to determine internally whether auxiliary info should be generated or not, rather than requiring the caller to handle that logic.

Are you suggesting that `Generate` will check if `IsSufficient` will return true, and if so, it will just not output anything? We still need `IsSufficient` because we don't want to call `Generate` when we verify, right?

msm/approvals.go
yacovm: whoops, a mistake caused by find-replace

msm/msm.go
yacovm: This is what we're trying to facilitate. 

That if we restart the nodes with a higher version, they will operate in the older one and not upgrade the version mid epoch.

msm/msm.go
yacovm: In most cases, the aux info is supposed to be nil. Therefore we have to make sure it is nil.

msm/msm.go
yacovm: changed accordingly 
https://github.com/ava-labs/Simplex/pull/402/changes/3cdd8e0106a440ba96aeb4000b877a9f290e74b1#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R170-R179

msm/approvals.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/3cdd8e0106a440ba96aeb4000b877a9f290e74b1#diff-5e383ab850834861309c7e77b4123dbe73b6b4d3e2b96f96832869f5847b48fdL14-L15

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/3cdd8e0106a440ba96aeb4000b877a9f290e74b1#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R1113

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/3cdd8e0106a440ba96aeb4000b877a9f290e74b1#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R1126

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/changes/3cdd8e0106a440ba96aeb4000b877a9f290e74b1#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65R1111-R1113

msm/msm.go
samliok: i was moreso thinking that Generate would simply return nil if there was nothing to generate and we were ready to transition to an epoch. But I guess this doesn't work because we may return nil and not be ready to transition to an epoch?

msm/msm.go
samliok: oh i meant to say `!isAuxInfoReadyForEpochTransition` and then get a `auxInfo == nil`

msm/msm.go
```diff
@@ -1008,33 +1072,114 @@ func (sm *StateMachine) computeNewApprovals(parentBlock StateMachineBlock) (*app
...
+	// (4) When a block with an empty Info field is built on a parent block that has AuxiliaryInfo,
+	// if its parent block has a non-empty Info field, then the block's PrevAuxInfoSeq points to its parent block.
+	// Else, its parent block has an empty Info field, then the block's PrevAuxInfoSeq is inherited from its parent block's PrevAuxInfoSeq.
+
+	auxInfo := block.Metadata.AuxiliaryInfo
+	currentSeq := startSeq
+	for auxInfo != nil {
+		if len(auxInfo.Info) > 0 {
+			history = append(history, auxInfo.Info)
+			seqs = append(seqs, currentSeq)
```
samliok: instead of having making an array for seqs, we could just set `lastSeq = currentSeq` here

msm/fuzz_test.go
```diff
@@ -242,21 +247,26 @@ func buildEpochChain(tb testing.TB, logger common.Logger) ([]*StateMachineBlock,
 	block3 := build(3, 2, 1, block2)
 	addBlock(3, block3, nil)
 
+	// The noopTestAuxInfoApp is always "ready" with an empty aux info history, so the candidate
+	// aux info digest the builder signs over is sha256 of the empty history. Peer approvals must
+	// carry the same digest to survive sanitizeApprovals' digest filter.
+	auxInfoDigest := sha256.Sum256(nil)
```
samliok, line 260: should we add a case to the fuzzer that isn't a `noopTestAuxInfoApp`? I think it would make sense fuzzing the different scenarios we can get from the auxInfo as well.

msm/util_test.go
```diff
@@ -375,56 +445,70 @@ func (failingAggregator) IsQuorum([]common.NodeID) bool {
...
-	return b.height
+func (t *noopTestAuxInfoApp) DefaultVersionID() VersionID {
+	return 1
 }
 
-func (b *testVMBlock) Timestamp() time.Time {
-	return time.Now()
+type voteCountingAuxInfoApp struct {
+	threshold  int
+	randomTape func() []byte
```
samliok, line 472: not sure I understand what `randomTape` implies, especially since our only implementation of it is deterministic. Can we make a more clearer name and maybe a brief comment on what this auxiliary app is supposed to represent?

msm/msm.go
samliok: > do we have any tests for upgrading the version in the middle of an epoch, or is this not testable until we implement AuxiliaryInfoGenVerifier

makes sense, I still think we should have a test where the version ID varies between blocks in an epoch if possible

msm/util_test.go
yacovm: random tape is just a way to inject randomness. It's a term I borrowed from [PTM](https://en.wikipedia.org/wiki/Probabilistic_Turing_machine)s.

The idea is that an algorithm that is non deterministic can be structured around a deterministic rule set and its non determinism stems from reading a random tape that someone prepared beforehand.

msm/msm.go
yacovm: I'm not sure that's a good API design, because look for example at a case where we have already acted in a DKG protocol and our own input is part of the history. Due to crashes and network conditions, we have a series of empty blocks and now we are back to be the block proposer. 

We call Generate and we get back nil, because we are not supposed to generate crypto material twice in the same epoch. 

Yet, we may not be ready to transition to a new epoch because we are still waiting on other nodes to act.

msm/msm.go
yacovm: notice that `auxInfo` is not `auxInf`.

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/commits/3a0eaf1263c8dd10f78248d3d2ccf72cf44f71b9

msm/fuzz_test.go
yacovm: There is no way to fuzz fields that are not authoritative fields: 

```
// authoritativeField names a consensus field that the verifier reconstructs itself
// (from the parent block and the state machine), rather than copying it from the
// proposer.
```

Auxiliary info is not an authoritative field so we can verify its content using the `AuxiliaryInfoGenVerifier` API but we cannot reconstruct it ourselves, therefore we cannot fuzz it because any validity decision is entirely up to `AuxiliaryInfoGenVerifier`.

Having said that, most of the states require auxiliary info to be nil, so I can add something like:

```
		if fieldIdx % 2 == 1 && block.Metadata.AuxiliaryInfo == nil {
			fuzzedMD.AuxiliaryInfo = &AuxiliaryInfo{PrevAuxInfoSeq: value}
		}
```

Because `nil` is a value we do enforce in some states.

yacovm: https://github.com/ava-labs/Simplex/pull/402/commits/d15309a184568f6973df02ca6777d6d7a43f27ee

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/402/commits/2f2a2a08bd09d2c1e02c9f3cfd8b9e997257c1a3

samliok (approved):

PR #407, opened 2026-06-15
Pass validator membership to QuorumCertificate.Verify()

yacovm:
QuorumCertificates were carried around the network already parsed, which forced parsing to happen before a message reached the epoch.

Since a QC has an ability to verify its own signature, it implicitly means that a QC implementation needs knowledge of the public keys for the epoch the QC was created for.

Instead of wiring the public keys of the epoch to the QC, the QuorumCertificate.Verify() now receives the set of nodes so the verifier has access to validator membership (and their public keys) when checking a quorum certificate. The nodes argument is threaded through verifyContextQC, the Finalization/Notarization/EmptyNotarization Verify() methods, the verifiableMessage interface, and VerifyQC, with e.nodes supplied at every call site in epoch.go.

To support this, common.Node is reworked:
  - rename the Node.Node field to Node.Id to disambiguate it from the surrounding type
  - add a PK []byte field to carry each node's public key

### DIFF
--- a/common/api.go
+++ b/common/api.go
@@ -147,15 +147,16 @@ type Nodes []Node
 func (nws Nodes) NodeIDs() []NodeID {
 	nodes := make([]NodeID, len(nws))
 	for i, nw := range nws {
-		nodes[i] = nw.Node
+		nodes[i] = nw.Id
 	}
 	return nodes
 }
 
-// Node is a struct that pairs a node with its weight in the signature aggregator.
+// Node is a struct that pairs a node ID with its weight and public key.
 type Node struct {
-	Node   NodeID
+	Id     NodeID
 	Weight uint64
+	PK     []byte
 }
 
 // SignatureAggregatorCreator creates a SignatureAggregator from a list of nodes and their weights.
--- a/simplex/util.go
+++ b/simplex/util.go
@@ -19,7 +19,7 @@ import (
 )
 
 type verifiableMessage interface {
-	Verify() error
+	Verify(nodes common.Nodes) error
 }
 
 // RetrieveLastIndexFromStorage retrieves the latest block and finalization from storage.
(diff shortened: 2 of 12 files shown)


common/api.go
```diff
@@ -147,15 +147,16 @@ type Nodes []Node
 func (nws Nodes) NodeIDs() []NodeID {
 	nodes := make([]NodeID, len(nws))
 	for i, nw := range nws {
-		nodes[i] = nw.Node
+		nodes[i] = nw.Id
 	}
 	return nodes
 }
 
 // Node is a struct that pairs a node with its weight in the signature aggregator.
```
samliok: this common is out of date now

simplex/util.go
```diff
@@ -54,7 +54,7 @@ func hasSomeNodeSignedTwice(nodeIDs []common.NodeID, logger common.Logger) (comm
 	return common.NodeID{}, false
 }
 
-func VerifyQC(qc common.QuorumCertificate, logger common.Logger, messageType string, isQuorum func(signers []common.NodeID) bool, eligibleSigners map[string]struct{}, messageToVerify verifiableMessage, from common.NodeID) error {
+func VerifyQC(qc common.QuorumCertificate, logger common.Logger, messageType string, isQuorum func(signers []common.NodeID) bool, eligibleSigners map[string]struct{}, messageToVerify verifiableMessage, from common.NodeID, nodes common.Nodes) error {
```
samliok, line 57: maybe for a separate pr we can remove the params `logger, messageType, and from` for `verifyQC`? We still return an error, so we can just log it at the call site.

samliok (approved):

simplex/util.go
yacovm: https://github.com/ava-labs/Simplex/pull/409

PR #411, opened 2026-06-15
random assortment of thoughts

samliok:


samliok:
closed because I think this is too disorganized to be useful

PR #419, opened 2026-06-22
Fix TestEpochRebroadcastsEmptyVoteAfterBlockProposalReceived Flake 

samliok:


simplex/epoch_failover_test.go
```diff
@@ -1189,6 +1189,12 @@ func waitForEmptyVote(t *testing.T, comm *rebroadcastComm, e *Epoch, expectedRou
 		select {
 		case emptyVote := <-comm.emptyVotes:
 			require.Equal(t, expectedRound, emptyVote.Vote.Round)
+			// A single time advance can queue more than one broadcast (the
+			// proposal-timeout vote plus the timeout-handler's rebroadcast on
+			// the same tick).
+			for len(comm.emptyVotes) > 0 {
```
yacovm: why not do:

```
if len(comm.emptyVotes) > 0 {
      continue
}
```

?

yacovm: In the `time.After` case, that is.

samliok: wouldn't empty votes be sent async so checking right after `AdvanceTime` might say len == 0, but then its added later?

yacovm (approved):

PR #420, opened 2026-06-22
Move msm metadata encoding to common

yacovm:


common/record.go
```diff
@@ -0,0 +1,296 @@
...
+	return Notarization{
+		Vote: vote,
+		QC:   qc,
+	}, nil
+}
+
+func BlockRecord(bh BlockHeader, blockData []byte) []byte {
+	mdBytes := bh.Bytes()
+
+	buff := make([]byte, len(mdBytes)+len(blockData)+2)
```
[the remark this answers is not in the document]
yacovm: I'm just moving code, not implementing this

msm/approvals_test.go
```diff
@@ -48,7 +49,7 @@ func TestApprovalStoreHandleApproval(t *testing.T) {
 			name:       "approval from unknown node is dropped",
 			validators: 3,
 			approvals: []approvalAndTimestamp{
-				{ValidatorSetApproval{NodeID: makeNodeID(99), PChainHeight: 1}, 1},
+				{common.ValidatorSetApproval{NodeID: makeNodeID(99), PChainHeight: 1}, 1},
```
samliok, line 52: `ValidatorSetApprovals` are only used in the msm code? if we move everything to the common repo aren't we back to where we started?

msm/block_type.go
```diff
@@ -29,7 +33,7 @@ func (bt BlockType) String() string {
 	}
 }
 
-func IdentifyBlockType(nextBlockMD StateMachineMetadata, prevBlockMD StateMachineMetadata, prevSeq uint64) BlockType {
```
samliok, line 32: same with the `StateMachineMetadata` doesn't it make sense to keep this in the msm?

samliok (commented):
i get moving some of the stuff, but for the MSM specific structs shouldn't we keep them in the MSM package?

msm/approvals_test.go
yacovm: Right now it's only used in the msm code, but eventually we'll need a way to send it over the wire, which means we might need to move it to be closer to `msg.go`

msm/block_type.go
yacovm: I thought that it might not be ideal that we have encodings of stuff spread out across the entire repo.

Having said that, I think I'll close this PR and see if it's possible to continue implementing the orchestration layer without moving the msm encoding to common.

PR #422, opened 2026-06-23
Use explicit PK instead of NodeID in Verify()

yacovm:
Since the epoch now has its own membership reference that includes public keys, it's easier to support epoch changes by altering the API accordingly.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -95,7 +95,7 @@ type Epoch struct {
 	blockBuilderCancelFunc         context.CancelFunc
 	nodeIDs                        common.NodeIDs
 	nodes                          common.Nodes
-	eligibleNodeIDs                map[string]struct{}
+	validatorsToPKs                map[string][]byte
 	rounds                         map[uint64]*Round
 	emptyVotes                     map[uint64]*EmptyVoteSet
 	oldestNotFinalizedNotarization NotarizationTime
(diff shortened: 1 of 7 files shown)


simplex/epoch.go
```diff
@@ -913,7 +917,13 @@ func (e *Epoch) handleEmptyVoteMessage(message *common.EmptyVote, from common.No
 
 	signature := message.Signature
 
-	if err := vote.Verify(signature.Value, e.Verifier, signature.Signer); err != nil {
+	pk, exists := e.nodesToPKs[string(signature.Signer)]
+	if !exists {
```
samliok, line 916: these checks should never happen right? we filter messages here and eligibleNodeIds should have the same exact key set as `nodeToPks`
```go
	_, known := e.eligibleNodeIDs[string(from)]
```

samliok: i think it would be better to combine these two maps(`nodeToPks` and `eligibleNodeIds`) into one? maybe `eligibleSigners`?

yacovm: > these checks should never happen right? we filter messages here and eligibleNodeIds should have the same exact key set as nodeToPks

but this is not `from`, this is who signed the message. We bind the `from` and the signer everywhere except in the signer of a block message (`VerifyBlockMessageVote`). 
Even if we add a check there, I think we should still do the lookup right before we verify the signature, otherwise the risk is just too great. I don't know how BLST will behave with an empty public key and I don't want to find out... 

However, you are right that if we have the mapping, we don't need the eligible nodes anymore.

simplex/epoch.go
```diff
@@ -152,7 +152,7 @@ func (e *Epoch) HandleMessage(msg *common.Message, from common.NodeID) error {
 	}
 
 	// Guard against receiving messages from unknown nodes
-	_, known := e.eligibleNodeIDs[string(from)]
```
samliok, line 155: can we make this map named something like `validatorsToPKs` or `signers`? it would be helpful to note these as different than the non-validator nodes, especially since we will need to update this guard in a pr i'm working on.

yacovm: done

samliok (approved):

PR #426, opened 2026-06-24
Rebroadcast request for latest epoch until we have reached enough responses

### LINKED ISSUE #425: [non-validators] Keep broadcasting request for latest epoch until we have reached a threshold

samliok:
In case the first broadcast message was not responded to by a threshold of validators, we want to continue rebroadcasting requests for the latest epoch until we validate an epoch. 

However it's possible that we restarted our non-validator and its already up-to-date. In that case, we limit the number of retries to `maxRebroadcastAttempts` as to not indefinitely re-broadcast.


nonvalidator/non_validator.go
```diff
@@ -571,3 +589,23 @@ func (n *NonValidator) sendRequest(seq uint64, to common.NodeID) {
 func (n *NonValidator) nextSeqToCommit() uint64 {
 	return n.Storage.NumBlocks()
 }
+
+func (n *NonValidator) bootstrapRunner(taskIds []uint64) {
+	if len(taskIds) != 1 {
```
yacovm, line 596: can this happen? If not, maybe log it as a fatal?

nonvalidator/non_validator.go
```diff
@@ -571,3 +589,23 @@ func (n *NonValidator) sendRequest(seq uint64, to common.NodeID) {
...
+		return
+	}
+
+	// drop the task we just processed; we reschedule the next attempt below.
+	n.bootstrapRebroadcastHandler.RemoveTask(taskIds[0])
+
+	attempt := taskIds[0] + 1
+
+	// too many attempts, don't rebroadcast
+	if attempt > maxRebroadcastAttempts {
```
yacovm, line 607: I don't understand why limit the number of retries... what happens if there is no connectivity and we have exhausted the number of retries?

nonvalidator/non_validator.go
```diff
@@ -571,3 +589,23 @@ func (n *NonValidator) sendRequest(seq uint64, to common.NodeID) {
...
+		return
+	}
+
+	n.Logger.Debug("Rebroadcasting latest epoch", zap.Uint64("Next to commit", n.nextSeqToCommit()))
+	n.broadcastLatestEpoch()
```
yacovm, line 612: shouldn't this method be called under a lock?

nonvalidator/non_validator.go
yacovm: we can remove this now, no?

nonvalidator/non_validator.go
samliok: > However it's possible that we restarted our non-validator and its already up-to-date. In that case, we limit the number of retries to maxRebroadcastAttempts as to not indefinitely re-broadcast.

yacovm: I am not sure what's the point in rebroadcasting it after the validator boots but only for 5 iterations. 

What is the problem that we are trying to solve?

nonvalidator/non_validator_test.go
```diff
@@ -652,6 +652,82 @@ func TestNonValidator_VerifiesFinalizationDuringReplication(t *testing.T) {
...
+		ID:                         myNodeID,
+		StartTime:                  start,
+	})
+	require.NoError(t, err)
+	defer nv.Stop()
+
+	// Seed the first attempt, then tick a few intervals past the cap. Each tick
+	// fires one rebroadcast and schedules the next attempt; ticks past the cap
+	// should be no-ops.
+	nv.bootstrapRebroadcastHandler.AddTask(0)
```
yacovm, line 680: we should not be accessing internal fields in tests in such a manner, but instead relying on exported APIs to drive the test.

nonvalidator/non_validator_test.go
```diff
@@ -652,6 +652,82 @@ func TestNonValidator_VerifiesFinalizationDuringReplication(t *testing.T) {
...
+		MaxSequenceWindow:          simplex.DefaultMaxRoundWindow,
+		ID:                         myNodeID,
+		StartTime:                  start,
+	})
+	require.NoError(t, err)
+	defer nv.Stop()
+
+	// Validating a new epoch should drop every pending rebroadcast task.
+	nv.bootstrapRebroadcastHandler.AddTask(0)
+	nv.maybeValidateNextEpoch(tc.appendSealing(testNodes))
```
yacovm, line 720: same, we should not be doing this.

PR #427, opened 2026-06-29
Simplex orchestration layer preliminary implementation

yacovm:
This commit contains an implementation of an orchestration layer that orchestrates
epoch transition for validators and non-validators as they move through epochs.

The orchestration layer introduces an Instance that drives the epoch lifecycle,
switching between validator consensus and non-validator block tracking as the
node's role changes across epochs.

Key pieces:

- Instance: manages the per-epoch lifecycle, ticks the active epoch or
    non-validator, and handles the handoff at epoch boundaries.
- epochChangeSupression: drops outgoing messages and cancels VM operations
    while an epoch change is in progress, preventing side effects mid-transition.
- EpochAwareStorage: ignores blocks from previous epochs and signals when a
    new epoch is detected.
- Config: wiring for the P-chain, storage, crypto, and networking dependencies.


instance.go
```diff
@@ -0,0 +1,300 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package simplex
```
samliok, line 4: should the orchastration be in its own package?

instance.go
```diff
@@ -0,0 +1,300 @@
...
+			}
+			return i.startEpoch(config)
+		},
+	}
+
+	epochConfig := simplex.EpochConfig{
+		// Parameter config
+		Epoch:                      epochNum,
+		ReplicationEnabled:         true,
+		StartTime:                  time.Now(),
```
samliok: should the orchestration layer have its own time field? i don't see where we forward/call `advanceTime`

instance.go
```diff
@@ -0,0 +1,300 @@
...
+		MaxProposalWait:            i.Config.MaxProposalWait,
+		MaxRebroadcastWait:         i.Config.MaxRebroadcastWait,
```
samliok: i think we should change in a seperate pr these to be consistent with avalanche go
```go
MaxNetworkDelay    time.Duration   `json:"maxNetworkDelay"    yaml:"maxNetworkDelay"`
MaxRebroadcastWait time.Duration   `json:"maxRebroadcastWait" yaml:"maxRebroadcastWait"`
```

instance.go
```diff
@@ -0,0 +1,300 @@
...
+
+	"github.com/ava-labs/simplex/common"
+	metadata "github.com/ava-labs/simplex/msm"
+	"github.com/ava-labs/simplex/simplex"
+	"github.com/ava-labs/simplex/wal"
+	"go.uber.org/zap"
+)
+
+
+type Config struct {
```
samliok, line 26: maybe this type of structure is more organized? this way we can see exactly what is used for the MSM vs epoch vs shared

```go
type Config struct {
      // Shared by both the metadata state machine and the epoch.
      Common CommonConfig
      // Used only to build the metadata state machine.
      MSM    MSMConfig
      // Used only to build the simplex epoch.
      Epoch  EpochParams
}

type CommonConfig struct {
      Logger                     common.Logger
      Signer                     common.Signer
      SignatureAggregatorCreator common.SignatureAggregatorCreator
      VM                         VM
}

type MSMConfig struct {
      GenesisValidatorSet         metadata.NodeBLSMappings
      LastNonSimplexInnerBlock    metadata.VMBlock
      GetPChainHeightForProposing func() uint64
      GetPChainHeightForVerifying func() uint64
      ComputeICMEpoch             metadata.ICMEpochTransition
}

type EpochParams struct {
      ID                         common.NodeID
      Sender                     Sender
      Storage                    Storage
      QCDeserializer             common.QCDeserializer
      BlockDeserializer          common.BlockDeserializer
      Verifier                   common.SignatureVerifier
      MaxProposalWait            time.Duration
      MaxRoundWindow             uint64
      MaxRebroadcastWait         time.Duration
      FinalizeRebroadcastTimeout time.Duration
      MaxWALSize                 int
      WALCreator                 wal.Creator
      WALs                       []wal.DeletableWAL
}
```

samliok: this could also help with the organization we talked about at Monday's consensus sync. The caller can see which interfaces it needs to implement and which purposes

instance.go
```diff
@@ -0,0 +1,300 @@
...
+	MaxRoundWindow             uint64
+	MaxRebroadcastWait         time.Duration
+	FinalizeRebroadcastTimeout time.Duration
+
+	ComputeICMEpoch metadata.ICMEpochTransition
+	// PChainProgressListener listens for changes in the P-chain height to trigger block building or h transitions.
+	PChainProgressListener metadata.PChainProgressListener
+	// LastNonSimplexBlockPChainHeight is the P-chain height of the last block built by a non-Simplex proposer.
+	// It is used to determine the validator set of the first ever Simplex h.
+	LastNonSimplexBlockPChainHeight uint64
```
samliok: dont see where this is uesd

instance.go
```diff
@@ -0,0 +1,300 @@
...
+type Config struct {
+	MaxWALSize int
+	MaxProposalWait            time.Duration
+	MaxRoundWindow             uint64
+	MaxRebroadcastWait         time.Duration
+	FinalizeRebroadcastTimeout time.Duration
+
+	ComputeICMEpoch metadata.ICMEpochTransition
+	// PChainProgressListener listens for changes in the P-chain height to trigger block building or h transitions.
+	PChainProgressListener metadata.PChainProgressListener
```
samliok: same, this isn't used anywhere

adapters.go
```diff
@@ -0,0 +1,194 @@
...
+	parsedBlock := &ParsedBlock{
+		msm: e.msm,
+		StateMachineBlock: block,
+	}
+	return parsedBlock, *finalization, nil
+}
+
+func (e *EpochAwareStorage) Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error {
+	if block.BlockHeader().Epoch < e.Epoch {
+		// This is a Telock from a previous h, so we ignore it and do not index it.
```
samliok: how does this mean the block is a telock? don't telocks have the same epoch as the sealing block?

yacovm: this is a Telock from the previous epoch. We may collect a finalization on a Telock and then we first index the sealing block and then the Telocks. Once we index the sealing block, we increment our epoch, right? 
We don't want to index the Telocks too, so this is a safeguard against that.

instance.go
yacovm: I think it makes sense to put this in the top most package, to make it clear that it is the entry point to simplex.

adapters.go
```diff
@@ -0,0 +1,318 @@
...
+	"github.com/ava-labs/simplex/common"
+	metadata "github.com/ava-labs/simplex/msm"
+)
+
+// epochChangeSupression is used to suppress sending messages during an epoch change or using the VM.
+// The motivation is that an epoch change occurrs during indexing a block, and we don't want any further
+// external side effects to take place before the epoch change finishes.
+// We therefore drop all messages and cancel VM operations such as block building,
+// and delay VM transaction listening until the epoch change is complete.
+type epochChangeSupression struct {
```
samliok:
```suggestion
type epochChangeSuppression struct {
```

similar typo throughout

adapters.go
```diff
@@ -0,0 +1,318 @@
...
+	return ecs.active
+}
+
+func (ecs *epochChangeSupression) sendProhibited(seq uint64) bool {
+	ecs.lock.RLock()
+	defer ecs.lock.RUnlock()
+	if !ecs.active {
+		return false
+	}
+	return seq > ecs.sealingBlockSeq
```
samliok: we prohibit the send if the seq > sealingBlockSeq. i also looked further in the code and it seems we only have this suppression for a certain amount of time until we transition the epoch

This means we will block any emptyVotes whose `round > sealingBlockSeq` (which is a weird artifact of having the `msg.Seq()` function. However, these empty votes may actually be part of the correct epoch since rounds are always greater or equal to sequences. So, we will drop these msgs instead of responding to them with `e.maybeSendNotarizationOrFinalization` which could be necessary for the lagging node in order to catch up. 

Now once we transition, we should be able to process these messages so the actual downside is dropping 1, maybe 2 empty vote messages from lagging nodes. That being said, I still think the epoch should be able to decide wether it drops or keeps messages, rather than the orchestrator.

adapters.go
```diff
@@ -0,0 +1,318 @@
...
+	c.Broadcaster.Broadcast(msg)
+}
+
+func (c *Communication) Send(msg *common.Message, destination common.NodeID) {
+	if c.epochChangeSupression.sendProhibited(msg.Seq()) {
+		return
+	}
+
+	c.Sender.Send(msg, destination)
+}
```
samliok: i think we should remove this comm adapter completely.

IMO the `epoch` should be the one filtering these messages themselves. I think it would simplify the orchestrator code a bit, remove some weirdness with the `msg.Seq` function, and make the epoch more explicit about how it handles future epoch messages.

samliok: plus if we add other messages to the epoch we need to remember to properly adjust the `msg.Seq` which is not that intuitive. Because whether a message gets sent to the epoch is actually controlled by `msg.Seq()` which is not an ideal place.

For example the block digest request literally has a seq field and you would expect msg.seq to return `BlockDigestRequest.seq` but instead it returns 0.

samliok: it might make sense creating the epoch.go changes in a separate pr to be merged before this one?

msm/build_decision.go
```diff
@@ -35,7 +35,7 @@ type blockBuildingDecider struct {
 	// hasValidatorSetChanged should return whether the validator set has changed since the
 	// P-chain height referenced by the last block in the chain and until the provided P-chain height.
 	hasValidatorSetChanged func(pChainHeight uint64) (bool, NodeBLSMappings, error)
-	getPChainHeight        func() uint64
+	getPChainHeight        func(ctx context.Context) (uint64, error)
```
samliok: are these changes important to the orchestrator? if not can we separate them?

msm/msm.go
```diff
@@ -194,6 +194,9 @@ type StateMachine struct {
 
 // Config contains the dependencies and configuration parameters needed to initialize the StateMachine.
 type Config struct {
+	// SkipMSMVerification indicates whether to skip the verification of the state machine transition when verifying blocks,
+	// and only verify the inner block. This is useful when replicating finalized blocks as a non-validator.
+	SkipMSMVerification bool
```
samliok, line 199: hmm, wouldn't this mean a non-validator can accept and index a block that has the wrong MSM metadata? 

this can be a problem if the non-validator later becomes a validator. I.e. lets say we index a sealing block with the wrong msm data. we then become a validator due to the sealing block. Now our verification/block building as a validator will be all messed up since we will fetch an incorrectly indexed block

msm/util_test.go
```diff
@@ -330,21 +309,32 @@ func makeNonSimplexBlock(t *testing.T, startHeight uint64, start time.Time, h ui
...
 	signatureVerifier     signatureVerifier
 	signatureAggregator   signatureAggregator
-	blockBuilder          blockBuilder
+	blockBuilder          mockBlockBuilder
 	keyAggregator         keyAggregator
 	validatorSetRetriever validatorSetRetriever
 }
 
+// mockBlockBuilder is a test BlockBuilder whose BuildBlock returns a preconfigured block/error.
+type mockBlockBuilder struct {
```
samliok: i think the use of the word mock will be triggering to some members of the avalanchego team

samliok (commented):
leaving what i have, was not able to review as much as i hoped today

msm/msm.go
yacovm: > hmm, wouldn't this mean a non-validator can accept and index a block that has the wrong MSM metadata?

But the MSM metadata is signed by a QC no? How can it accept a block with the wrong metadata?

adapters.go
yacovm: https://github.com/ava-labs/Simplex/pull/433

external.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (p *ParsedBlock) Blacklist() common.Blacklist {
+	var blacklist common.Blacklist
+	_ = blacklist.FromBytes(p.Metadata.SimplexBlacklist) // TODO: encode blacklist with Canoto
+	return blacklist
+}
+
+func (p *ParsedBlock) Verify(ctx context.Context) (common.VerifiedBlock, error) {
+	if err := p.msm.VerifyBlock(ctx, &p.StateMachineBlock); err != nil {
```
samliok, line 67: alternatively we could do something like 

```go
if msm.IsValidator() {
msm.verifyBlock
} else {
p.Inner.VerifyBlock()
}
```

adapters.go
yacovm: These messages are dropped only after the epoch is sealed, right? So it doesn't matter, they are pointless as everything that happened after the epoch is sealed is rewinded backwards

msm/util_test.go
```diff
@@ -66,6 +70,10 @@ type fakeVMBlock struct {
 	height uint64
 }
 
+func (f *fakeVMBlock) Bytes() ([]byte, error) {
+	panic("implement me")
```
samliok, line 74: is this a TODO for the current pr?

nonvalidator/non_validator.go
```diff
@@ -16,6 +16,10 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	DefaultMaxSequenceWindow = 100
```
samliok: i think this should be a similar value to DefaultMaxRoundWindow, if not the same. Mainly because validators will reject replication that have more seqs than MaxRoundWindow.

BTW, I think we should change how we process replication requests, to not drop messages that request too many sequences but rather fill up the response with at most `MaxRoundWindow`. Also, we should ensure this message does not exceed the maximum number of bytes `AvalancheGo` clients can accept.

msm/build_decision.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/changes/277eb0a65d85c49f0fd0b4b401b5d24e5e44f33c#diff-6e7f518f13f3f50e665a925ee4fa645f1ad8d2a9c222dbd80950cb08f29e7d65

instance.go
```diff
@@ -0,0 +1,620 @@
...
+
+	lastBlock, _, err := i.Config.Storage.GetBlock(numBlocks - 1)
+	if err != nil {
+		return metadata.StateMachineBlock{}, 0, fmt.Errorf("error retrieving last block from storage: %w", err)
+	}
+
+	return lastBlock, numBlocks, nil
+}
+
+func (i *Instance) determineValidatorOrNot(nodes common.Nodes) bool {
```
samliok: thoughts on changing this to `nodes.Contains(myNodeId)`? 

or making the function header a bit more clear


```suggestion
func (i *Instance) isMyNodeValidator(validators common.Nodes) bool {
```

instance.go
```diff
@@ -0,0 +1,620 @@
...
+	Broadcaster Broadcaster
+	WALs        []wal.DeletableWAL
+	VM          VM
+	ID          common.NodeID
+}
+
+type MessageHandler interface {
+}
+
+type nodeRole byte
```
samliok, line 49: this nodeRole variable is confusing. It represents the previous role of the node right? Why do we need to know this?

instance.go
```diff
@@ -0,0 +1,620 @@
...
+	}
+
+	if config.Epoch != epochChange.epochNum {
+		i.Config.Logger.Error("Epoch number mismatch on epoch change", zap.Uint64("expected", epochChange.epochNum), zap.Uint64("actual", config.Epoch))
+		return
+	}
+
+	if err := i.startEpoch(config); err != nil {
+		i.Config.Logger.Error("Error starting new epoch on epoch change", zap.Error(err))
+	}
```
samliok: i think we should have one unified way of starting a validator. We have this logic here, but we have a different method `startValidator`

instance.go
```diff
@@ -0,0 +1,620 @@
...
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	// Stop the non-validator before doing anything else, so that we don't process any more messages while we are changing epochs.
+	i.stopNonValidator()
+
+	// First, figure out if I'm still a validator.
+	if i.determineValidatorOrNot(epochChange.validators) {
+		i.Config.Logger.Info("I am now a validator")
+		i.startValidator()
```
samliok: we are dropping the error here, shouldn't this be propagated up? Otherwise, we will just be dropping messages not knowing why we failed starting

instance.go
```diff
@@ -0,0 +1,620 @@
...
+	defer i.lock.Unlock()
+
+	i.stopCh = make(chan struct{})
+	i.epochChanges = make(chan epochChange)
+	context.AfterFunc(ctx, i.Stop)
+
+	cachedStorage := NewCachedStorage(i.Config.Storage)
+	i.cs = cachedStorage
+
+	lastBlock, numBlocks, err := i.lastBlock()
```
samliok, line 95: missing error check

instance.go
samliok: maybe something like this works? 
https://github.com/ava-labs/Simplex/pull/434

adapters.go
```diff
@@ -0,0 +1,252 @@
...
+
+// EpochAwareStorage is a wrapper around Storage that is aware of epoch changes.
+// Upon an epoch change, it will ignore blocks from previous epochs
+// and will call the OnEpochChange callback when a new epoch is detected.
+type EpochAwareStorage struct {
+	msm           *metadata.StateMachine
+	OnEpochChange func(seq uint64, validators common.Nodes) error
+	Storage
+	Epoch uint64
+}
```
samliok, line 42: why are some of these fields exported while others aren't?

config.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// ComputeICMEpoch computes the ICM epoch transition given the input parameters.
+	ComputeICMEpoch(input metadata.ICMEpochInput) metadata.ICMEpochInfo
+}
+
+type Storage interface {
+	// GetBlock retrieves the block and finalization at [seq] with the given digest.
+	// If the digest is nil, the block with the given sequence number is returned.
+	// If [seq] the block cannot be found, returns ErrBlockNotFound.
+	GetBlock(seq uint64) (metadata.StateMachineBlock, *common.Finalization, error)
```
samliok, line 67: can we use the same terminology as simplex.Storage? maybe we can embed simplex.Storage in this API?

```go
type Storage interface {
simplex.Storage

// CreateWAL creates a new Write-Ahead Log (WAL).
	CreateWAL() (wal.DeletableWAL, error)
```

samliok: why do couple the WAL with the storage anyways?

external.go
yacovm: But what we have right now is just checking a single boolean. 

Checking if the node is a validator involves a linear search if it's a slice, or a map lookup which is O(1) amortized but is much heavier than a boolean check.

msm/util_test.go
yacovm: No, it's just so that the code will compile. It's just used for some test but this is never invoked.

instance.go
```diff
@@ -0,0 +1,620 @@
...
+	}
+	return nil
+}
+
+func (i *Instance) transitionEpochNonValidator(epochChange epochChange) {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	// Stop the non-validator before doing anything else, so that we don't process any more messages while we are changing epochs.
+	i.stopNonValidator()
```
samliok, line 546: i dont get why we need to stop and restart a non-validator for every epoch change. This actually seems like a problem during a non-validators bootstrapping phase.

Non-validators backwards hash-chain validate epochs. This process involves storing requested future sealing blocks in memory and then indexing in order. However, when we index a sealing block, we would restart the non-validator and therefore clear its cached memory of sealing blocks meaning it needs to restart the requesting phase. 

The non-validator is epoch agnostic so I think if we transition from non-validator -> non-validator all we need to do is update the communication interface to return the latest validator set.

samliok: even further, if we go from non-validator -> validator, and we are aware of a future epoch we should still be in the non-validating phase?

config.go
```diff
@@ -0,0 +1,107 @@
...
+	// Index indexes the given block and finalization in the storage.
+	Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error
+
+	// CreateWAL creates a new Write-Ahead Log (WAL).
+	CreateWAL() (wal.DeletableWAL, error)
+}
+
+type CryptoOps interface {
+	Sign(message []byte) ([]byte, error)
+	AggregateKeys(keys ...[]byte) ([]byte, error)
```
samliok, line 78: instead of redefining the functions should we just inline the interfaces from msm
```suggestion
	metadata.KeyAggregator
```

same with the others
```go
	common.Signer
	common.SignatureVerifier
	common.QCDeserializer
	metadata.KeyAggregator
```

config.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/ava-labs/simplex/common"
+	metadata "github.com/ava-labs/simplex/msm"
+	"github.com/ava-labs/simplex/wal"
+)
+
+type ParameterConfig struct {
+	// WalMaxEntryCount is the maximum number of entries in the write-ahead log before it is closed.
+	WALMaxEntryCount int
+	// MaxnetworkDelay is the assumed upper bound on the network delay for messages to be delivered.
```
samliok:
```suggestion
	// MaxNetworkDelay is the assumed upper bound on the network delay for messages to be delivered.
```

config.go
yacovm: > can we use the same terminology as simplex.Storage? maybe we can embed simplex.Storage in this API?

We can't, `Retrieve` conflicts because the parameters are different. 

```
./instance.go:468:12: cannot use i.cs (variable of type *CachedStorage) as Storage value in struct literal: *CachedStorage does not implement Storage (wrong type for method Retrieve)
		have Retrieve(uint64, [32]byte) (common.VerifiedBlock, *common.Finalization, error)
		want Retrieve(uint64) (common.VerifiedBlock, common.Finalization, error)
```

> why do couple the WAL with the storage anyways?

Yeah, so i was trying to minimize the number of dependencies. I was too frugal to dedicate an entire interface for a single method. I can do it though

nonvalidator/non_validator.go
yacovm: Yeah good catch, I removed and changed it [here](https://github.com/ava-labs/Simplex/pull/427/changes#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR183).

instance.go
yacovm: Changed it to `iAmValidator` before seeing this comment, I think it's equivalent.

instance.go
yacovm: We need to know this in order to know which node type to shutdown and whether to skip shutting down a non-validator in case it is still a non validator in the next epoch.

I changed the code significantly before looking at this suggestion so please let me know if you think it's still relevant.

instance.go
yacovm: Agreed, I changed accordingly.

instance.go
yacovm: whoops, good catch. Made us [halting](https://github.com/ava-labs/Simplex/pull/427/changes#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR345-R359) the node entirely if we fail.

instance.go
yacovm: fixed, thanks

adapters.go
yacovm: Fixed.

config.go
samliok: yea i think it makes sense to separate it

yacovm: https://github.com/ava-labs/Simplex/pull/427/changes#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR39

instance.go
yacovm: Yeah, all very good callouts 👍 

I was aware that the current approach isn't ideal but was in a hurry to meet the end of month deadline.

Take a look at the latest code version in the meantime. I will add a test a bit later that will check that we don't needlessly restart non-validators.

yacovm: https://github.com/ava-labs/Simplex/pull/427/changes#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR153-R165

yacovm: https://github.com/ava-labs/Simplex/pull/427/changes#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR523-R535

common/msg.go
```diff
@@ -40,6 +40,31 @@ func (m *Message) IsReplicationMessage() bool {
 	}
 }
 
+func (m *Message) Seq() uint64 {
```
samliok: wait why is this added back?

yacovm: I rebased and out of habit added it back

config.go
yacovm: 🦅 👁️

config.go
yacovm: I disagree. All these methods are single method ones. 

What's the benefit of embedding an interface if it doesn't reduce the volume of the interface? 

If anything, this explicit definition makes it much more clear what the integrator needs to implement.

msm/build_decision.go
```diff
@@ -72,7 +72,8 @@ func (bbd *blockBuildingDecider) shouldBuildBlock(
 
 		// If the P-chain height changed, re-evaluate again whether we should transition to a new epoch,
 		// or continue waiting to build a block.
-		if bbd.getPChainHeight() != pChainHeight {
+		h := bbd.getPChainHeight()
```
samliok: any reason for this diff?

instance.go
```diff
@@ -0,0 +1,627 @@
...
+	// Storage is the interface to the block storage layer for the simplex instance.
+	Storage Storage
+	Logger  common.Logger
+	Sender  Sender
+	WALs    []wal.DeletableWAL
+	VM      VM
+	ID      common.NodeID
+}
+
+type MessageHandler interface {
```
samliok: can be removed?

adapters.go
```diff
@@ -0,0 +1,252 @@
...
+	parsedBlock := &ParsedBlock{
+		msm:               e.msm,
+		StateMachineBlock: block,
+	}
+	return parsedBlock, *finalization, nil
+}
+
+func (e *EpochAwareStorage) Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error {
+	if block.BlockHeader().Epoch < e.epoch {
+		// This is a Telock from a previous h, so we ignore it and do not index it.
```
samliok: telock from a previous height? or previous epoch?

adapters.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil
+	}
+	if err := e.Storage.Index(ctx, block, certificate); err != nil {
+		return err
+	}
+	if block.SealingBlockInfo() != nil {
+		if err := e.onEpochChange(block.BlockHeader().Seq, block.SealingBlockInfo().ValidatorSet); err != nil {
+			return err
+		}
+		// We are now in a new h, so we update the h number to prevent indexing Telocks from the previous h.
```
samliok: can we use the term epoch instead of `h`? unless `h` is supposed to represent height, then can we rename to `height`

adapters.go
```diff
@@ -0,0 +1,252 @@
...
+		cb.cache.insertBlock(cb.ParsedBlock)
+	}
+	return vb, err
+}
+
+type CachedStorage struct {
+	msm  *metadata.StateMachine
+	lock sync.RWMutex
+	Storage
+	cache map[[32]byte]cachedBlock
```
samliok: i think we were planning on using simplex.digests instead of [32]byte? maybe for a follow up pr?

adapters.go
```diff
@@ -0,0 +1,252 @@
...
+	msm *metadata.StateMachine
+}
+
+func (bp *blockDeserializer) DeserializeBlock(ctx context.Context, bytes []byte) (common.Block, error) {
+	var rawBlock RawBlock
+	if err := rawBlock.UnmarshalCanoto(bytes); err != nil {
+		return nil, err
+	}
+
+	block, err := bp.vm.ParseBlock(ctx, rawBlock.InnerBlockBytes)
```
samliok, line 242: isn't this inner block technically still a simplex.Block? It feels weird to ask the vm to parse the simlpex.Block, as the vm should not know about simplex blocks.

instance.go
```diff
@@ -0,0 +1,627 @@
...
+		return nil
+	}
+
+	// Stop the non-validator before doing anything else, so that we don't process any more messages while we are changing epochs.
+	i.stopNonValidator()
+
+	return i.startValidatorOrNonValidator(epochChange.validators, epochChange.epochNum)
+}
+
+func (i *Instance) startValidatorOrNonValidator(validators common.Nodes, epoch uint64) error {
```
samliok: what about calling it `startInstanceAtEpoch`?

instance.go
```diff
@@ -0,0 +1,627 @@
...
+	lastBlock, numBlocks, err := i.lastBlock()
+
+	lastNonSimplexHeight := i.Config.LastNonSimplexInnerBlock.Height()
+	genesisValidatorSet := i.Config.PlatformChain.GenesisValidatorSet()
+	nodes, epochNum, err := constructEpochAndValidatorSet(lastNonSimplexHeight, genesisValidatorSet, numBlocks, ParsedBlock{StateMachineBlock: lastBlock}, i.Config.Storage)
+	if err != nil {
+		return fmt.Errorf("error determining latest epoch and validator set: %w", err)
+	}
+
+	i.startValidatorOrNonValidator(nodes, epochNum)
```
samliok: dropping the err returned

instance.go
samliok: hmm, don't see the fix

instance.go
```diff
@@ -0,0 +1,627 @@
...
+	wal          *wal.GarbageCollectedWAL
+	msm          *metadata.StateMachine
+	e            *simplex.Epoch
+	nv           *nonvalidator.NonValidator
+	epochOrNV    timeAdvancer
+	epochChanges chan epochChange
+	stopCh       chan struct{}
+}
+
+func (i *Instance) Start(ctx context.Context) error {
```
samliok, line 88: should we keep this consistent with the other structs and have a constructor and then a start function? 

```go
// New creates an Instance from the given config. 
func New(config Config) *Instance {
	return &Instance{
		Config:       config,
		cs:           NewCachedStorage(config.Storage),
		stopCh:       make(chan struct{}),
		epochChanges: make(chan epochChange, 1),
	}
}
```

instance.go
```diff
@@ -0,0 +1,627 @@
...
+	case validator:
+		if err := i.transitionEpochValidator(epochChange); err != nil {
+			i.Config.Logger.Error("Error transitioning epoch for validator", zap.Error(err))
+			i.Stop()
+		}
+	default: // This should never happen, but we log it just in case.
+		i.Config.Logger.Fatal("Unknown node role on epoch change",
+			zap.String("role", fmt.Sprintf("%v", epochChange.nodeRole)))
+	}
+}
```
samliok, line 372: thoughts on this nit? 
```suggestion
func (i *Instance) processEpochChange(epochChange epochChange) {
	var err error
	switch epochChange.nodeRole {
	case nonValidator:
		err = i.transitionEpochNonValidator(epochChange)
	case validator:
		err = i.transitionEpochValidator(epochChange)
	default: // This should never happen, but we log it just in case.
		i.Config.Logger.Fatal("Unknown node role on epoch change",
			zap.String("role", fmt.Sprintf("%v", epochChange.nodeRole)))
		return
	}

	if err != nil {
		i.Config.Logger.Error("Error transitioning epoch", zap.Uint8("role", uint8(epochChange.nodeRole)), zap.Error(err))
		i.Stop()
	}
}
```

instance.go
```diff
@@ -0,0 +1,627 @@
...
+
+	switch {
+	case msg.BlockMessage != nil:
+		err := i.wireBlockMessage(msg)
+		if err != nil {
+			i.Config.Logger.Debug("Error wiring block message", zap.Error(err))
+			return nil
+		}
+	case msg.ReplicationResponse != nil:
+		err := i.wireReplicationResponse(msg)
```
samliok, line 277: can we add some comments to why we are doing this? I'll probably forget in 3 months and then have to re-answer this question. 

I assume its because we need to wrap the blocks to msm blocks right?

instance.go
```diff
@@ -0,0 +1,627 @@
...
+	if i.nv != nil {
+		return i.nv.HandleMessage(msg, from)
+	}
+	return nil
+}
+
+func (i *Instance) wireReplicationResponse(msg *common.Message) error {
+	resp := msg.ReplicationResponse
+	if resp.LatestRound != nil && resp.LatestRound.Block != nil {
+		pb, isParsedBlock := resp.LatestRound.Block.(*ParsedBlock)
```
samliok: can we separate this into a helper and use throughout?

```go
// cacheBlock wires the current MSM into a block received from the network,
// and wraps it so that reads go through the block cache.
func (i *Instance) cacheBlock(block common.Block) (common.Block, error) {
	pb, isParsedBlock := block.(*ParsedBlock)
	if !isParsedBlock {
		return nil, fmt.Errorf("expected ParsedBlock, got %T", block)
	}
	pb.msm = i.msm
	return &cachedBlock{
		cache:       i.cs,
		ParsedBlock: pb,
	}, nil
}
```

instance.go
```diff
@@ -0,0 +1,627 @@
...
+
+	select {
+	case <-i.stopCh:
+		// Already stopped, do nothing
+		return
+	default:
+		close(i.stopCh)
+	}
+
+	if i.e != nil {
```
samliok: unnecessary since this is checked in stopValidator. Or is this defensive?

instance.go
```diff
@@ -0,0 +1,627 @@
...
+		i.stopValidator()
```
samliok: why don't we potentially stop the non-validator as well?

instance.go
```diff
@@ -0,0 +1,627 @@
...
+			return fmt.Errorf("expected ParsedBlock, got %T", datum.Block)
+		}
+		resp.Data[j].Block = &cachedBlock{
+			cache:       i.cs,
+			ParsedBlock: pb,
+		}
+		pb.msm = i.msm
+	}
+	return nil
+}
```
samliok, line 321: simplification if we implement the shared helper
```suggestion
func (i *Instance) wireReplicationResponse(msg *common.Message) error {
	resp := msg.ReplicationResponse
	quorumRounds := []*common.QuorumRound{resp.LatestRound, resp.LatestSeq}
	for j := range resp.Data {
		quorumRounds = append(quorumRounds, &resp.Data[j])
	}

	for _, qr := range quorumRounds {
		if qr == nil || qr.Block == nil {
			continue
		}
		block, err := i.cacheBlock(qr.Block)
		if err != nil {
			return err
		}
		qr.Block = block
	}
	return nil
}
```

instance.go
```diff
@@ -0,0 +1,627 @@
...
+
+func (i *Instance) createEpochConfig() (simplex.EpochConfig, error) {
+	lastBlock, numBlocks, err := i.lastBlock()
+	if err != nil {
+		return simplex.EpochConfig{}, err
+	}
+
+	lastNonSimplexHeight := i.Config.LastNonSimplexInnerBlock.Height()
+	genesisValidatorSet := i.Config.PlatformChain.GenesisValidatorSet()
+	nodes, epochNum, err := constructEpochAndValidatorSet(lastNonSimplexHeight, genesisValidatorSet, numBlocks, ParsedBlock{StateMachineBlock: lastBlock}, i.Config.Storage)
```
samliok: nit: this is duplicated in the Start function as well, might make sense to create another shared helper?


```go

// currentEpochState reads the last persisted block and derives the epoch
// number and validator set the chain is currently in.
func (i *Instance) currentEpochState() (metadata.StateMachineBlock, common.Nodes, uint64, error) {
	lastBlock, numBlocks, err := i.lastBlock()
	if err != nil {
		return metadata.StateMachineBlock{}, nil, 0, err
	}

	lastNonSimplexHeight := i.Config.LastNonSimplexInnerBlock.Height()
	genesisValidatorSet := i.Config.PlatformChain.GenesisValidatorSet()
	nodes, epochNum, err := constructEpochAndValidatorSet(lastNonSimplexHeight, genesisValidatorSet, numBlocks, ParsedBlock{StateMachineBlock: lastBlock}, i.Config.Storage)
	if err != nil {
		return metadata.StateMachineBlock{}, nil, 0, fmt.Errorf("error determining latest epoch and validator set: %w", err)
	}

	return lastBlock, nodes, epochNum, nil
}
```

instance.go
```diff
@@ -0,0 +1,627 @@
...
+		if sealingBlock.Metadata.SimplexEpochInfo.BlockValidationDescriptor == nil {
+			return nil, 0, fmt.Errorf("expected sealing block at seq %d, but got a non-sealing block", sealingBlockSeq)
+		}
+		validatorSet = constructValidatorSetFromSealingBlock(ParsedBlock{StateMachineBlock: sealingBlock})
+		nodes = validatorSetToNodes(validatorSet)
+	}
+	return nodes, epochNum, nil
+}
+
+func validatorSetToNodes(genesisValidatorSet metadata.NodeBLSMappings) common.Nodes {
```
samliok:
```suggestion
func validatorSetToNodes(validatorSet metadata.NodeBLSMappings) common.Nodes {
```

instance.go
```diff
@@ -0,0 +1,627 @@
...
+	epochConfig, err := i.createEpochConfig()
+	if err != nil {
+		return err
+	}
+
+	if err := i.startEpoch(epochConfig); err != nil {
+		return err
+	}
+
+	return nil
```
samliok:
```suggestion
	return i.startEpoch(epochConfig)
```

instance.go
```diff
@@ -0,0 +1,627 @@
...
+	i.Config.WALs = nil
+	// On epoch change, garbage collect the WAL to remove all entries from previous epochs.
+	if err := i.wal.GarbageCollect(math.MaxUint64); err != nil {
+		i.Config.Logger.Error("Error garbage collecting epoch config on epoch change", zap.Error(err))
+	}
+
+	return i.startValidatorOrNonValidator(epochChange.validators, epochChange.epochNum)
+}
+
+func constructEpochAndValidatorSet(lastNonSimplexInnerBlockHeight uint64, genesisValidatorSet metadata.NodeBLSMappings, numBlocks uint64, lastBlock ParsedBlock, storage Storage) (common.Nodes, uint64, error) {
```
samliok: since this is an isolated function, should we have claude write some tests that we review for it?

alternatively since this is only used once should we just update the caller? I told claude to do this and it seemed to have made some good cleanups


```go
// currentEpochState reads the last persisted block and derives the epoch
// number and validator set the chain is currently in.
func (i *Instance) currentEpochState() (metadata.StateMachineBlock, common.Nodes, uint64, error) {
	lastBlock, numBlocks, err := i.lastBlock()
	if err != nil {
		return metadata.StateMachineBlock{}, nil, 0, err
	}

	parsed := ParsedBlock{StateMachineBlock: lastBlock}
	sealingInfo := parsed.SealingBlockInfo()

	switch {
	// If all we have in the ledger is non-Simplex blocks, load the validator set from genesis.
	case i.Config.LastNonSimplexInnerBlock.Height()+1 == numBlocks:
		genesisValidatorSet := i.Config.PlatformChain.GenesisValidatorSet()
		return lastBlock, validatorSetToNodes(genesisValidatorSet), i.Config.LastNonSimplexInnerBlock.Height() + 1, nil
	// If the last block persisted is a sealing block, then we are in the next epoch,
	// which is identified by the sealing block's sequence and carries its validator set.
	case sealingInfo != nil:
		return lastBlock, sealingInfo.ValidatorSet, parsed.BlockHeader().Seq, nil
	// Else, we have at least one Simplex block in the ledger, and it's not a sealing block.
	// Its Epoch field is the sequence of the sealing block that started this epoch.
	default:
		sealingBlockSeq := parsed.BlockHeader().Epoch
		sealingBlock, _, err := i.Config.Storage.GetBlock(sealingBlockSeq)
		if err != nil {
			return metadata.StateMachineBlock{}, nil, 0, fmt.Errorf("error retrieving sealing block from storage: %w", err)
		}
		info := (&ParsedBlock{StateMachineBlock: sealingBlock}).SealingBlockInfo()
		if info == nil {
			return metadata.StateMachineBlock{}, nil, 0, fmt.Errorf("expected sealing block at seq %d, but got a non-sealing block", sealingBlockSeq)
		}
		return lastBlock, info.ValidatorSet, sealingBlockSeq, nil
	}
}
```

msm/build_decision.go
yacovm: Once we introduce a context and an error return value, we will need to do this, so the diff will be smaller then.

adapters.go
yacovm: epoch

adapters.go
yacovm: I don't know why Goland refactored epoch to `h` 🤦‍♂️

adapters.go
yacovm: No, look at the definition of `RawBlock`:

```
type RawBlock struct {
	Metadata        metadata.StateMachineMetadata `canoto:"value,1"`
	InnerBlockBytes []byte                        `canoto:"bytes,2"`

	canotoData canotoData_RawBlock `canoto:"nocopy"`
}
```

The inner block is the VM block.


`DeserializeBlock` takes a raw block and then returns a `common.Block`:

When we look at at `common.Block`, it has all the stuff Simplex needs: 

```
type Block interface {
	// BlockHeader encodes a succinct and collision-free representation of a block.
	BlockHeader() BlockHeader

	Blacklist() Blacklist

	// Verify verifies the block by speculatively executing it on top of its ancestor.
	Verify(ctx context.Context) (VerifiedBlock, error)

	// non nil only for sealing blocks & first ever simplex block
	SealingBlockInfo() *SealingBlockInfo
}
```

instance.go
yacovm: we need to wire the cache and MSM, yes.

instance.go
yacovm: I am not sure it's worth it, because we essentially add a new function that returns a bunch of parameters and then we lose their context/purpose, and we don't actually save a lot in the number of code lines saves as it's only duplicated once.

instance.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/changes/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdL49-L51

adapters.go
yacovm: I don't know why Goland decided to make this refactor. I think my Mac Goland shortcuts are somehow messed up, it ended up also renaming comments. 

https://github.com/ava-labs/Simplex/pull/427/changes/7006400c61a07773e3c57293712bf802a5db20ff#diff-d35ab1faa89afd82022e8f5c63a36111a059ecf21f54fbbccc767da9b2a62283L56-L68

adapters.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/changes/7006400c61a07773e3c57293712bf802a5db20ff#diff-d35ab1faa89afd82022e8f5c63a36111a059ecf21f54fbbccc767da9b2a62283R68

adapters.go
yacovm: Changed accordingly

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR77-R107

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR107

instance.go
yacovm: whoops, https://github.com/ava-labs/Simplex/pull/427/changes/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR96-R98

adapters.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/changes/f33677252192aa94d19debb3e7feee538b9592b3#diff-d35ab1faa89afd82022e8f5c63a36111a059ecf21f54fbbccc767da9b2a62283R39-R41

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR79-R87

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR347-R361

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR258-R326

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdL233-L235

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdL233-L235

instance.go
yacovm: I understand the refactoring, Claude also recommended to me something like this, but I would rather call `wireBlock` explicitly instead of defining a slice and then continuously re-allocate it. 

It also reads weird to me that we build a slice composed of two independent fields and another slice, just to assign values. 

I am also a bit paranoid that `qr.Block` will not actually have any effect, what if `qr` is a copy-by-value? This reads pretty brittle to me.

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR603

instance.go
yacovm: https://github.com/ava-labs/Simplex/commit/7006400c61a07773e3c57293712bf802a5db20ff#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdL116-L121

instance.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/commits/9617211e33a30353846de1060562389892569552

external.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, vdr := range bdc.AggregatedMembership.Members {
+		nodes = append(nodes, common.Node{
+			Id:     vdr.NodeID[:],
+			Weight: vdr.Weight,
+			PK:     vdr.BLSKey,
+		})
+	}
+
+	return &common.SealingBlockInfo{
+		ValidatorSet: nodes,
```
samliok: what about setting the previous validator set?

`PrevSealingBlockHash Digest`

common/msg.go
```diff
@@ -136,7 +136,7 @@ func verifyContext(signature []byte, verifier SignatureVerifier, msg []byte, con
 	if err != nil {
 		return err
 	}
-	return verifier.Verify(toBeSigned, signature, pk)
+	return verifier.VerifySignature(toBeSigned, signature, pk)
```
samliok, line 139: so we have a lot of different declarations of verify signature.

```go
// api.go
type SignatureVerifier interface {
	VerifySignature(message []byte, signature []byte, publicKey []byte) error
}
```

```go
// config.Go <- crypto ops interface
VerifySignature(message []byte, signature []byte, publicKey []byte) error
```

```go
//msm.go
type SignatureVerifier interface {
	VerifySignature(signature []byte, message []byte, publicKey []byte) error
}
```

does it make sense to consolidate this into one? i think there may be a current bug in our test code since we use the same verifier to implement all 3 interfaces, even though the function signatures are different.

adapters.go
```diff
@@ -0,0 +1,252 @@
...
+
+func (e *EpochAwareStorage) Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error {
+	if block.BlockHeader().Epoch < e.epoch {
+		// This is a Telock from a previous epoch, so we ignore it and do not index it.
+		return nil
+	}
+	if err := e.Storage.Index(ctx, block, certificate); err != nil {
+		return err
+	}
+	if block.SealingBlockInfo() != nil {
```
samliok: we need to check if this is not the zero block. otherwise we wipe the wal mid epoch(when we index the first simplex block), and if we crash we may not be able to recover.

external.go
```diff
@@ -0,0 +1,92 @@
...
+	return rawBlock.MarshalCanoto(), nil
+}
+
+func (p *ParsedBlock) BlockHeader() common.BlockHeader {
+	var md *common.ProtocolMetadata
+	var err error
+	if len(p.Metadata.SimplexProtocolMetadata) > 0 {
+		md, err = common.ProtocolMetadataFromBytes(p.Metadata.SimplexProtocolMetadata)
+		if err != nil {
+			panic(err) // TODO: handle error
```
samliok, line 47: i think this will fail some tests in avalanchego since the panic is not caught anywhere. for the todo, are we going to add an error to the function return? 

I feel like that would be kinda gross

instance.go
```diff
@@ -0,0 +1,632 @@
...
+		return nil
+	}
+
+	// Stop the non-validator before doing anything else, so that we don't process any more messages while we are changing epochs.
+	i.stopNonValidator()
+
+	return i.startAtEpoch(epochChange.validators, epochChange.epochNum)
+}
+
+func (i *Instance) startAtEpoch(validators common.Nodes, epoch uint64) error {
```
samliok, line 551: should we check if the instance is stopped before this? otherwise we could have a race where 

1. we get notified on the channel an epoch transition is necessary
2. someone calls stop()
3. we call startAtEpoch() given the channel is stopped

therefore we will never be able to stop() the epoch or non-validator created during this race

adapters.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/commits/ea315d67254d56d770cc4a1fc35cdbf59b75a8e2

external.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/commits/ea315d67254d56d770cc4a1fc35cdbf59b75a8e2

common/msg.go
yacovm: Probably, but the real reason we have this bug is because we don't use types and use []byte everywhere.

Can we do it in a follow-up PR though?

external.go
yacovm: After https://github.com/ava-labs/Simplex/pull/435 will be merged, we won't have this error.

instance.go
yacovm: https://github.com/ava-labs/Simplex/pull/427/commits/3f09ad3f42e6158148bd4ca1e462840e367cb08a

samliok (approved):

instance.go
```diff
@@ -0,0 +1,646 @@
...
+	if err != nil {
+		return nonvalidator.Config{}, err
+	}
+
+	comm := &Communication{Sender: i.Config.Sender, Broadcaster: i.Config.Broadcaster}
+	comm.SetValidators(validators)
+
+	epochAwareStorage := &EpochAwareStorage{
+		epoch:   epochNum,
+		Storage: i.Config.Storage,
```
samliok, line 152: is there a reason we are not using the cached storage?

instance.go
```diff
@@ -0,0 +1,646 @@
...
+	Config       Config
+	lock         sync.Mutex
+	cs           *CachedStorage
+	wal          *wal.GarbageCollectedWAL
+	msm          *metadata.StateMachine
+	e            *simplex.Epoch
+	nv           *nonvalidator.NonValidator
+	epochOrNV    timeAdvancer
+	epochChanges chan epochChange
+	stopCh       chan struct{}
```
samliok, line 76: can we add some helpful comments to these fields?

instance.go
```diff
@@ -0,0 +1,646 @@
...
+	}
+}
+
+func (i *Instance) tick() {
+	ticker := time.NewTicker(tickInterval)
+	for {
+		select {
+		case now := <-ticker.C:
+			i.lock.Lock()
+			timeAdvancer := i.epochOrNV
```
samliok, line 211: can we remove `i.epochOrNV`? it think it would be fine to just check if `i.e != nil { }` or `i.nonValidator != nil`

instance.go
```diff
@@ -0,0 +1,646 @@
...
+	return i.startEpoch(epochConfig)
+}
+
+func (i *Instance) startNonValidator(epochNum uint64, validators common.Nodes) error {
+	config, err := i.createNonValidatorConfig(epochNum, validators)
+	if err != nil {
+		return err
+	}
+
+	nonValidator, err := nonvalidator.NewNonValidator(config)
```
samliok, line 131: wait, doesn't this restart the validator? so in the case of non-validator -> non-validator we still run into the problem of resetting the memory & needing to rebroadcast

samliok (commented):
i think my previous approval was premature, so i'm just noting some more comments while i'm working on related code.

instance.go
yacovm: No, if we have a nonvalidator-->nonvalidator we do not do an epoch change: https://github.com/ava-labs/Simplex/pull/427/changes/BASE..12fcee7776a03e91664accc93cd440fa7d46b610#diff-f4581317ccb65231765f0fa1aec4aa2bb1c08b09fe9dc1faf095d9039483e4fdR164

yacovm:
```
		onEpochChange: func(epoch uint64, validators common.Nodes) error {
			height := i.Config.PlatformChain.GetCurrentHeight()
			vdrs, err := i.Config.PlatformChain.GetValidatorSet(height)
			if err != nil {
				i.Config.Logger.Error("error getting validator set", zap.Error(err))
				return fmt.Errorf("error getting validator set from platform chain: %w", err)
			}
			comm.SetValidators(validators)
			if i.iAmValidator(vdrs.Nodes()) {
				i.notifyEpochChange(epoch, validators, nonValidator)
			} else {
				i.Config.Logger.Debug("I am still a non-validator at the tip of the P-chain, skipping role change",
					zap.Uint64("height", height))
			}
			return nil
		},
```

instance.go
yacovm: we can remove it... sure. Do you want to make a PR and I'll approve?

instance.go
yacovm: we only need it for MSM

msm/msm.go
samliok, line 1528: What happens if other nodes timed out and we call generate twice? maybe the validator set is small so if 2 nodes time out, we will come back to our node and generate again

adapters.go
```diff
@@ -0,0 +1,253 @@
...
+func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metadata.StateMachineBlock, *common.Finalization, error) {
+	block, finalization, err := cs.Retrieve(seq, digest)
+	if err != nil {
+		return metadata.StateMachineBlock{}, nil, err
+	}
+
+	return block.(*ParsedBlock).StateMachineBlock, finalization, nil
+}
+
+func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.VerifiedBlock, *common.Finalization, error) {
```
samliok, line 114: `Retrieve` may return a block if its only been verified. Is this a problem in the MSM code on the `GetBlock` callsites? I initially thought GetBlock would return a notarized or finalized block, in this case it can return neither.

msm/msm.go
samliok, line 1131: is it possible for the `AuxInfoDigest` to change once we are in the collecting approvals phase? i.e we are in the approvals phase if `IsSufficient` returns true, but can't another block add more aux info and `IsSufficient` would still be true, however the last digest would change.

samliok, line 1076: do we ever marshal using canato? doesn't seem like it, so why do we include canator data?

instance.go
```diff
@@ -0,0 +1,646 @@
...
+		epoch:   epochNum,
+		Storage: i.Config.Storage,
+		onEpochChange: func(epoch uint64, validators common.Nodes) error {
+			height := i.Config.PlatformChain.GetCurrentHeight()
+			vdrs, err := i.Config.PlatformChain.GetValidatorSet(height)
+			if err != nil {
+				i.Config.Logger.Error("error getting validator set", zap.Error(err))
+				return fmt.Errorf("error getting validator set from platform chain: %w", err)
+			}
+			comm.SetValidators(validators)
```
samliok, line 160: i think this is actually a pretty big bug. The non-validator expects `comm.Validators` to be the most up-to-date validator set, rather than the validator set of the transitioning epoch. 

What if we have a completely new set of validators at the tip and set the validators to a disjoint set here? Any replication requests would fail, and we would be stuck.

instance.go
```diff
@@ -0,0 +1,646 @@
...
+			return fmt.Errorf("error parsing protocol metadata from last block: %w", err)
+		}
+		if err := i.wal.GarbageCollect(md.Round); err != nil {
+			return fmt.Errorf("error garbage collecting WALs: %w", err)
+		}
+	}
+	return nil
+}
+
+func (i *Instance) transitionEpochNonValidator(epochChange epochChange) error {
```
samliok, line 531: i think i mentioned this before, or maybe you did but we need to ensure the epoch we are changing into is the most current. If they are in this epochs validator set, yet are behind we should restart them as a non-validator. 

Just noting this so we don't forget

instance.go
samliok: Also we can't just use the latest pchain tip, we need to know the pchain height that is referenced by the last epoch...

instance.go
samliok: i dont think this can work because the tip of the pchain might not correspond to the latest validator set. 

I.e. our epoch may be transitioning to a new epoch, which means the latest pchain tip has a different validator set than the validators transitioning the epoch.

yacovm: all I'm saying is, if we remain a non-validator we do not do an epoch change.

PR #428, opened 2026-07-02
Garbage collect blacklist upon epoch change

yacovm:
This commit garbage collects the blacklist upon an epoch change by initializing an empty one by constructing the previous blacklist in verification or block building.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2274,6 +2274,11 @@ func (e *Epoch) verifyProposalMetadataAndBlacklist(block common.Block) bool {
 
 		prevBlacklist = prevBlock.Blacklist()
 
+		// If the previous block belongs to an earlier epoch, the blacklist should be empty.
+		if prevBlock.BlockHeader().Epoch < e.Epoch {
+			prevBlacklist = common.NewBlacklist(uint16(len(e.validatorNodeIDs)))
+		}
+
 		if prevBlacklist.IsEmpty() {
 			prevBlacklist = common.NewBlacklist(uint16(len(e.validatorNodeIDs)))
 		}
@@ -2428,6 +2433,11 @@ func (e *Epoch) retrieveBlacklistOfParentBlock(metadata common.ProtocolMetadata)
 		}
 
 		blacklist = prevBlock.Blacklist()
+
+		// If the previous block belongs to a previous epoch, we need to reset the blacklist to be empty.
+		if metadata.Epoch > prevBlock.BlockHeader().Epoch {
+			blacklist = common.NewBlacklist(uint16(len(e.validatorNodeIDs)))
+		}
 	}
 
 	if blacklist.IsEmpty() {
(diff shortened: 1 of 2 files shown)


simplex/epoch_test.go
```diff
@@ -737,6 +737,43 @@ func TestEpochSimpleFlow(t *testing.T) {
...
+	// Restart the node into epoch 2 with a two-validator set.
+	nodes := []NodeID{{1}, {2}}
+	bb := testutil.NewTestBlockBuilder()
+	conf, wal, _ := testutil.DefaultTestNodeEpochConfig(t, NodeID{2}, testutil.NewNoopComm(nodes), bb)
+	conf.Storage = storage
+	conf.Epoch = 2
+
+	e, err := NewEpoch(conf)
+	require.NoError(t, err)
+	// Run as epoch 2 while the last committed block still belongs to epoch 1.
```
samliok: can we confirm we are running as epoch two after we start?

require.Equal(t, uint64(2), e.Metadata().Epoch)

samliok: while it passes with this now, i'm concerned the test may become fragile if we make future epoch changes

simplex/epoch_test.go
```diff
@@ -737,6 +737,43 @@ func TestEpochSimpleFlow(t *testing.T) {
...
 }
 
+func TestEpochResizesBlacklistOnEpochChange(t *testing.T) {
+	storage := testutil.NewInMemStorage()
+	epoch1Block := testutil.NewTestBlock(ProtocolMetadata{Epoch: 1, Round: 0, Seq: 0}, NewBlacklist(1))
+	require.NoError(t, storage.Index(context.Background(), epoch1Block, Finalization{}))
+	require.Equal(t, uint16(1), epoch1Block.Blacklist().NodeCount,
+		"blacklist must contain exactly one node")
+
+	// Restart the node into epoch 2 with a two-validator set.
```
samliok: we are not restarting though?

simplex/epoch.go
```diff
@@ -2274,6 +2274,11 @@ func (e *Epoch) verifyProposalMetadataAndBlacklist(block common.Block) bool {
 
 		prevBlacklist = prevBlock.Blacklist()
 
+		// If the previous block belongs to an earlier epoch, the blacklist should be empty.
+		if prevBlock.BlockHeader().Epoch < e.Epoch {
```
samliok, line 2278: should we add a test for this?

simplex/epoch.go
samliok: as an aside, this function name doesn't retrieve the blacklist of the parent block. it retrieves the blacklist we should use for block building, which is not always the prev blacklist

yacovm: renamed

simplex/epoch.go
yacovm: changed the test accordingly

simplex/epoch_test.go
yacovm: removed

simplex/epoch_test.go
yacovm: added

samliok (approved):

PR #429, opened 2026-07-02
Move approval store into MSM

yacovm:
This commit moves the approval store into the MSM, instead of being an external dependency.

The reason is that the approval store needs to be initialized with the validator set of the *next* epoch and not the *current epoch*, which can only be computed inside the MSM.

### DIFF
--- a/msm/approvals.go
+++ b/msm/approvals.go
@@ -5,6 +5,7 @@ package metadata
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/ava-labs/simplex/common"
 	"go.uber.org/zap"
@@ -27,8 +28,12 @@ type ApprovalStore struct {
 	validators        NodeBLSMappings
 	logger            common.Logger
 	pkByNodeID        map[nodeID][]byte
-	approvalsByNodes  map[nodeID]approvalsByPChainHeightAndAuxInfoDigest
-	storedCount       int
+	// lock guards the mutable state below (approvalsByNodes, storedCount). The
+	// store is accessed concurrently: approvals are handled as they arrive while
+	// the block builder reads the accumulated approvals when building a block.
+	lock             sync.RWMutex
+	approvalsByNodes map[nodeID]approvalsByPChainHeightAndAuxInfoDigest
+	storedCount      int
 }
 
 func NewApprovalStore(signatureVerifier SignatureVerifier, validators NodeBLSMappings, logger common.Logger) *ApprovalStore {
@@ -52,6 +57,9 @@ func NewApprovalStore(signatureVerifier SignatureVerifier, validators NodeBLSMap
 }
 
 func (as *ApprovalStore) Approvals() ValidatorSetApprovals {
+	as.lock.RLock()
+	defer as.lock.RUnlock()
+
 	approvals := make(ValidatorSetApprovals, 0, as.storedCount)
 	for _, approvalsByHeight := range as.approvalsByNodes {
 		for _, approval := range approvalsByHeight {
@@ -70,17 +78,21 @@ func (as *ApprovalStore) HandleApproval(approval *ValidatorSetApproval, timestam
 		return nil
 	}
 
-	// Second thing we check is if we already have an approval for this height from this node.
-	if as.approvalExistsAndUpToDate(approval, timestamp) {
-		as.logger.Debug("Already have an approval from the node", zap.String("nodeID",
+	// Second thing we check is if the signature of the approval is valid.
+	// We need it to be valid in order for nodes to be able to aggregate it later on along with other approvals.
+	// This is checked before taking the lock, as it only reads immutable state.
+	if err := as.checkApprovalSignature(approval, pk); err != nil {
+		as.logger.Debug("Received an approval with an invalid signature", zap.String("nodeID",
 			fmt.Sprintf("%x", approval.NodeID)), zap.Uint64("pChainHeight", approval.PChainHeight))
 		return nil
 	}
 
-	// Third thing we check is if the signature of the approval is valid.
-	// We need it to be valid in order for nodes to be able to aggregate it later on along with other approvals.
-	if err := as.checkApprovalSignature(approval, pk); err != nil {
-		as.logger.Debug("Received an approval with an invalid signature", zap.String("nodeID",
+	as.lock.Lock()
+	defer as.lock.Unlock()
+
+	// Third thing we check is if we already have an approval for this height from this node.
+	if as.approvalExistsAndUpToDate(approval, timestamp) {
+		as.logger.Debug("Already have an approval from the node", zap.String("nodeID",
 			fmt.Sprintf("%x", approval.NodeID)), zap.Uint64("pChainHeight", approval.PChainHeight))
 		return nil
 	}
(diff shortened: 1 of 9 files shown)


msm/approvals.go
```diff
@@ -27,8 +28,12 @@ type ApprovalStore struct {
 	validators        NodeBLSMappings
 	logger            common.Logger
 	pkByNodeID        map[nodeID][]byte
```
samliok: unrelated but should we change this to `nodeIDToPK`. we already have `validatorsToPKs                map[string][]byte` in epoch.go

msm/approvals.go
```diff
@@ -27,8 +28,12 @@ type ApprovalStore struct {
...
-	approvalsByNodes  map[nodeID]approvalsByPChainHeightAndAuxInfoDigest
-	storedCount       int
+	// lock guards the mutable state below (approvalsByNodes, storedCount). The
+	// store is accessed concurrently: approvals are handled as they arrive while
+	// the block builder reads the accumulated approvals when building a block.
+	lock             sync.RWMutex
+	approvalsByNodes map[nodeID]approvalsByPChainHeightAndAuxInfoDigest
+	storedCount      int
```
samliok, line 36: why do we need the `storedCount` field? can't we just calculate it by summing the lengths in `approvalsByNodes`?

msm/approvals.go
samliok: unrelated: can we remove the getPKOfNode() function? it seems unnecessary

msm/approvals.go
```diff
@@ -167,3 +179,28 @@ func (as *ApprovalStore) approvalExistsAndUpToDate(approval *ValidatorSetApprova
 
 	return existingApproval.Timestamp >= timestamp
 }
+
+func (as *ApprovalStore) PutApprovals(approvalStore *ApprovalStore) {
+	// Snapshot the approvals under our lock, then hand them to the destination
+	// store (which takes its own lock). Copying first avoids holding two store
+	// locks at once.
+	as.lock.Lock()
```
samliok: we can just grab the read part of the lock rather than the entire lock

msm/approvals.go
```diff
@@ -167,3 +179,28 @@ func (as *ApprovalStore) approvalExistsAndUpToDate(approval *ValidatorSetApprova
...
+	type approvalWithTimestamp struct {
```
samliok: why do we create a new struct? can we use `approvalAndTimestamp`?

msm/approvals.go
samliok: nvm i guess it would be inefficient to iterate and find exactly how many entries there are, and probably hard to get an accurate estimate

msm/approvals_test.go
```diff
@@ -222,6 +222,152 @@ func TestApprovalStoreHandleApproval(t *testing.T) {
...
+			}
+
+			src.PutApprovals(dst)
+
+			// storedCount must always stay in sync with the number of retrievable approvals.
+			require.Len(t, dst.Approvals(), dst.storedCount)
+			tc.verify(t, dst, src, tc.srcApprovals, tc.dstApprovals)
+		})
+	}
+}
```
samliok, line 389: should we add one where the timestamp is equal

```go
{
			// A source approval with the same timestamp as an existing destination approval at the same
			// (NodeID, PChainHeight) does not overwrite it: the >= tie in approvalExistsAndUpToDate goes
			// to the incumbent.
			name:          "equal-timestamp source does not overwrite destination",
			srcValidators: 2,
			dstValidators: 2,
			dstApprovals: []approvalAndTimestamp{
				{ValidatorSetApproval{NodeID: makeNodeID(1), PChainHeight: 7, Signature: signApproval(7, [32]byte{})}, 100},
			},
			srcApprovals: []approvalAndTimestamp{
				{ValidatorSetApproval{NodeID: makeNodeID(1), PChainHeight: 7, Signature: signApproval(7, [32]byte{})}, 100},
			},
			verify: func(t *testing.T, dst, _ *ApprovalStore, _, dstSent []approvalAndTimestamp) {
				got := dst.Approvals()
				require.Len(t, got, 1)
				require.Equal(t, 1, dst.storedCount)
				require.Equal(t, dstSent[0].ValidatorSetApproval, got[0], "the destination approval is retained on a timestamp tie")
			},
		},
```

msm/msm.go
```diff
@@ -618,6 +650,12 @@ func (sm *StateMachine) verifyNextPChainRefHeightNormal(prevMD StateMachineMetad
 			errValidatorSetUnchanged, next.NextPChainReferenceHeight, prev.PChainReferenceHeight)
 	}
 
+	// Else, ! currentValidatorSet.Equal(newValidatorSet) || next.NextPChainReferenceHeight == 0
```
```suggestion
	// Else, !currentValidatorSet.Equal(newValidatorSet) || next.NextPChainReferenceHeight == 0
```

msm/msm.go
```diff
@@ -468,6 +499,7 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
 	if decisionToBuildBlock.transitionEpoch && isSealingBlockFinalized {
 		sm.Logger.Debug("Transitioning epoch after building block", zap.Uint64("newPChainRefHeight", decisionToBuildBlock.pChainHeight))
 		newSimplexEpochInfo.NextPChainReferenceHeight = decisionToBuildBlock.pChainHeight
+		sm.maybeInitializeApprovalStore(decisionToBuildBlock.validatorSet)
```
samliok, line 502: but if the sealing block is finalized, we must have already computed an approval store. furthermore, since this is for the next epoch aren't these validators not the right validators anyways? we want the validator set for the next, next epoch which we can't know since we are only transitioning to the next epoch.

msm/approvals.go
yacovm: not sure I follow, what is the problem with the field name?

msm/approvals.go
yacovm: It's so we can allocate the exact size in `Approvals()` and avoid memory copying

msm/approvals.go
yacovm: yes, I can change it accordingly.

Though you know, when this is called, the approval store has a few micro seconds left to live 💀

msm/msm.go
yacovm: This is the sealing block of the previous epoch, not this epoch. 
The sealing block of this epoch hasn't been created yet. 

> since this is for the next epoch aren't these validators not the right validators anyways? 

I changed `shouldBuildBlock` to return the validator set at `decisionToBuildBlock.pChainHeight` which is written into `newSimplexEpochInfo.NextPChainReferenceHeight`. 

> we want the validator set for the next, next epoch which we can't know since we are only transitioning to the next epoch.

This function determines the P-chain reference height for the next epoch, and that's the validator set we are seeding the approval store with.

msm/approvals.go
samliok: i was just suggesting we rename to `pkByNodeID` -> `nodeIDToPk

Just to keep it consistent with how we name the same thing in `epoch.go`

yacovm: well I renamed it https://github.com/ava-labs/Simplex/pull/429/changes/8bfb075f387d342eb8589ff5cc7dee5cc74a6a32#diff-5e383ab850834861309c7e77b4123dbe73b6b4d3e2b96f96832869f5847b48fdR30

msm/approvals.go
```diff
@@ -167,3 +174,24 @@ func (as *ApprovalStore) approvalExistsAndUpToDate(approval *ValidatorSetApprova
 
 	return existingApproval.Timestamp >= timestamp
 }
+
+func (as *ApprovalStore) PutApprovals(approvalStore *ApprovalStore) {
```
samliok, line 179: can we clarify this function signature a bit more? 

something like 
```go
// puts approvals from `as` into `dest`
PutApprovals(dest *ApprovalsStore)

CopyApprovalsTo(dest *ApprovalStore)
```

msm/approvals.go
```diff
@@ -26,9 +27,13 @@ type ApprovalStore struct {
...
+	nodeIDToPK        map[nodeID][]byte
+	// lock guards the mutable state below (approvalsByNodes, storedCount). The
+	// store is accessed concurrently: approvals are handled as they arrive while
+	// the block builder reads the accumulated approvals when building a block.
+	lock             sync.RWMutex
+	approvalsByNodes map[nodeID]approvalsByPChainHeightAndAuxInfoDigest
+	storedCount      int
 }
 
 func NewApprovalStore(signatureVerifier SignatureVerifier, validators NodeBLSMappings, logger common.Logger) *ApprovalStore {
```
samliok, line 39: what if we make `putApprovals` private and pass in a `prev *ApprovalStore` here? this way we can simplify the callsite in `maybeInitializeApprovalStore` to 

```go
		sm.approvalStore = NewApprovalStore(sm.SignatureVerifier, validatorSet, sm.Logger, sm.approvalStore)
```

yacovm: It will slightly simplify the function `maybeInitializeApprovalStore` in `msm.go` but it will make the approval store API awkward. 

It's very rare to see a constructor take its own return type, and we'll also need to lock the receiver approvals instance, which is even worse (object B locking internal fields of object A).

msm/approvals.go
yacovm: https://github.com/ava-labs/Simplex/pull/429/changes/0be37b08299b7cb1b8a05bcf88d4dc1c097804da#diff-5e383ab850834861309c7e77b4123dbe73b6b4d3e2b96f96832869f5847b48fdR178

samliok (approved):

PR #430, opened 2026-07-02
Implement Close() for GarbageCollectedWAL

yacovm:
GarbageCollectedWAL is the top most WAL that the Epoch holds, but it needs to have Close() in order to fully implement WriteAheadLog.

### DIFF
--- a/wal/gc.go
+++ b/wal/gc.go
@@ -10,6 +10,10 @@ import (
 	"github.com/ava-labs/simplex/common"
 )
 
+const (
+	defaultMaxWALSize = 100 * 1024 * 1024 // 100 MB
+)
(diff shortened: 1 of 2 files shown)


samliok (approved):
approving but i looked in the orchestrator pr and we always set retention term to max int. if we are doing this why even have retention term at all?

yacovm:
> approving but i looked in the orchestrator pr and we always set retention term to max int. if we are doing this why even have retention term at all?

Well, if we use retention term maxint - it's to garbage collect everything.

However, we cannot only garbage collect upon epoch change. We also want to do that occasionally on some cadence to free up storage space.

PR #431, opened 2026-07-02
fix replication requests exceeding maxRoundWindow + splitting resends

jadalsmail:
Motivation: 
In resendReplicationRequests, timed-out sequences were compressed into contiguous segments and each segment was sent whole to a single node, which can be larger than maxRoundWindow. Thus, nodes drop such requests without responding, so a lagging node would resend the same oversized request forever and never catch up. 

initial requests and timed-out retries now share a single send path. A new function, BatchSequences, which distributes seqs as evenly as possible among nodes, returning one batch per node share. Every batch contains at most maxSize sequences. 

We also introduce a new variable, MaxRoundRequests, which is the max number of rounds or sequences that fit in one replication request. Every batch is capped at this size by construction. It is set to 10 to match DefaultMaxRoundWindow

Removed: CompressSequences, DistributeSequenceRequests, and the Segment type (plus their tests). the segment/range model they implemented is replaced by batching, and requests may now contain non-contiguous sequences (responders sort incoming requests).

Also, unit test + table tests (for BatchSequences) for testing this functionality were added.


simplex/util.go
```diff
@@ -262,6 +262,22 @@ func DistributeSequenceRequests(start, end uint64, numNodes int) []Segment {
...
+func DistributeMissingSequences(missingSeqs []uint64, numNodes int, maxSize uint64) []Segment {
+	var segments []Segment
+	if maxSize == 0 {
+		return nil
+	}
+	for _, segment := range CompressSequences(missingSeqs) {
+		for _, distributed := range DistributeSequenceRequests(segment.Start, segment.End, numNodes) {
+			for start := distributed.Start; start <= distributed.End; start += maxSize {
+				segments = append(segments, Segment{Start: start,
+					End: min(start+maxSize-1, distributed.End)})
```
jadalsmail: This is where we are enforcing the maxRoundWindow

simplex/util.go
```diff
@@ -262,6 +262,22 @@ func DistributeSequenceRequests(start, end uint64, numNodes int) []Segment {
 	return segments
 }
 
+func DistributeMissingSequences(missingSeqs []uint64, numNodes int, maxSize uint64) []Segment {
```
samliok: since this function is exported, it should have a function comment

simplex/util.go
```diff
@@ -262,6 +262,22 @@ func DistributeSequenceRequests(start, end uint64, numNodes int) []Segment {
...
+	var segments []Segment
+	if maxSize == 0 {
+		return nil
+	}
+	for _, segment := range CompressSequences(missingSeqs) {
+		for _, distributed := range DistributeSequenceRequests(segment.Start, segment.End, numNodes) {
+			for start := distributed.Start; start <= distributed.End; start += maxSize {
```
samliok: i see, so `maxSize` caps each individual segment to at most `maxSize` in length. Can we make this a bit more clear with some comments and variable names?

simplex/util_test.go
```diff
@@ -378,6 +378,113 @@ func TestDistributeSequenceRequests(t *testing.T) {
...
+			missingSeqs: []uint64{9, 7, 3, 0, 1, 2, 8},
+			numNodes:    2,
+			maxSize:     10,
+			expected: []Segment{
+				{Start: 0, End: 1},
+				{Start: 2, End: 3},
+				{Start: 7, End: 8},
+				{Start: 9, End: 9},
+			},
+		},
```
samliok: since our `maxSize` is 10, isn't this the ideal distribution?

```go
{Start: 0, End: 3},
{Start: 7, End: 9},
```

this way we don't send twice as many requests as necessary.

samliok: i think it stems from the fact that we `CompressSequences` making segments, and then for each segment we distribute them evenly amount the nodes. 

So if we are missing ids 1,2,3,4,5 ---- 11,12,13,14,15 ----, and so on. the algorithm would spit each segment across 5 nodes. So we would send soo many more segments than necessary(even if maxSize is MaxUint64).

samliok: ex this test passes, but would result in the sending of too many requests

```go
{
			name:        "1",
			missingSeqs: []uint64{0, 1, 2, 3, 4, 6, 7, 8, 9, 10, 12, 13, 14, 15},
			numNodes:    5,
			maxSize:     40,
			expected: []Segment{
				{Start: 0, End: 0},
				{Start: 1, End: 1},
				{Start: 2, End: 2},
				{Start: 3, End: 3},
				{Start: 4, End: 4},
				{Start: 6, End: 6},
				{Start: 7, End: 7},
				{Start: 8, End: 8},
				{Start: 9, End: 9},
				{Start: 10, End: 10},
				{Start: 12, End: 13},
				{Start: 14, End: 15},
			},
		},

```

simplex/requestor.go
```diff
@@ -128,7 +128,10 @@ func (r *requestor) resendReplicationRequests(missingIds []uint64) {
 	r.epochLock.Lock()
 	defer r.epochLock.Unlock()
 
-	segments := CompressSequences(missingIds)
+	// split each contiguous range among the nodes that signed,
+	// just like the initial requests, so that a single request never exceeds
+	// the maxRoundWindow limit enforced by the responding nodes
+	segments := DistributeMissingSequences(missingIds, len(r.highestObserved.signers), r.maxRoundWindow)
```
samliok: ideally a nice solution should be used by both the initial requests and future requests as to not maintain two different code paths

simplex/epoch.go
```diff
@@ -30,6 +30,7 @@ const (
 	DefaultEmptyVoteRebroadcastTimeout    = 5 * time.Second
 	DefaultFinalizeVoteRebroadcastTimeout = 6 * time.Second
 	EmptyVoteTimeoutID                    = "rebroadcast_empty_vote"
+	MaxRoundRequests                      = 10
```
samliok: lets add a comment as to what this is used for and why its needed(ik the other ones aren't commented) but probably good to do

simplex/requestor.go
```diff
@@ -157,76 +153,69 @@ func (r *requestor) observedSignedQuorum(observed *signedQuorum, currentSeqOrRou
...
 		index := (i + r.requestIterator) % numNodes
-		r.sendRequestToNode(seqsOrRounds.Start, seqsOrRounds.End, r.highestObserved.signers[index])
+		r.sendRequestToNode(batch, signers[index])
 	}
+	r.requestIterator++
 }
 
 // sendRequestToNode requests [start, end] from nodes[index].
 // In case the nodes[index] does not respond, we create a timeout that will
 // re-send the request.
```
samliok: this function comment is out of sync now

simplex/requestor.go
```diff
@@ -157,76 +153,69 @@ func (r *requestor) observedSignedQuorum(observed *signedQuorum, currentSeqOrRou
...
 
-	if len(seqsOrRound) == 0 {
+	if len(toRequest) == 0 {
 		return
 	}
 
-	if r.highestRequested < end {
-		r.highestRequested = end
+	if last := toRequest[len(toRequest)-1]; last > r.highestRequested {
+		r.highestRequested = last
```
samliok, line 216: this assumes that the slice is already sorted. just from looking at the function its not clear why can assume this. I think we should note in the function header `seqsOrRounds` is expected to be sorted.

simplex/requestor.go
```diff
@@ -157,76 +153,69 @@ func (r *requestor) observedSignedQuorum(observed *signedQuorum, currentSeqOrRou
...
+	}
 	r.logger.Debug("Node is behind, attempting to request missing values", zap.Uint64("value", observedSeqOrRound), zap.Uint64("start", uint64(start)), zap.Uint64("end", uint64(end)), zap.Bool("seq requestor", r.replicateSeqs))
-	r.sendReplicationRequests(uint64(start), uint64(end))
+	r.sendRequests(seqsOrRounds)
 }
 
-// sendReplicationRequests sends requests for missing sequences for the
-// range of sequences [start, end] <- inclusive. It does so by splitting the
-// range of sequences equally amount the nodes that have signed [highestObserved].
-func (r *requestor) sendReplicationRequests(start uint64, end uint64) {
```
samliok, line 171: lets not remove the function comments, but rather make them relevant

simplex/util.go
```diff
@@ -185,81 +185,31 @@ func (block *oneTimeVerifiedBlock) Verify(ctx context.Context) (common.VerifiedB
...
-// compressSequences takes a slice of uint64 values representing
-// missing sequence numbers and compresses consecutive numbers into segments.
-// Each segment represents a continuous block of missing sequence numbers.
-func CompressSequences(missingSeqs []uint64) []Segment {
-	slices.Sort(missingSeqs)
-	var segments []Segment
-
-	if len(missingSeqs) == 0 {
-		return segments
+func BatchSequences(seqs []uint64, numNodes int, maxSize uint64) [][]uint64 {
```
samliok: same comment about function headers

samliok: can we pass in numNodes as a uint64? we convert it to a uint64 anyways later and i think it makes more sense for it to be unsigned

simplex/util.go
```diff
@@ -185,81 +185,31 @@ func (block *oneTimeVerifiedBlock) Verify(ctx context.Context) (common.VerifiedB
...
-	nodeStart := start
-
-	for i := 0; i < numNodes && nodeStart <= end; i++ {
-		segmentLength := seqsPerNode
-		if remainder > 0 {
-			segmentLength++
-			remainder--
+	var batches [][]uint64
+	start := uint64(0)
+	for i := 0; i < numNodes && start < numSeqs; i++ {
```
samliok: you can define i := uint64(0) instead of converting later

samliok: we can use `slices.Chunk()` to make this much simpler

```go
batchSize := min((numSeqs+uint64(numNodes)-1)/uint64(numNodes), maxSize)
return slices.Collect(slices.Chunk(seqs, int(batchSize)))
```

samliok: although i would declare a helper variable to make it more readable

```go
seqsPerNode := (numSeqs + uint64(numNodes) - 1) / uint64(numNodes)
batchSize := min(seqsPerNode, maxSize)
return slices.Collect(slices.Chunk(seqs, int(batchSize)))
```

simplex/util_test.go
```diff
@@ -242,137 +242,135 @@ func TestGetHighestQuorumRound(t *testing.T) {
...
+			},
 		},
 		{
-			name:     "zero nodes",
-			start:    0,
-			end:      10,
-			numNodes: 0,
-			expected: nil,
+			name:     "node shares exceed maxSize and are chunked",
+			seqs:     []uint64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14},
```
samliok, line 360: isn't it ideal to be sending 5 requests here? regardless i think the slices.Chunk stuff should fix this

yacovm (commented):
I also want to review this change before it gets merged so please don't merge it yet

yacovm:
> This PR fixes the following bug:
> 
> * in resendReplicationRequests, we aggregate continuous sequences which can be larger than maxRoundWindow
> 
> Also, unit tests testing this functionality were added.

Can we please have a more descriptive and elaborate PR description?

.gitignore
samliok, line 1: lets undo this diff

samliok (approved):

simplex/epoch.go
```diff
@@ -30,6 +30,7 @@ const (
 	DefaultEmptyVoteRebroadcastTimeout    = 5 * time.Second
 	DefaultFinalizeVoteRebroadcastTimeout = 6 * time.Second
 	EmptyVoteTimeoutID                    = "rebroadcast_empty_vote"
+	MaxRoundRequests                      = 10 // max number of rounds or sequences that fit in one replication request
```
yacovm: maxItemCountPerRequest?

simplex/requestor.go
```diff
@@ -157,76 +156,75 @@ func (r *requestor) observedSignedQuorum(observed *signedQuorum, currentSeqOrRou
...
 	}
 
-	if len(seqsOrRound) == 0 {
+	if len(toRequest) == 0 {
 		return
 	}
 
-	if r.highestRequested < end {
-		r.highestRequested = end
+	if last := toRequest[len(toRequest)-1]; last > r.highestRequested {
```
yacovm, line 215: I think it wouldn't hurt to put here:

```
	if ! slices.IsSorted(toRequest) {
		slices.Sort(toRequest)
	}
```

Because we are given this `seqsOrRounds` as input.

Worst case with it, it will scan the slice once.

Worst case without this, is that we call this function without a sorted input and introduce a bug.

simplex/replication_timeout_test.go
```diff
@@ -677,3 +677,70 @@ func TestReplicationResendsFinalizedBlocksThatFailedVerification(t *testing.T) {
 	require.Equal(t, uint64(1), storage.NumBlocks())
 	require.Equal(t, block, storedBlock)
 }
+
+// TestReplicationResendSplitsRequests ensures that when replication requests time out, the missing sequences
```
yacovm: can we not put the description here? We should strive to be uniform with other tests. 

```
func TestReplicationResendSplitsRequests(t *testing.T) {
	// Ensures that when replication requests time out, the missing sequences
	// are re-requested split across the nodes that signed the highest observed quorum, just like the initial requests are.
...
}
```

simplex/replication_timeout_test.go
```diff
@@ -677,3 +677,70 @@ func TestReplicationResendsFinalizedBlocksThatFailedVerification(t *testing.T) {
...
+// are re-requested split across the nodes that signed the highest observed quorum, just like the initial requests are.
+func TestReplicationResendSplitsRequests(t *testing.T) {
```
yacovm, line 681: This tests doesn't reproduce the original problem described in the PR. 

> In resendReplicationRequests, timed-out sequences were compressed into contiguous segments and each segment was sent whole to a single node, which can be larger than maxRoundWindow. Thus, nodes drop such requests without responding

Can't we make a test that properly reproduces the problem with a real epoch instance that acts as the responder?

yacovm (changes requested):

simplex/replication_timeout_test.go
```diff
@@ -678,69 +678,53 @@ func TestReplicationResendsFinalizedBlocksThatFailedVerification(t *testing.T) {
...
-	// no node responds, so the requests time out and are re-sent
-	e.AdvanceTime(e.StartTime.Add(2 * simplex.DefaultReplicationRequestTimeout))
+	testutil.NewControlledSimplexNode(t, nodes[0], net, newNodeConfig(nodes[0]))
+	testutil.NewControlledSimplexNode(t, nodes[1], net, newNodeConfig(nodes[1]))
+	testutil.NewControlledSimplexNode(t, nodes[2], net, newNodeConfig(nodes[2]))
+	laggingNode := testutil.NewControlledSimplexNode(t, nodes[3], net, &testutil.TestNodeConfig{
+		ReplicationEnabled: true,
+		// twice the responder's window: the timed-out sequences span more
+		// than a responder is willing to server in a single request
+		MaxRoundWindow: startSeq,
```
yacovm, line 706: Ideally we would reproduce the problem without specifying unique configuration such as modifying the max round window.

jadalsmail: I agree, but here's the thing, the bug was reproducible with default config before #432, but that fix removed its trigger (outstanding requests can no longer exceed the responder limit when windows match). Since this PR merges both send paths into one function, the aggregation half can't fire on its own anymore. The window override recreates the same trigger condition #432 eliminated, so the test actually exercises the drop-and-wedge path and proves the new code survives it.

yacovm: > I agree, but here's the thing, the bug was reproducible with default config before #432, but that fix removed its trigger (outstanding requests can no longer exceed the responder limit when windows match). Since this PR merges both send paths into one function, the aggregation half can't fire on its own anymore. The window override recreates the same trigger condition #432 eliminated, so the test actually exercises the drop-and-wedge path and proves the new code survives it.

Can you explain what you mean by: 

> drop-and-wedge

?

jadalsmail: Drop the requests that are coming from the peer nodes.
wedge: keeping the node behind (lagging forever)

yacovm: > Since this PR merges both send paths into one function, the aggregation half can't fire on its own anymore. 

I do not expect to create a scenario that is influenced by existing production code changes from this PR. 

Is it not possible to create a test with two epochs or more that if I completely revert all production code changes and leave the test intact, the test will fail because of the reason the PR claims to be the issue?

jadalsmail: I am a bit confused, could you explain to me how to go about writing this test? 
Right now, if I revert all of the changes and leave the test intact, it would fail. But I can't think of a way to reproduce the problem without specifying unique configuration. I would appreciate some guidance on how to create the test for this PR.

yacovm (approved):

PR #432, opened 2026-07-06
fix off by one replication request error and add unit test

jadalsmail:
This PR fixes the following bug:

- When a node sends a replication requests, it sends more than maxRoundWindow (off-by-one)

Also one unit test for this functionality was added.

### DIFF
--- a/simplex/replication_test.go
+++ b/simplex/replication_test.go
@@ -1760,3 +1760,51 @@ messages:
 	notarization := wal.AssertNotarization(uint64(len(blocks)))
 	require.Equal(t, common.EmptyNotarizationRecordType, notarization)
 }
+
+
+// TestReplicationRequestsWithinMaxRoundWindow ensures that a node that observes a finalization more than MaxRoundWindow
+// sequences ahead of its next sequence to commit only requests sequences up to MaxRoundWindow-1 ahead of it
+func TestReplicationRequestsWithinMaxRoundWindow(t *testing.T) {
+	bb := testutil.NewTestBlockBuilder()
+	nodes := []common.NodeID{{1}, {2}, {3}, {4}}
+	sentMessages := make(chan *common.Message, 100)
+	conf, _, _ := testutil.DefaultTestNodeEpochConfig(t, nodes[1], &recordingComm{
+		Communication: testutil.NewNoopComm(nodes),
+		SentMessages:  sentMessages,
+	}, bb)
+	conf.ReplicationEnabled = true
+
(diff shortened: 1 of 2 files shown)


simplex/replication_test.go
```diff
@@ -1760,3 +1760,50 @@ messages:
...
+	// observe a finalization more than MaxRoundWindow sequences ahead of us
+	seqCount := 2 * conf.MaxRoundWindow
+	finalization := createBlocks(t, nodes, seqCount)[seqCount-1].Finalization
+	err = e.HandleMessage(&common.Message{Finalization: &finalization}, nodes[0])
+	require.NoError(t, err)
+
+	// collect the replication requests sent out in response to the finalization
+	requested := make(map[uint64]struct{})
+	timeout := time.After(30 * time.Second)
+	for uint64(len(requested)) < conf.MaxRoundWindow {
```
samliok: this catches the bug, but i'm not sure its the most bulletproof way. This exits once requested >= conf.MaxRoundWindow. What if we send batches like this [0-maxRoundWindow), [maxRoundWindow + 1]. In this case the test would pass because it exits early, yet we are still sending more than we should

simplex/replication_test.go
```diff
@@ -1760,3 +1760,50 @@ messages:
...
+			for _, seq := range msg.ReplicationRequest.Seqs {
+				requested[seq] = struct{}{}
+			}
+		case <-timeout:
+			require.FailNow(t, "timed out waiting for replication requests")
+		}
+	}
+
+	// our next sequence to commit is 0, so we may request at most sequences [0, MaxRoundWindow-1]
+	for seq := range requested {
```
samliok, line 1801: in addition, might as well assert the length of the response is exactly what we want
`	require.Equal(t, conf.MaxRoundWindow, uint64(len(requested)))`

simplex/replication_test.go
```diff
@@ -1760,3 +1760,50 @@ messages:
...
+	seqCount := 2 * conf.MaxRoundWindow
+	finalization := createBlocks(t, nodes, seqCount)[seqCount-1].Finalization
+	err = e.HandleMessage(&common.Message{Finalization: &finalization}, nodes[0])
+	require.NoError(t, err)
+
+	// collect the replication requests sent out in response to the finalization
+	requested := make(map[uint64]struct{})
+	for done := false; !done; {
+		select {
+		case msg := <-sentMessages:
```
samliok: this can be simplified by just looping over the sentMessages channel

samliok (approved):

PR #433, opened 2026-07-08
Abort block building and sending upon commit of an epoch sealing block

yacovm:
Once a block carrying SealingBlockInfo is committed (designates a sealing block), mark the epoch as sealed and stop it from doing any further work: it no longer verifies blocks, builds/proposes blocks, or broadcasts/sends messages.

The motivation is to prevent the epoch from interacting with the VM or other nodes, as a sealed epoch is supposed to be terminated right after its sealing.

High level changes:

- Added epochSealed (an atomic.Bool)
- indexFinalization() now sets epochSealed and returns without progressing to the next round when the committed block is a sealing block.
- broadcast/send now abort if sealed, and the block verification and block-building tasks abort when the epoch is sealed.

Add TestEpochStopsAtSealingBlock, which drives an epoch through a chain via block messages, finalizations and notarizations and asserts that: without a sealing block every block is verified and the epoch proposes a block for the round it leads; with a sealing block only the blocks up to and including it are verified and committed.

TestEpochBuildsInRoundAfterSealingBlock which drives an epoch through a chain via block messages and finalizations and ensures that a block is not built in the round after the epoch is sealed.

### DIFF
--- a/common/msg_test.go
+++ b/common/msg_test.go
@@ -1,3 +1,6 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
 package common_test
 
 import (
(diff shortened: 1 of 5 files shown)


simplex/epoch.go
```diff
@@ -1318,12 +1323,28 @@ func (e *Epoch) rebroadcastPastFinalizeVotes() error {
...
 		}
 		e.Logger.Debug("Rebroadcasting finalize vote", zap.Uint64("round", r), zap.Uint64("seq", finalizeVoteMessage.FinalizeVote.Finalization.Seq))
-		e.Comm.Broadcast(finalizeVoteMessage)
+		e.broadcast(finalizeVoteMessage)
 	}
 
 	return nil
 }
 
+func (e *Epoch) broadcast(msg *common.Message) {
```
samliok, line 1332: shouldn't we still allow certain messages to get through? i.e. we should be able to respond to replication requests, and digest requests.

Also when a node sends an empty vote for an older round we respond with our latest finalization, round and maybe notarization or finalization.
https://github.com/ava-labs/Simplex/blob/ee44f7adf2b82b022245c06d1ab279c3ca47a148/simplex/epoch.go#L897

Also, we have 
https://github.com/ava-labs/Simplex/blob/ee44f7adf2b82b022245c06d1ab279c3ca47a148/simplex/epoch.go#L495

if our epochs are transitioning during this time, the lagging node will not be able to catch up. At least I don't think we ever retry broadcasting the replication sync if no nodes respond.

simplex/epoch.go
```diff
@@ -2016,6 +2047,10 @@ func (e *Epoch) createBlockVerificationTask(block common.Block, from common.Node
 			e.Logger.Debug("Block verification ended", zap.Uint64("round", md.Round), zap.Duration("elapsed", elapsed))
 		}()
 
+		if e.isEpochSealed() {
```
samliok, line 2058: i dont think this code path will ever be hit. We actually will return early from `handleBlockMessage` is the metadata does not equal expected. And if the next block is build on top of a finalized sealing block, the block epoch number will not equal the expected epoch number

yacovm: It is hit in the tests, run with code coverage and see

simplex/epoch.go
yacovm: All right good call, [Fixed](https://github.com/ava-labs/Simplex/pull/433/changes/12ded99aa430c402cd8b967fbb73d50e928ccb57#diff-b6a2eb766dc190b568ec16bd7277c8633013e63dcd6e590db5c72dcc0b69d220L1333-L1342).

I also realized that we drop the finalization message of the sealing block this way so I also fixed that: https://github.com/ava-labs/Simplex/pull/433/changes/05acb41a1d964529e15613324c31cef528309a1a

simplex/epoch.go
samliok: hmm i see it does get hit by coverage. but how is that possible, we should be rejecting it before this point.

samliok: oh its because we are building of an incorrect metadata since we never increase the epoch

common/msg_test.go
```diff
@@ -9,6 +9,96 @@ import (
...
+			msg:      common.Message{FinalizeVote: &common.FinalizeVote{}},
+			expected: false,
+		},
+		{
+			name:     "finalization",
+			msg:      common.Message{Finalization: &common.Finalization{}},
+			expected: false,
+		},
+		{
+			// A block digest request is a request but not a replication message.
```
samliok, line 82: shouldn't we send this?

simplex/epoch.go
yacovm: No, we can hit it in different ways - for example we may already have these tasks scheduled and just waiting to be executed once we finalize the dependency.

common/msg_test.go
yacovm: yeah probably

yacovm: https://github.com/ava-labs/Simplex/pull/433/commits/aab118949f75e5d0bbe84d830b1a7ab1db6426ba

samliok (approved):

PR #435, opened 2026-07-13
Convert protocolMetadata, Blacklist and BlockHeader to canoto

yacovm:
Currently, protocol metadata and blacklist are encoded as opaque bytes in the StateMachineBlocks and have a custom serialization format.

This makes it cumbersome and inefficient to extract each time there is a need to read a protocol metadata or blacklist field.

This commit simply converts them to plain structs encoded with canoto, which removes custom serialization code and removes the need to manually deserialize them each time.


common/encoding.go
```diff
@@ -227,21 +257,75 @@ func ParseEmptyVoteRecord(rawEmptyVote []byte) (ToBeSignedEmptyVote, error) {
 }
 
 func BlockRecordRetentionTerm(record []byte) (uint64, error) {
```
samliok, line 259: should we also canoto the records?

simplex/replication_request_test.go
```diff
@@ -489,6 +491,35 @@ func TestReplicationRequestSeqsAndRoundsTruncated(t *testing.T) {
...
 	resp := msg.VerifiedReplicationResponse
-	require.Equal(t, expected, resp.Data)
+	requireEqualQuorumRounds(t, expected, resp.Data)
+
+}
+
+// requireEqualQuorumRounds compares quorum rounds semantically, avoiding
+// reflect.DeepEqual on canoto-encoded values (whose unexported size cache and
+// nil-vs-empty slices differ between freshly-built and round-tripped values).
+func requireEqualQuorumRounds(t *testing.T, expected, actual []common.VerifiedQuorumRound) {
```
samliok, line 500: // nil-vs-empty slices differ between freshly-built and round-tripped values

?

samliok (dismissed):
`TestGarbageCollectedWALAppendOutOfOrder` is failing

common/encoding.go
yacovm: The point of canoto-ing the protocol metadata and blacklist was that we won't need to parse them in the StateMachineBlock. 

If we are to canoto the block records, that's a separate change.

simplex/replication_request_test.go
yacovm: It just means it is different if the test built it or it's a message sent in the test. I will remove it.

samliok (approved):

PR #438, opened 2026-07-15
Send Replication response up to maxRoundWindow instead of dropping the response

jadalsmail:
When a replication request is above the maxRoundWindow limit, it was dropped. In this PR, we instead send a response with up to maRoundWindow seqs/rounds instead of dropping the request. 
Added two unit tests to test this new functionality for sequences and rounds.


simplex/replication_request_test.go
```diff
@@ -410,3 +410,104 @@ func TestMalformedReplicationResponse(t *testing.T) {
...
+	// request 2*MaxRoundWindow seqs, unsorted (highest first),
+	// to also pin down that truncation happens after sorting
+	requested := make([]uint64, 0, numBlocks)
+	for i := numBlocks; i > 0; i-- {
+		requested = append(requested, i-1)
+	}
+
+	req := &common.Message{
+		ReplicationRequest: &common.ReplicationRequest{
+			Seqs: requested,
```
samliok: can we throw in some unsorted rounds as well?

samliok (commented):
looks good, just that one comment and I'd happily hit approve

simplex/replication_request_test.go
```diff
@@ -410,3 +411,105 @@ func TestMalformedReplicationResponse(t *testing.T) {
...
+	// to also pin down that truncation happens after sorting
+	requested := make([]uint64, 0, numBlocks)
+	for i := numBlocks; i > 0; i-- {
+		requested = append(requested, i-1)
+	}
+
+	req := &common.Message{
+		ReplicationRequest: &common.ReplicationRequest{
+			Seqs:   requested,
+			Rounds: slices.Clone(requested),
```
samliok: We aren't returning any rounds though. I was hoping for something where we return `x` seqs, and `y` rounds and ensure that both the returned rounds and sequences are truncated.

Currently, this test has no way of knowing if the rounds are being properly truncated because the length of `roundData` will always be zero. 

I think you're going to need to only notarize a few blocks past the indexed blocks in order to test this.

samliok (approved):

simplex/replication_request_test.go
```diff
@@ -410,3 +410,132 @@ func TestMalformedReplicationResponse(t *testing.T) {
...
+			Rounds: requestedRounds,
+		},
+	}, nodes[1]))
+
+	msg := <-comm.in
+	resp := msg.VerifiedReplicationResponse
+
+	require.Len(t, resp.Data, int(conf.MaxRoundWindow)+2)
+
+	// firts maxRoundWindow items: 0...9 in order
```
yacovm: firts?

simplex/epoch.go
```diff
@@ -3071,13 +3071,23 @@ func (e *Epoch) handleReplicationRequest(req *common.ReplicationRequest, from co
...
+	}
+	rounds := req.Rounds
+	slices.Sort(rounds)
+	if len(rounds) > int(e.MaxRoundWindow) {
+		e.Logger.Debug("Truncating replication request rounds",
+			zap.Stringer("from", from),
 			zap.Int("num rounds", len(req.Rounds)),
 			zap.Uint64("max round window", e.MaxRoundWindow))
-		return nil
+		rounds = rounds[:e.MaxRoundWindow]
```
yacovm, line 3090: shouldn't we truncate to what's left from `e.MaxRoundWindow - len(seqs)`?

@samliok ?

simplex/replication_request_test.go
```diff
@@ -410,3 +410,132 @@ func TestMalformedReplicationResponse(t *testing.T) {
...
+	for i := uint64(0); i < conf.MaxRoundWindow; i++ {
+		data := resp.Data[i]
+		require.Equal(t, i, data.VerifiedBlock.BlockHeader().Seq)
+		require.NotNil(t, data.Finalization)
+		require.Equal(t, seqs[i].Finalization, *data.Finalization)
+	}
+
+	// last two items: notarized rounds 12, 13 -- 14..16 truncated as well
+	for i, expectedRound := range []uint64{numIndexed, numIndexed + 1} {
```
yacovm: I think an easier way of asserting, would be to simply create an explicit expected response and just compare what we got with what we expected, instead of iterating over and over.

simplex/replication_request_test.go
```diff
@@ -410,3 +410,132 @@ func TestMalformedReplicationResponse(t *testing.T) {
...
+		require.NotNil(t, data.Notarization)
+		require.Nil(t, data.Finalization)
+	}
+
+}
+
+// TestReplicationRequestRoundsTruncated ensures a request with more rounds
+// than MaxRoundWindow is answered with the lowest MaxRoundWindow rounds
+// rather than being dropped.
+func TestReplicationRequestRoundsTruncated(t *testing.T) {
```
yacovm: can you explain why is this test not covered by the previous one?

simplex/epoch.go
samliok: i think currently this is fine, depends on how we want to handle the number of responses.

we can process `MaxRoundWindow` rounds and `MaxRoundWindow` sequences in the response. In the future, i think Jad has a pr to cap the data returned at a maximum number of bytes which is probably a better gaurdrail.

simplex/replication_request_test.go
jadalsmail: It is. before, I had updated the previous one to include both seqs and rounds and didn't think about removing the only rounds one. removed it.

simplex/replication_request_test.go
```diff
@@ -410,3 +410,77 @@ func TestMalformedReplicationResponse(t *testing.T) {
...
+
+	// oversized, unsorted rounds request: 4..24
+	// truncation keeps rounds 4...13
+	roundStart := numIndexed - 8 // 4
+	requestedRounds := make([]uint64, 0, 2*conf.MaxRoundWindow)
+	for i := roundStart + 2*conf.MaxRoundWindow; i > roundStart; i-- {
+		requestedRounds = append(requestedRounds, i-1)
+	}
+
+	e.Comm = comm
```
yacovm: we should override this before we start `e`, not after.

simplex/replication_request_test.go
```diff
@@ -410,3 +410,77 @@ func TestMalformedReplicationResponse(t *testing.T) {
...
+		ReplicationRequest: &common.ReplicationRequest{
+			Seqs:   requestedSeqs,
+			Rounds: requestedRounds,
+		},
+	}, nodes[1]))
+
+	msg := <-comm.in
+	resp := msg.VerifiedReplicationResponse
+
+	expected := make([]common.VerifiedQuorumRound, 0, conf.MaxRoundWindow+2)
```
yacovm: Great, one last thing - can we move the code that prepares all the stuff:

```
	// notarize a few rounds past the indexed blocks: rounds 12...16
	// not finalized
	numNotarized := uint64(5)
	notarized := make([]common.VerifiedQuorumRound, 0, numNotarized)
	for i := uint64(0); i < numNotarized; i++ {
		block, notarization := advanceRoundFromNotarization(t, e, bb)
		notarized = append(notarized, common.VerifiedQuorumRound{
			VerifiedBlock: block,
			Notarization:  notarization,
		})
	}

	// oversized, unsortedseqs seqs request: 0 ...2*MaxRoundWindow
	// truncation keeps seqs 0...9
	requestedSeqs := make([]uint64, 0, 2*conf.MaxRoundWindow)
	for i := 2 * conf.MaxRoundWindow; i > 0; i-- {
		requestedSeqs = append(requestedSeqs, i-1)
	}

	// oversized, unsorted rounds request: 4..24
	// truncation keeps rounds 4...13
	roundStart := numIndexed - 8 // 4
	requestedRounds := make([]uint64, 0, 2*conf.MaxRoundWindow)
	for i := roundStart + 2*conf.MaxRoundWindow; i > roundStart; i-- {
		requestedRounds = append(requestedRounds, i-1)
	}

```

and 

```

	for i := uint64(0); i < conf.MaxRoundWindow; i++ {
		expected = append(expected, common.VerifiedQuorumRound{
			VerifiedBlock: seqs[i].VerifiedBlock,
			Finalization:  &seqs[i].Finalization,
		})
	}
	expected = append(expected, notarized[:2]...)
```

to be before we initialize `e` ? 

This cleanly separates between the test content and the setup.

jadalsmail: Ok i had to change the way the notarized rounds were created because advanceRoundFromNotarization(t, e, bb) takes in e. but now the order is exactly like you are describing.

yacovm (approved):

PR #440, opened 2026-07-16
Replication response size cap

jadalsmail:
This PR introduces a size cap for the replication response messages so that avalanchego clients doesn't reject messages above the size limit. 
Also added unit tests to test for the size limit.


common/msg.go
```diff
@@ -367,6 +367,34 @@ func (q *VerifiedQuorumRound) GetRound() uint64 {
...
+// to account for encoding overhead (in avalanchego each item adds protobuf
+// field tags and length prefixes on top of the block, QC, and header bytes)
+// it is tens of bytes in practivce, 512 is a generous bound
+const quoromRoundSizeSlack = 512
+
+func (q *VerifiedQuorumRound) EstimateSize() (int, error) {
+	size := quoromRoundSizeSlack
+
+	if q.VerifiedBlock != nil {
+		blockBytes, err := q.VerifiedBlock.Bytes()
```
yacovm: Instead of `Bytes()` we can just have an instance method `Size` for all these sub-message types (Notarization, EmptyNotarization, etc.) and we can just call the `Size()` and it will recursively compute the size. 

We don't want  to serialize the block to bytes just to know the size of the block.

common/msg.go
```diff
@@ -367,6 +367,34 @@ func (q *VerifiedQuorumRound) GetRound() uint64 {
 	return 0
 }
 
+// quoromRoundSizeSlack is added to the estimated size of each VerifiedQuoromRound
+// to account for encoding overhead (in avalanchego each item adds protobuf
+// field tags and length prefixes on top of the block, QC, and header bytes)
+// it is tens of bytes in practivce, 512 is a generous bound
+const quoromRoundSizeSlack = 512
+
+func (q *VerifiedQuorumRound) EstimateSize() (int, error) {
```
yacovm: I don't think we should have this method return an error. Ideally we would be able to estimate the size of a message without an error. 

If we don't marshal the fields to bytes we can remove the error returned.

external.go
```diff
@@ -37,6 +38,21 @@ func (p *ParsedBlock) Bytes() ([]byte, error) {
 	}
 	return rawBlock.MarshalCanoto(), nil
 }
+func (p *ParsedBlock) Size() int {
```
yacovm:
```
type ParsedBlock struct {
	metadata.StateMachineBlock
	msm *metadata.StateMachine
}
```

Shouldn't the Size() just be a method of `StateMachineBlock` ? 

If `StateMachineBlock` implements `Size()` then since `ParsedBlock` embeds `StateMachineBlock`, we get it implicitly.

jadalsmail: But here Size() doesn't measure a stateMachineBlock, it measure the RawBlock encoding of one, which is defined in the root package, not in msm (where stateMachineBlock is).
We can't move the Size() method to the msm package, because it doesn't import canotoTag_RawBlock__InnerBlockBytes and canotoTag_RawBlock__Metadata, and msm can't import them because the root already imports msm. (see https://github.com/ava-labs/Simplex/blob/main/external.go#L10)

yacovm: but the raw block is just an `msm` reference which isn't serialized, and the block. 

So the `Size()` of the block should be an instance method of the block, in my opinion.

> We can't move the Size() method to the msm package, because it doesn't import canotoTag_RawBlock__InnerBlockBytes and canotoTag_RawBlock__Metadata,

I don't understand.

jadalsmail: ok, in order to do that I just moved the Bytes() and the Size() methods to msm/block.go and they are a method of StateMachineBlock, also moved the RawBlock type there as well. And updated the RawBlock calls to be metadata.RawBlock. 

What do you think? Does this work?

yacovm: I don't really care where the RawBlock lives, that's fine moving it there.

However, I'm not entirely convinced that using the canoto size cache is the right way to go, it seems to me very ad-hoc and overly dependent on canoto. 

I think maybe a better solution would be to pre-compute the size of the `StateMachineMetadata` by marshaling it once the `RawBlock` is instantiated, and then `Size()` can just return the cached version.

yacovm: This would mean that `Size` can return to be in `RawBlock` as initially.

jadalsmail: Ok now I understand. I removed the dependency on canoto entirely, now the size is precomputed when the block is first serialized in Bytes() and we return the cached version. we are serializing the whole block tho not only the StateMachineMetadata (because if we were only marshaling the StateMachineMetadata we would still need to combine with the inner block size and the canoto tags).

msm/block.go
```diff
@@ -120,3 +132,31 @@ func (smb *StateMachineBlock) SealingBlockInfo() *common.SealingBlockInfo {
 		PrevSealingBlockHash: smb.Metadata.SimplexEpochInfo.PrevSealingBlockHash,
 	}
 }
+
+func (smb *StateMachineBlock) Bytes() ([]byte, error) {
```
yacovm, line 134: This change technically works, but - do we really need the code to be like this?

The only reason the code works is because `ParsedBlock` embeds `StateMachineBlock`. 

`ParsedBlock` is what implements `VerifiedBlock`, right? It implements all methods of `VerifiedBlock`. 

If we're adding `Size()` to the interface of `VerifiedBlock` then it makes sense that `ParsedBlock` will implement `Size()` and not `StateMachineBlock`. 

I think that `size` can be in `ParsedBlock` and `Size()` can be there too, and we can have the lazy caching of `size` under a lock that can be a field in `ParsedBlock`, so that we can be invoked `Size()` concurrently.

yacovm: I realize that I initially asked to move `Size()` from `ParsedBlock` to `StateMachineBlock`, sorry about that 🤦 

However, the way the code was back then made it odd that we access the fields of StateMachineBlock in such a way: 

```
(&p.Metadata).CalculateCanotoCache()
metadataSize := (&p.Metadata).CachedCanotoSize()
```

So I thought it would make sense to have the instance method touch its own fields. 

However, now when I thought about it some more, I think it actually makes sense to have it in `ParsedBlock` and reduce the complexity of `StateMachineBlock`

jadalsmail: Ok that makes sense. Let me know what you think of it now.

common/api.go
```diff
@@ -119,6 +119,9 @@ type VerifiedBlock interface {
 	// Bytes returns a byte encoding of the block
 	Bytes() ([]byte, error)
 
+	//Size returns the number of the bytes encoding of the block
```
yacovm: // Size 

(make a space)

external.go
```diff
@@ -5,37 +5,19 @@ package simplex
...
-		}
-		innerBlockBytes = rawInnerBlock
-	}
-	rawBlock := &RawBlock{
-		Metadata:        p.Metadata,
-		InnerBlockBytes: innerBlockBytes,
-	}
-	return rawBlock.MarshalCanoto(), nil
+	// sizeLock guards size, so Size() can be invoked concurrently
+	sizeLock sync.Mutex
```
yacovm: it's fine to say it guards size, but we can just call it `lock`.

msm/block.go
```diff
@@ -120,3 +130,19 @@ func (smb *StateMachineBlock) SealingBlockInfo() *common.SealingBlockInfo {
...
+	var innerBlockBytes []byte
+	if smb.InnerBlock != nil {
+		rawInnerBlock, err := smb.InnerBlock.Bytes()
+		if err != nil {
+			return nil, err
+		}
+		innerBlockBytes = rawInnerBlock
+	}
+	rawBlock := &RawBlock{
```
yacovm, line 143: I know it sounds counter-intuitive, but this is racy. 

We should clone the metadata to avoid data races:

```
	rawBlock := &RawBlock{
		Metadata:        smb.Metadata.Clone(),
		InnerBlockBytes: innerBlockBytes,
	}
```

external.go
yacovm: I meant, we can call `sizeLock` - `lock` instead of `sizeLock`

simplex/epoch.go
```diff
@@ -3089,19 +3096,26 @@ func (e *Epoch) handleReplicationRequest(req *common.ReplicationRequest, from co
...
 		if e.lastBlock != nil && e.lastBlock.Finalization.Finalization.Seq > req.LatestFinalizedSeq {
-			response.LatestFinalizedSeq = &common.VerifiedQuorumRound{
+			latestFinalizedSeq := &common.VerifiedQuorumRound{
 				VerifiedBlock: e.lastBlock.VerifiedBlock,
 				Finalization:  &e.lastBlock.Finalization,
 			}
+			size := latestFinalizedSeq.Size()
+			response.LatestFinalizedSeq = latestFinalizedSeq
+			remainingBytes -= size
+
```
yacovm: At this point we have both `LatestFinalizedSeq` and `LatestRound` initialized to quorum rounds (blocks with extra stuff), but we never check what happens if both blocks together are too much.

external.go
```diff
@@ -69,3 +51,15 @@ func (p *ParsedBlock) Verify(ctx context.Context) (common.VerifiedBlock, error)
 	}
 	return p, nil
 }
+func (p *ParsedBlock) Size() int {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	if p.size == 0 {
+		bytes, err := p.Bytes()
+		if err != nil {
```
yacovm, line 62: Isn't it a problem that we're returning here 0? 

`p.Bytes()` returns an error when the inner block is not nil and cannot be serialized.

If we can't determine the size of the inner block then we shouldn't return 0, as it will be silently swallowed up in the addition of other blocks.

Looking at avalanchego source code, the `Bytes()` method doesn't return an error, so I think we can just add a TODO here mentioning https://github.com/ava-labs/Simplex/issues/465 and then once we do https://github.com/ava-labs/Simplex/issues/465 we just remove this `if err != nil` branch.

yacovm (approved):
LGTM but maybe @samliok can also take a look?

common/api.go
```diff
@@ -119,6 +119,9 @@ type VerifiedBlock interface {
 	// Bytes returns a byte encoding of the block
 	Bytes() ([]byte, error)
 
+	// Size returns the number of the bytes encoding of the block
```
samliok: nit: grammar

common/msg.go
```diff
@@ -56,15 +56,17 @@ type ToBeSignedEmptyVote struct {
...
+	bytes := make([]byte, emptyVoteLen)
 	binary.BigEndian.PutUint64(bytes[1:9], v.EmptyVoteMetadata.Epoch)
 	binary.BigEndian.PutUint64(bytes[9:17], v.EmptyVoteMetadata.Round)
 	return bytes
 }
 
 func (v *ToBeSignedEmptyVote) FromBytes(buff []byte) error {
-	if len(buff) != 17 {
+	if len(buff) != emptyVoteLen {
 		return fmt.Errorf("invalid buffer length, expected 17, got %d", len(buff))
```
samliok:
```suggestion
		return fmt.Errorf("invalid buffer length, expected %d, got %d", emptyVoteLen, len(buff))
```

simplex/epoch.go
```diff
@@ -31,6 +31,9 @@ const (
 	DefaultFinalizeVoteRebroadcastTimeout = 6 * time.Second
 	EmptyVoteTimeoutID                    = "rebroadcast_empty_vote"
 	maxItemCountPerRequest                = 10 // max number of rounds or sequences that fit in one replication request
+	// DefaultMaxReplicationResponseSize is the max size of a replication response. avalanchego rejects messages larger
+	// than 2 MiB. we cap at 80% (4/5) same as avalanchego (see utils/constants/network.go)
+	DefaultMaxReplicationResponseSize = 2 * 1024 * 1024 * 4 / 5
```
samliok: can we set a constant in the `avalanchego` package and use that constant here?

i.e. copy/paste the 2MiB constant `DefaultMaxMessageSize`

U may need to update the branch to see the avalanchego package

simplex/epoch.go
```diff
@@ -3089,19 +3096,28 @@ func (e *Epoch) handleReplicationRequest(req *common.ReplicationRequest, from co
 			zap.Uint64("max round window", e.MaxRoundWindow))
 		rounds = rounds[:e.MaxRoundWindow]
 	}
+	remainingBytes := e.MaxReplicationResponseSize
```
samliok, line 3187: no action: why is e.MaxReplicationResponseSize an int? I see its an int in avalanchego but still i feel like it shouldn't be. 🤷

simplex/epoch.go
```diff
@@ -3089,19 +3096,28 @@ func (e *Epoch) handleReplicationRequest(req *common.ReplicationRequest, from co
...
 	if req.LatestFinalizedSeq > 0 {
 		if e.lastBlock != nil && e.lastBlock.Finalization.Finalization.Seq > req.LatestFinalizedSeq {
-			response.LatestFinalizedSeq = &common.VerifiedQuorumRound{
+			latestFinalizedSeq := &common.VerifiedQuorumRound{
 				VerifiedBlock: e.lastBlock.VerifiedBlock,
 				Finalization:  &e.lastBlock.Finalization,
 			}
+			size := latestFinalizedSeq.Size()
+			response.LatestFinalizedSeq = latestFinalizedSeq
+			remainingBytes -= size
```
samliok, line 3205: can we log an error log if `size` is > remainingBytes and also return early?

samliok (approved):
approving, feel free to hit merge once the nitty comments are addressed

simplex/replication_request_test.go
```diff
@@ -558,6 +558,7 @@ func TestReplicationRequestSizeLimitedLatestFinalizedSeq(t *testing.T) {
 	ctx := context.Background()
 	conf, _, _ := testutil.DefaultTestNodeEpochConfig(t, nodes[0], comm, bb)
 	conf.ReplicationEnabled = true
+	conf.Logger.(*testutil.TestLogger).Silence()
```
yacovm, line 562: is this on purpose?

simplex/epoch.go
```diff
@@ -3192,16 +3193,29 @@ func (e *Epoch) handleReplicationRequest(req *common.ReplicationRequest, from co
 				Finalization:  &e.lastBlock.Finalization,
 			}
 			size := latestFinalizedSeq.Size()
+			if size > remainingBytes {
```
yacovm, line 3196: This can never happen right?

simplex/epoch.go
```diff
@@ -3192,16 +3193,29 @@ func (e *Epoch) handleReplicationRequest(req *common.ReplicationRequest, from co
...
 			remainingBytes -= size
-
 		}
 	}
 	if req.LatestRound > 0 {
 		latestRound := e.getLatestVerifiedQuorumRound()
 		if latestRound != nil && latestRound.GetRound() > req.LatestRound {
 			size := latestRound.Size()
-			if size <= remainingBytes || response.LatestFinalizedSeq == nil {
+			if size > remainingBytes {
```
yacovm: I don't agree to this change. 

This can certainly happen, why is this an error?

yacovm (changes requested):
Not sure I agree to logging an error if the latest round cannot be returned after exhausting the quota from the latest seq

simplex/epoch.go
samliok: yea i agree, we should log it as an error for the `req.LatestFinalizedSeq` because that should never happen. However, since we have reduced the size and potentially included the sequence from before this _can_ happen and shouldn't result in an error log

simplex/epoch.go
samliok: yea it shouldn't be possible, but i wanted to log an error here in case. Otherwise `remainingBytes` would be negative after including this finalized sequence which I thought was good to guard against.

simplex/replication_request_test.go
samliok: i assume it's because error log will break CI, maybe add a comment noting why this was done

jadalsmail: yes, i saw you guys do it in other tests, is it bad practice?

yacovm (approved):

PR #446, opened 2026-07-20
Move AvalancheGo helpers into avalanchego package

samliok:
In future PRs, i want to be able to send approval + auxiliary info messages. These are going to be defined in the `common/msg.go` file, and will require us to change a lot of the imports. It also means we will have a circular dependency, if we don't move the avalanchego helpers into a separate file.


common/msg.go
```diff
@@ -8,21 +8,28 @@ import (
 	"encoding/asn1"
 	"encoding/binary"
 	"fmt"
+
+	"github.com/ava-labs/simplex/avalanchego"
 )
 
 type Message struct {
```
samliok, line 15: just for visual organization

common/msg.go
```diff
@@ -376,3 +383,13 @@ type BlockDigestRequest struct {
 	Seq    uint64
 	Digest Digest
 }
+
+// VersionID is an identifier for applications that care about epoch changes.
+type VersionID uint32
```
samliok, line 388: VersionID will be included in messages sending Auxiliary info.

common/msg.go
```diff
@@ -376,3 +383,13 @@ type BlockDigestRequest struct {
...
+
+type ValidatorSetApproval struct {
```
samliok, line 390: ValidatorSetApproval will be sent in an approvals message

yacovm (approved):

PR #453, opened 2026-07-22
Fix retention term

yacovm:
When a finalization is written to the WAL, the garbage collecting WAL invokes RetentionTerm()
as part of its Append() operation.
However, the current implementation returned an error in case it was a finalization.

This commit fixes this by adding the finalization option to the RetentionTerm function.

### DIFF
--- a/common/encoding.go
+++ b/common/encoding.go
@@ -284,7 +284,7 @@ func (wrr *WALRetentionReader) RetentionTerm(entry []byte) (uint64, error) {
 	switch recordType {
 	case BlockRecordType:
 		return BlockRecordRetentionTerm(entry)
-	case NotarizationRecordType:
+	case NotarizationRecordType, FinalizationRecordType:
 		return QuorumRecordRetentionTerm(entry)
 	case EmptyNotarizationRecordType:
 		return QuorumRecordRetentionTerm(entry)
--- a/instance_test.go
+++ b/instance_test.go
@@ -386,151 +386,6 @@ func TestInstanceRestartAcrossEpochs(t *testing.T) {
 	waitForNumBlocks(t, storage, storage.NumBlocks()+2)
 }
 
-type noopComm struct{}
-
-func (noopComm) Broadcast(*common.Message)           {}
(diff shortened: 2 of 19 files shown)


common/encoding.go
```diff
@@ -284,7 +284,7 @@ func (wrr *WALRetentionReader) RetentionTerm(entry []byte) (uint64, error) {
 	switch recordType {
 	case BlockRecordType:
 		return BlockRecordRetentionTerm(entry)
-	case NotarizationRecordType:
```
samliok, line 287: why dont we just have one branch for all `QuorumRecordRetentionTerm(entry)` calls?

instance_test.go
```diff
@@ -386,6 +386,151 @@ func TestInstanceRestartAcrossEpochs(t *testing.T) {
 	waitForNumBlocks(t, storage, storage.NumBlocks()+2)
 }
 
+type noopComm struct{}
+
+func (noopComm) Broadcast(*common.Message)           {}
+func (noopComm) Send(*common.Message, common.NodeID) {}
```
samliok: i think we have testuitl.NoopComm already

instance_test.go
```diff
@@ -386,6 +386,151 @@ func TestInstanceRestartAcrossEpochs(t *testing.T) {
...
+
+func TestInstanceOutOfOrderFinalizationPersistedToWAL(t *testing.T) {
```
samliok: aren't we just testing the garbage collection WAL? Can't we replicate this behavior in a more simple manner without the instance?

common/encoding.go
yacovm: Because once we merge the canoto PR we won't have one branch: https://github.com/ava-labs/Simplex/pull/435/changes#diff-1393209f6936423d26252ab8fbe095edfcb077b497d3913b439e139dd17e5d49R369-R377

instance_test.go
yacovm: fixed

instance_test.go
yacovm: fixed

samliok (approved):

PR #454, opened 2026-07-23
Fix non-validator nil pointer panic

yacovm:


nonvalidator/non_validator_test.go
```diff
@@ -672,3 +672,38 @@ func advanceUntil(nv *NonValidator, epochs *testEpochs, msgQueue *messageQueue,
 		nv.AdvanceTime(startTime)
 	}
 }
+
+func TestPoC_C1NilBlockPanic(t *testing.T) {
```
samliok: this test name is confusing

samliok (approved):

nonvalidator/non_validator_test.go
yacovm: Fixed

PR #456, opened 2026-07-23
Ignore non-validators when receiving quorum rounds

yacovm:
This commit makes non-validators ignore quorum rounds sent from non-validators.

### DIFF
--- a/common/api.go
+++ b/common/api.go
@@ -173,6 +173,15 @@ func (nws Nodes) NodeIDs() []NodeID {
 	return nodes
 }
 
+func (nws Nodes) Contains(nodeID NodeID) bool {
+	for _, nw := range nws {
+		if bytes.Equal(nw.Id, nodeID) {
+			return true
+		}
+	}
+	return false
+}
+
 // Node is a struct that pairs a node ID with its weight and public key.
 type Node struct {
 	Id     NodeID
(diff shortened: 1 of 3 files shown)


nonvalidator/non_validator.go
```diff
@@ -484,6 +484,13 @@ func (n *NonValidator) processQuorumRound(qr *common.QuorumRound, from common.No
 		return nil
 	}
 
+	validators := n.highestEpochCollector.latestValidatorSetRetriever.Validators()
```
samliok: i think we'd rather have this check inside `epochDigestCounter`.
https://github.com/ava-labs/Simplex/blob/55853f3e0f69778a05a4be44cf34d7e834975cbb/nonvalidator/epochs.go#L163

 What's the issue with a non-validator sending us a valid quorum round?

yacovm: why would the non-validator do this? We only replicate from validators both in snowman and simplex.

samliok: sure, but i feel like this is a weird place to do this. Why not put it at the top level of `handleReplicationResponse` or even do something like we did with `handleMessage` in `epoch.go` where we blocked/allowed certain messages from non-validators.

Also, if we receive 100 quorum rounds we will run `validators.Contains` on all 100 of them which is unnecessary since `from` never changes.

yacovm: > Why not put it at the top level of handleReplicationResponse or even do something like we did with handleMessage in epoch.go where we blocked/allowed certain messages from non-validators.

Moved it to `handleReplicationResponse`.

> Also, if we receive 100 quorum rounds we will run validators.Contains on all 100 of them which is unnecessary since from never changes.

Yeah, agreed.

samliok (approved):

PR #457, opened 2026-07-23
Discard equivocated round when processing notarizations

yacovm:
When receiving a notarization, check if the block received beforehand corresponds to the received notarization, and if not, delete the round.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -483,13 +483,11 @@ func (e *Epoch) loadFinalizationRecord(r []byte) error {
 		e.Logger.Debug("Finalization already indexed, skipping restoration", zap.Uint64("Sequence", finalization.Finalization.Seq))
 		return nil
 	}
-
-	round, ok := e.rounds[finalization.Finalization.Round]
-	if !ok {
-		return fmt.Errorf("round not found for finalization")
+	err = e.storeFinalization(&finalization)
+	if err != nil {
+		return err
 	}
 	e.Logger.Info("Finalization Recovered From WAL", zap.Uint64("Round", finalization.Finalization.Round))
-	round.finalization = &finalization
 	return nil
 }
 
@@ -762,18 +760,15 @@ func (e *Epoch) handleFinalizationMessage(message *common.Finalization, from com
 		return nil
 	}
 
-	round, exists := e.rounds[message.Finalization.Round]
+	_, exists := e.rounds[message.Finalization.Round]
 	if !exists {
 		e.handleFinalizationForPendingOrFutureRound(message, message.Finalization.Round, nextSeqToCommit)
 		return nil
 	}
 
-	if round.finalization != nil {
-		e.Logger.Debug("Received finalization for an already finalized round", zap.Uint64("round", message.Finalization.Round))
+	if err := e.storeFinalization(message); err != nil {
 		return nil
 	}
-
-	round.finalization = message
 
 	return e.persistFinalization(*message)
 }
@@ -1204,10 +1199,10 @@ func (e *Epoch) maybeCollectFinalization(round *Round) error {
 		return nil
 	}
 
-	return e.assembleFinalization(round, finalizations)
+	return e.assembleFinalization(finalizations)
 }
 
-func (e *Epoch) assembleFinalization(round *Round, finalizationVotes []*common.FinalizeVote) error {
+func (e *Epoch) assembleFinalization(finalizationVotes []*common.FinalizeVote) error {
 	for _, vote := range finalizationVotes {
 		e.Logger.Debug("Collected a finalize vote from node", zap.Stringer("NodeID", vote.Signature.Signer), zap.Uint64("round", vote.Finalization.Round), zap.Uint64("seq", vote.Finalization.Seq))
 	}
@@ -1217,7 +1212,10 @@ func (e *Epoch) assembleFinalization(round *Round, finalizationVotes []*common.F
 		return err
 	}
 
-	round.finalization = &finalization
+	if err := e.storeFinalization(&finalization); err != nil {
+		return nil
+	}
+
 	return e.persistFinalization(finalization)
 }
 
@@ -1954,7 +1952,10 @@ func (e *Epoch) processFinalizedBlock(block common.Block, finalization *common.F
 			delete(e.rounds, round.num)
 			return e.processFinalizedBlock(block, finalization)
 		}
-		round.finalization = finalization
+		if err := e.storeFinalization(finalization); err != nil {
+			e.Logger.Error("Failed storing finalization", zap.Error(err))
+			return err
+		}
 		prevEpochRound := e.round
 		if err := e.indexFinalizations(round.num); err != nil {
 			e.Logger.Error("Failed to index finalization", zap.Error(err))
@@ -2182,8 +2183,10 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block common.Block, finaliz
 
 		// Store the verified block in rounds map so subsequent blocks can find it as a dependency
 		roundEntry := NewRound(verifiedBlock)
-		roundEntry.finalization = finalization
 		e.rounds[md.Round] = roundEntry
+		if err := e.storeFinalization(finalization); err != nil {
+			return md.Digest
+		}
 		e.Logger.Debug("Stored finalized replicated block in rounds map",
 			zap.Uint64("round", md.Round),
 			zap.Uint64("seq", md.Seq),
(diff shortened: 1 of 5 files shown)


simplex/epoch.go
```diff
@@ -365,15 +365,48 @@ func (e *Epoch) loadNotarizationRecord(r []byte) error {
...
+	}
+
+	roundNum := notarization.Vote.Round
+	round, exists := e.rounds[roundNum]
+	if !exists {
+		return fmt.Errorf("could not find round %d", roundNum)
+	}
+
+	if round.notarization != nil {
+		return fmt.Errorf("notarization for round %d already exists", roundNum)
```
samliok, line 400: how is this ever possible? we can't create two unique notarizations?

simplex/epoch.go
```diff
@@ -758,19 +789,16 @@ func (e *Epoch) handleFinalizationMessage(message *common.Finalization, from com
...
-	round, exists := e.rounds[message.Finalization.Round]
+	_, exists := e.rounds[message.Finalization.Round]
 	if !exists {
 		e.handleFinalizationForPendingOrFutureRound(message, message.Finalization.Round, nextSeqToCommit)
 		return nil
 	}
 
-	if round.finalization != nil {
-		e.Logger.Debug("Received finalization for an already finalized round", zap.Uint64("round", message.Finalization.Round))
+	if err := e.storeFinalization(message); err != nil {
```
samliok, line 812: if we receive a valid finalization, but the leader equivocated shouldn't we trigger replication?

simplex/epoch.go
```diff
@@ -2819,6 +2866,38 @@ func (e *Epoch) retrieveLastPersistedBlacklist() (common.Blacklist, bool) {
...
+		return fmt.Errorf("round %d has no block to associate finalization with", roundNum)
+	}
+
+	expectedBlockHeader := round.block.BlockHeader()
+	if !expectedBlockHeader.Equals(&finalization.Finalization.BlockHeader) {
+		e.Logger.Debug("Equivocation detected: finalization block header does not match round block header",
+			zap.Uint64("round", roundNum),
+			zap.Stringer("block header", &expectedBlockHeader),
+			zap.Stringer("finalized block header", &finalization.Finalization.BlockHeader))
+		delete(e.rounds, roundNum)
```
samliok, line 2930: hmm, I'm wondering if anything bad can happen if we delete the round. 

Maybe a couple things? For instance we could have already progressed to the next round and incremented `e.round`. This means we will have no entry in the rounds map, and we may have no entry in the empty rounds map so I wonder if any lookup breaks.

Also if we restart the node it will re-set the rounds map when we read from the wal. But I don't think this is an issue

simplex/epoch.go
yacovm: https://github.com/ava-labs/Simplex/pull/457/changes/4e2f6f64ce3d959f8d0e6e0a84eddf9d2c5e92d9

simplex/epoch.go
yacovm: we can't progress to the next round at the point where we delete the round because we discard the notarization.

yacovm: > Also if we restart the node it will re-set the rounds map when we read from the wal. But I don't think this is an issue

It will simply delete the round once more, but in order to have a higher round than that, it will need to replicate the round, no? so it will eventually overwrite the rounds map when it reloads the WAL.

But good point, I should make a UT that checks this.

yacovm: https://github.com/ava-labs/Simplex/pull/457/commits/cab5e29f24a521f1b1d78e21d343017eb981a1e3

simplex/epoch_test.go
```diff
@@ -451,6 +451,259 @@ func TestEpochIndexFinalization(t *testing.T) {
...
+			}
+		}
+	}, 5*time.Second, 50*time.Millisecond, "node did not reply to the replication request")
+
+	require.Len(t, resp.Data, 1)
+	gotBlock := resp.Data[0].VerifiedBlock
+	require.Equal(t, blockB.BlockHeader().Digest, gotBlock.BlockHeader().Digest)
+	require.NotEqual(t, blockA.BlockHeader().Digest, gotBlock.BlockHeader().Digest)
+}
+
```
samliok:
```suggestion

// TestEquivocatedBlockRecovery tests that when a Byzantine leader equivocates —
// sending block A to the node while the honest majority agrees on a different
// block B for the same round — the node never acts on the equivocated round it
// holds, recovers the authentic block B via replication, and after a restart
// only ever serves block B (never the equivocated block A that only it saw).
//
// The equivocation is exercised both when the network only notarized block B and
// when it finalized it.
func TestEquivocatedBlockRecovery(t *testing.T) {
	for _, test := range []struct {
		name string
		// injectEquivocation delivers the honest majority's evidence for block B
		// (a notarization or finalization) and asserts the node does not act on the
		// equivocated round it holds, since it never received block B.
		injectEquivocation func(t *testing.T, e *Epoch, blockB VerifiedBlock, from NodeID, wal *testutil.TestWAL, storage *testutil.InMemStorage)
	}{
		{
			name: "notarized",
			injectEquivocation: func(t *testing.T, e *Epoch, blockB VerifiedBlock, from NodeID, wal *testutil.TestWAL, storage *testutil.InMemStorage) {
				validators := e.Comm.Validators()
				quorum := Quorum(len(validators))
				sigAggr := e.SignatureAggregatorCreator(validators)
				notarizationB, err := testutil.NewNotarization(e.Logger, sigAggr, blockB, validators.NodeIDs()[:quorum])
				require.NoError(t, err)
				require.Equal(t, blockB.BlockHeader().Digest, notarizationB.Vote.BlockHeader.Digest)

				testutil.InjectTestNotarization(t, e, notarizationB, from)

				// The node must NOT advance the round: it never received block B, so it
				// cannot notarize the round it has (which holds block A). The equivocated
				// round is thrown away.
				require.Never(t, func() bool {
					return e.Metadata().Round > 0
				}, time.Second, 100*time.Millisecond)
				// The equivocated notarization is not persisted to the WAL (still just the
				// block record).
				wal.AssertWALSize(1)
			},
		},
		{
			name: "finalized",
			injectEquivocation: func(t *testing.T, e *Epoch, blockB VerifiedBlock, from NodeID, wal *testutil.TestWAL, storage *testutil.InMemStorage) {
				validators := e.Comm.Validators()
				quorum := Quorum(len(validators))
				sigAggr := e.SignatureAggregatorCreator(validators)
				finalizationB, _ := testutil.NewFinalizationRecord(t, sigAggr, blockB, validators.NodeIDs()[:quorum])
				require.Equal(t, blockB.BlockHeader().Digest, finalizationB.Finalization.BlockHeader.Digest)

				testutil.InjectTestFinalization(t, e, &finalizationB, from)

				// Block 0 should not commit because the node has never received block B.
				storage.EnsureNoBlockCommit(t, 0)
			},
		},
	} {
		t.Run(test.name, func(t *testing.T) {
			bb := testutil.NewTestBlockBuilder()
			nodes := []NodeID{{1}, {2}, {3}, {4}}

			recordingComm := &recordingComm{
				Communication:     testutil.NewNoopComm(nodes),
				BroadcastMessages: make(chan *Message, 100),
				SentMessages:      make(chan *Message, 100),
			}
			conf, wal, storage := testutil.DefaultTestNodeEpochConfig(t, nodes[1], recordingComm, bb)
			conf.ReplicationEnabled = true

			leader := LeaderForRound(nodes, 0)
			require.NotEqual(t, conf.ID, leader) // Ensure that the node is not the leader for the first round.

			e, err := NewEpoch(conf)
			require.NoError(t, err)
			require.NoError(t, e.Start())

			// (1) Byzantine leader equivocates and sends a block to the node.
			md := e.Metadata()
			_, ok := bb.BuildBlock(context.Background(), md, emptyBlacklist)
			require.True(t, ok)
			blockA := bb.GetBuiltBlock()

			voteA, err := testutil.NewTestVote(blockA, leader)
			require.NoError(t, err)
			require.NoError(t, e.HandleMessage(&Message{
				BlockMessage: &BlockMessage{
					Vote:  *voteA,
					Block: blockA,
				},
			}, leader))

			// (2) Ensure the node has written the block to the WAL.
			wal.AssertWALSize(1)

			// (3) The honest majority agrees on a *different* block B for the same round.
			blockB := testutil.NewTestBlock(blockA.BlockHeader().ProtocolMetadata, emptyBlacklist)
			blockB.Data = []byte("equivocated-block-B")
			blockB.ComputeDigest()
			// Ensure that the two blocks are different.
			require.NotEqual(t, blockA.BlockHeader().Digest, blockB.BlockHeader().Digest)

			// (4) Deliver the honest majority's evidence for block B and assert the node
			// does not act on the equivocated round it holds.
			test.injectEquivocation(t, e, blockB, nodes[2], wal, storage)

			// (5) Wait until we receive a request to replicate the authentic block, block B.
			for msg := range recordingComm.SentMessages {
				if msg.ReplicationRequest != nil {
					break
				}
			}

			// (6) Feed the node the replication response containing the correctly
			// finalized block B.
			validators := e.Comm.Validators()
			quorum := Quorum(len(validators))
			sigAggr := e.SignatureAggregatorCreator(validators)
			finalizationB, _ := testutil.NewFinalizationRecord(t, sigAggr, blockB, validators.NodeIDs()[:quorum])
			replicationResponse := &ReplicationResponse{
				Data: []QuorumRound{
					{
						Block:        blockB,
						Finalization: &finalizationB,
					},
				},
			}
			require.NoError(t, e.HandleMessage(&Message{
				ReplicationResponse: replicationResponse,
			}, nodes[2]))

			// (7) The node commits the block the network actually agreed on
			// (block B) at seq 0, not the equivocated block A.
			require.Equal(t, blockB, storage.WaitForBlockCommit(0))
			require.Equal(t, uint64(1), storage.NumBlocks())

			// The rest of the test ensures that when the node restarts,
			// the node will have in its memory only the correct block and not the equivocated block.
			e.Stop()
			e, err = NewEpoch(conf)
			require.NoError(t, err)
			require.NoError(t, e.Start())
			t.Cleanup(e.Stop)

			// Drain the output channel
			for len(recordingComm.SentMessages) > 0 {
				<-recordingComm.SentMessages
			}

			// Send a replication request to the restarted node and expect to get the
			// correct block B and not the equivocated block A.
			require.NoError(t, e.HandleMessage(&Message{
				ReplicationRequest: &ReplicationRequest{
					Seqs: []uint64{0},
				},
			}, nodes[2]))

			// The node should reply with the block the network finalized (block B), never the
			// equivocated block A that only it ever saw.
			var resp *VerifiedReplicationResponse
			require.Eventually(t, func() bool {
				for {
					select {
					case msg := <-recordingComm.SentMessages:
						if msg.VerifiedReplicationResponse != nil {
							resp = msg.VerifiedReplicationResponse
							return true
						}
					default:
						return false
					}
				}
			}, 5*time.Second, 50*time.Millisecond, "node did not reply to the replication request")

			require.Len(t, resp.Data, 1)
			gotBlock := resp.Data[0].VerifiedBlock
			require.Equal(t, blockB.BlockHeader().Digest, gotBlock.BlockHeader().Digest)
			require.NotEqual(t, blockA.BlockHeader().Digest, gotBlock.BlockHeader().Digest)
		})
	}
}
```

samliok (dismissed):

samliok (approved):

PR #459, opened 2026-07-25
Harden GC WAL error handling

yacovm:
- Properly handle error return values where the error is ignored
- Properly increment the cumulative size of the WAL when appending records.
- Add unit tests for edge cases


wal/gc.go
```diff
@@ -125,16 +127,17 @@ func (gcw *GarbageCollectedWAL) Append(payload []byte) error {
 			retentionTerm: rt,
 			size:          len(payload),
 		})
+	} else {
+		// We're still using the last WAL, so we need to update its size
+		gcw.wals[len(gcw.wals)-1].size += len(payload)
```
samliok, line 132: surprised we never caught this

samliok (approved):

wal/gc.go
yacovm: because we never used the GC WAL until now

PR #462, opened 2026-07-25
Prevent double start of simplex instance

yacovm:
Guard against starting an instance while it was already started.

### DIFF
--- a/instance.go
+++ b/instance.go
@@ -67,6 +67,7 @@ type timeAdvancer interface {
 type Instance struct {
 	Config       Config
 	lock         sync.Mutex
+	started      bool
 	cs           *CachedStorage
 	wal          *wal.GarbageCollectedWAL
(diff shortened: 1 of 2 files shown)


instance.go
```diff
@@ -91,6 +92,12 @@ func (i *Instance) Start(ctx context.Context) error {
 	i.lock.Lock()
 	defer i.lock.Unlock()
 
+	if i.started {
```
samliok, line 95: is it expected that if we call `Start`, `Stop`  and then `Start` again we should not be able to start?

yacovm: To be able to restart the instance we need to add a test that checks that it can be safely done. 

An alternative for restarting the instance is just stopping it, and re-creating it, and then starting it.

This commit just fixes the problem that it is unsafe to start the instance twice.

samliok (approved):

PR #463, opened 2026-07-25
Skip signature verification in PutApprovals when same PK

yacovm:
PutApprovals copies approvals from the current store to a given store.
It snapshots the approvals and then calls HandleApproval on each approval,
but it also performs a signature verification which is redundant, because the signature
was already checked the first time it was inserted into the current approval store.
    
This commit simply refactors the code to enable PutApprovals to skip the signature check if it has the same PK.

### DIFF
--- a/instance.go
+++ b/instance.go
@@ -67,6 +67,7 @@ type timeAdvancer interface {
 type Instance struct {
 	Config       Config
 	lock         sync.Mutex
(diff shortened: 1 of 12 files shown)


samliok (commented):
Can we pause on merging this for now? I have a few things touching the approval store and I think this will conflict

PR #471, opened 2026-07-28
Make node start the round after replication notarization

yacovm:
Currently, if a node replicates a notarization it will advance to the next round, but will not start it.

This commit makes sure it starts the round in case it has advanced past the replicated seqs/rounds, just as we do for finalizations.

Added a unit test for both notarization and finalization, that a node proposes a block if it's the leader of the round that is after the round it replicated.

### DIFF
--- a/simplex/replication_test.go
+++ b/simplex/replication_test.go
@@ -1803,3 +1803,138 @@ func TestReplicationRequestsWithinMaxRoundWindow(t *testing.T) {
 			"requested a sequence more than MaxRoundWindow ahead of the next sequence to commit")
 	}
 }
+
+func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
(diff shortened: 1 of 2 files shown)


simplex/replication_test.go
```diff
@@ -1803,3 +1803,138 @@ func TestReplicationRequestsWithinMaxRoundWindow(t *testing.T) {
 			"requested a sequence more than MaxRoundWindow ahead of the next sequence to commit")
 	}
 }
+
+func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
+	// This test makes sure that a node that replicates a notarization or a finalization,
```
samliok, line 1808: nit: is it not standard to add these comments above the function

samliok (approved):

simplex/replication_test.go
yacovm: No! That's Claude's standard, not ours!

PR #475, opened 2026-07-30
Don't restart non-validators that are behind

samliok:
- Checks if we have validated a higher epoch before transitioning to a validator
- Also cleans up the instance code and removes the `nodeRole` struct


instance.go
```diff
@@ -146,32 +131,30 @@ func (i *Instance) startNonValidator(epochNum uint64, validators common.Nodes) e
...
+	}
+	comm := newCommunication(i.Config.Sender, i.Config.Broadcaster, nodes)
 
 	epochAwareStorage := &EpochAwareStorage{
 		epoch:   epochNum,
 		Storage: i.Config.Storage,
 		onEpochChange: func(epoch uint64, validators common.Nodes) error {
-			height := i.Config.PlatformChain.GetCurrentHeight()
-			vdrs, err := i.Config.PlatformChain.GetValidatorSet(height)
+			// set the communication to the highest validator set, since this node is a non-validator and may be behind.
```
yacovm: is this basically a poor man's caching of the highest validator set from the P-chain?

instance.go
```diff
@@ -362,19 +344,27 @@ func (i *Instance) listenForEpochChanges() {
...
+	if i.isStopped() {
+		i.Config.Logger.Info("instance is already stopped, skipping epoch change")
+		return
+	}
+
 	var err error
-	switch epochChange.nodeRole {
-	case nonValidator:
+	switch {
+	case i.e != nil && i.nv != nil:
```
yacovm: we removed the role but can we at least put some more meaningful code than i.e != nil and i.nv != nil? 

Maybe:

amValidator := i.e != nil 
amNonValidator := i.nv != nil


?

instance.go
samliok: it uses the same mechanism as before, i just extracted it to a helper.

But yea, we should feed the non-validator the highest pchain validators. I think alternatively we could just pass in the platform chain interface into the non-validator. This could help if the non-validator was offline and the validator set diverged.

instance.go
samliok: updated

0_build.txt
yacovm, line 1: I am not sure this file belongs to the PR

instance.go
yacovm: Good point 👍  if the non-validator was offline and the validator set has drifted away, it may still be able to replicate the P-chain and get the latest validator set.

instance.go
```diff
@@ -197,16 +180,38 @@ func (i *Instance) createNonValidatorConfig(epochNum uint64, validators common.N
...
+	for {
+		select {
+		case i.epochChanges <- ec:
+			return
+		case <-i.stopCh:
+			return
+		default:
+			// The slot holds a stale epoch change: take it and keep the newer of the two.
+			select {
+			case pending := <-i.epochChanges:
```
yacovm: how can this happen?

instance_test.go
```diff
@@ -503,6 +850,46 @@ func TestInstanceDoubleStartFails(t *testing.T) {
 	require.ErrorContains(t, inst.Start(t.Context()), "instance already started")
 }
 
+// currentNonValidator returns the non-validator the instance is running, or nil if it
```
yacovm: this is a weird thing to do. Can't we achieve the same thing in a different way?

instance.go
```diff
@@ -197,16 +180,38 @@ func (i *Instance) createNonValidatorConfig(epochNum uint64, validators common.N
...
+	}
+
+	for {
+		select {
+		case i.epochChanges <- ec:
+			return
+		case <-i.stopCh:
+			return
+		default:
+			// The slot holds a stale epoch change: take it and keep the newer of the two.
```
yacovm: I don't understand why we can't just wait until the goroutine has picked up the epoch change from the channel? 

This way, we will never move on to higher epochs.

samliok: so we have a suppeerrr annoying deadlock here if we don't do something along these lines. 

Essentially the `index` from both `nonvalidator` and `epoch` run under their own lock. So for example, when we are indexing a sealing block as a non-validator the non-validator lock is being held. Then we try to send on this channel, but it will block until the consumer can process it. 

But say we have been indexing sealing blocks very fast into storage(i.e in the tests). There may be a scenario where the consumer goroutine(`listenForEpochChanges`), is already processing an epoch change. So we have sent an epoch change, and the channel is now full. 

Now imagine immediately after we index _another_ sealing block. The channel is now full, so our non-validator blocks. 

The goroutine that is processing an epoch change may call back into the non-validator and request the non-validator lock(ex. `HighestValidatedEpoch`). 

This means we are deadlocked because the `index` goroutine can only let go of the lock if the epoch change channel is empty, but in order for the epoch change channel to get empty it needs to acquire the lock.

instance_test.go
```diff
@@ -273,6 +274,352 @@ func TestInstanceNonValidatorBootstraps(t *testing.T) {
...
+			if tt.wantValidator {
+				require.Eventually(t, func() bool {
+					return isRunningValidator(inst)
+				}, 20*time.Second, 100*time.Millisecond, "expected the node to restart as a validator")
+				require.Nil(t, runningNonValidator(inst))
+				return
+			}
+
+			// The epoch change is acted on by another goroutine, so the node has to be given the
+			// chance to restart before concluding that it did not.
```
yacovm: before concluding it did not what?

instance_test.go
```diff
@@ -273,6 +274,352 @@ func TestInstanceNonValidatorBootstraps(t *testing.T) {
...
+// the pointer over time detects one.
+func runningNonValidator(inst *Instance) *nonvalidator.NonValidator {
+	inst.lock.Lock()
+	defer inst.lock.Unlock()
+	return inst.nv
+}
+
+// sealingBlockAt builds the sealing block of epoch `epoch` at sequence `seq`
+func sealingBlockAt(seq, epoch uint64, validators metadata.NodeBLSMappings) *ParsedBlock {
+	md := common.ProtocolMetadata{Epoch: epoch, Round: seq, Seq: seq}
```
yacovm: what about the prev hash pointer? i guess this works because we replicate it and we don't check the prev pointer because we have a QC?

PR #479, opened 2026-08-03
Use the prev inner block hash of the last non simplex block

yacovm:
Previously to this commit, when building the first ever simplex block, simplex was using the previous block hash of the protocol metadata, which is computed differently and incorrectly because the block previous to the first ever simplex block is not a simplex block.

### DIFF
--- a/adapters.go
+++ b/adapters.go
@@ -35,8 +35,9 @@ func (c *Communication) Validators() common.Nodes {
 // Upon an epoch change, it will ignore blocks from previous epochs
 // and will call the onEpochChange callback when a new epoch is detected.
 type EpochAwareStorage struct {
-	msm           *metadata.StateMachine
-	onEpochChange func(seq uint64, validators common.Nodes) error
+	lastNonSimplexHeight uint64
+	msm                  *metadata.StateMachine
+	onEpochChange        func(seq uint64, validators common.Nodes) error
 	Storage
 	epoch uint64
 }
@@ -49,6 +50,7 @@ func (e *EpochAwareStorage) Retrieve(seq uint64) (common.VerifiedBlock, common.F
 	parsedBlock := &ParsedBlock{
 		msm:               e.msm,
 		StateMachineBlock: block,
+		legacyBlock:       seq <= e.lastNonSimplexHeight,
 	}
 	return parsedBlock, *finalization, nil
 }
--- a/msm/msm.go
+++ b/msm/msm.go
@@ -73,7 +73,7 @@ var (
 	errNilBlock                       = errors.New("block is nil")
 	errInvalidPChainHeight            = errors.New("invalid P-chain height")
 	errZeroBlockHasInnerBlock         = errors.New("zero block must not have an inner block")
-	errZeroBlockInnerDigestMismatch   = errors.New("zero block inner block digest does not match last non-Simplex inner block digest")
+	errZeroBlockPrevDigestMismatch    = errors.New("zero block previous digest does not match last non-Simplex inner block digest")
 	errZeroBlockTimestampMismatch     = errors.New("zero block timestamp does not match last non-Simplex inner block timestamp")
 	errPrevSealingBlockNotFinalized   = errors.New("previous sealing block is not finalized")
 	errBlockDigestMismatch            = errors.New("does not match proposed block digest")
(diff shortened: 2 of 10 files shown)


msm/msm.go
```diff
@@ -849,8 +848,10 @@ func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock Stat
 	if block.InnerBlock != nil {
 		return errZeroBlockHasInnerBlock
 	}
-	if prevBlock.InnerBlock.Digest() != sm.LastNonSimplexInnerBlock.Digest() {
```
samliok, line 852: how was this passing before? prevBlock.InnerBlock is supposed to be nil after the `if`

msm/msm_test.go
```diff
@@ -169,7 +169,7 @@ func TestMSMFirstSimplexBlockAfterPreSimplexBlocks(t *testing.T) {
 		Round: 0,
 		Seq:   43,
 		Epoch: 43,
-		Prev:  preSimplexParent.Digest(),
+		Prev:  preSimplexParent.InnerBlock.Digest(),
```
samliok, line 172: but we are calling `StateMachine.Digest` which encodes the inner blocks digest. So with this change this says the last non-simplex block is no longer wrapped in simplex protocol metadata?

adapters.go
```diff
@@ -49,6 +50,7 @@ func (e *EpochAwareStorage) Retrieve(seq uint64) (common.VerifiedBlock, common.F
 	parsedBlock := &ParsedBlock{
 		msm:               e.msm,
 		StateMachineBlock: block,
+		legacyBlock:       seq <= e.lastNonSimplexHeight,
```
samliok, line 53: wouldn't it be easier to just wrap any non-simplex blocks in the default protocol metadata?  this way we wouldn't need the legacy block fields and wouldn't need to use `inner.Digest`

msm/msm.go
yacovm: I think because `prevBlock.InnerBlock` is `sm.LastNonSimplexInnerBlock`

msm/msm_test.go
yacovm: we're building on top of the pre-simplex parent, so its digest is computed with the inner block.

The outer block doesn't exist for the pre-simplex parent, it's just a facade because we encode the ICMEpochInfo in the test outside of it.

adapters.go
yacovm: the protocol metadata doesn't contain the hash though, and the hash is computed differently for pre-simplex blocks.

msm/msm.go
```diff
@@ -984,8 +989,10 @@ func (sm *StateMachine) verifyBlockZero(block *StateMachineBlock, prevBlock Stat
...
-		return errZeroBlockInnerDigestMismatch
+
+	// The zero block must build upon the last non-Simplex block
+	if block.Metadata.SimplexProtocolMetadata.Prev != sm.LastNonSimplexInnerBlock.Digest() {
```
samliok, line 995: do we also verify the sequence is as expected?

adapters.go
```diff
@@ -108,16 +109,18 @@ func (cb *cachedBlock) Verify(ctx context.Context, verifyOpts ...common.VerifyOp
 }
 
 type CachedStorage struct {
-	msm  *metadata.StateMachine
-	lock sync.RWMutex
+	lastNonSimplexHeight uint64
```
samliok, line 114: can we add a comment?

adapters.go
```diff
@@ -151,11 +154,25 @@ func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.Veri
 
 	// We don't populate the cache here because we populate it externally.
 	block, finalization, err := cs.GetBlock(seq)
-	if digest != (common.Digest{}) && block.Digest() != digest {
-		return nil, nil, common.ErrBlockNotFound
+	if err != nil {
+		return nil, nil, err
+	}
+
+	legacy := seq <= cs.lastNonSimplexHeight
```
samliok, line 163: when are we going to need to retrieve something from storage < lastNonSimplexHeight? Aren't we only going to need to retrieve the lastNonSimplex block?

adapters.go
```diff
@@ -151,11 +154,25 @@ func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.Veri
...
+
+	if digest != (common.Digest{}) {
+		// A block predating Simplex is identified by its inner block's digest, not by a Simplex block digest.
+		blockDigest := block.Digest()
```
samliok, line 167: what would block.Digest() return on a legacy block?

it seems a bit weird to me that we cannot use a legacy blocks `Digest` method, yet it exposes a `Digest` method.

msm/msm.go
yacovm: We do it via line 974, because the expected block's digest depends on the sequence. 

We need this because we don't have a better way of ensuring the previous block points to the non-simplex block.

I can add an explicit check.

adapters.go
yacovm: We're not.

adapters.go
yacovm: added

samliok (approved):

PR #481, opened 2026-08-03
Close the WAL when it is no longer in use

samliok:
Closes the wal when we stop the validator


instance.go
```diff
@@ -429,6 +441,14 @@ func (i *Instance) createEpochConfig() (simplex.EpochConfig, error) {
 		return simplex.EpochConfig{}, err
 	}
 
+	// The epoch that used the previous WAL has already been stopped by now,
+	// so close it before replacing it, otherwise we leak the files it holds open.
+	// Failing to close it doesn't prevent the new epoch from running, so as with
+	// garbage collecting the WAL on an epoch change, we only log the error.
```
yacovm: I don't think this should be here. It should be in `stopValidator` instead. 

If we transition from a validator to a non-validator, we need to close the WAL, and putting the close here won't do it.

instance.go
```diff
@@ -250,6 +250,18 @@ func (i *Instance) Stop() {
...
+	}
+}
+
+// closeWAL closes the WAL currently in use, if any.
+// Must be called under the lock, and only once the epoch or non-validator using it has been stopped.
+func (i *Instance) closeWAL() error {
+	if i.wal == nil {
+		return nil
+	}
+	return i.wal.Close()
```
yacovm: Let's `i.wal = nil` after we close it.

instance.go
```diff
@@ -573,15 +588,17 @@ func (i *Instance) transitionEpochValidator(epochChange epochChange) error {
...
-	i.Config.WALs = nil
 	// On epoch change, garbage collect the WAL to remove all entries from previous epochs.
 	if err := i.wal.GarbageCollect(math.MaxUint64); err != nil {
 		i.Config.Logger.Error("Error garbage collecting epoch config on epoch change", zap.Error(err))
 	}
 
+	// Stop the epoch before doing anything else, so that we don't process any more messages while we are changing epochs.
+	i.stopValidator()
+
+	// Wipe out the WALs from the config so we won't try to load them again
```
yacovm, line 599: Wait, doesn't that effectively make us load an empty WAL next invocation of `createEpochConfig` ?

yacovm (changes requested):
I think we have a problem here when we need to create the GC WAL a second time, don't we? The value passed from the config is stale, but we still need something. 

I think we should just export `.wals` in `GarbageCollectedWAL` or alternatively, return the new WALs after GC:

```
GarbageCollect(retentionTerm uint64) ([]DeletableWAL ,error) 
```

PR #486, opened 2026-08-04
Do not write a second empty vote for a round we already timed out on

samliok:
`triggerEmptyBlockNotarization` is guarded only by `e.round > round`. It sets `emptyVotes.timedOut` but never reads it, even though `haveWeAlreadyTimedOutOnThisRound` exists, and three call sites reach it for the same round: a failed block verification, the blacklisted-proposer path in `buildBlock`, and `proposalWaitTimeExpired`.

Any two firing for one round append two empty vote records, which is what fails `TestLongRunningReplication` with `duplicate empty vote record for round 95`. The test times out on round 0, then delivers a proposal that fails verification, and asserts a second record never appears.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2680,6 +2680,14 @@ func (e *Epoch) triggerEmptyBlockNotarization(round uint64) {
 		return
 	}
 
+	// Several paths trigger the empty block agreement for the same round, so everything below,
+	// including the WAL append, must not run twice.
+	if e.haveWeAlreadyTimedOutOnThisRound(round) {
+		e.Logger.Debug("Not triggering empty block notarization because we already timed out on this round",
+			zap.Uint64("round", round))
+		return
+	}
+
 	emptyVote := common.ToBeSignedEmptyVote{EmptyVoteMetadata: common.EmptyVoteMetadata{
 		Round: round,
 		Epoch: e.Epoch,
--- a/simplex/epoch_failover_test.go
+++ b/simplex/epoch_failover_test.go
@@ -1376,10 +1376,8 @@ func TestNodeIsNotSuspectedAfterSigningALaterNotarization(t *testing.T) {
 	}
 }
 
-// countEmptyVoteRecords returns how many empty vote records the WAL holds for the round. It
-// returns an error rather than asserting, so it is safe to call from the condition goroutine
-// require.Never spawns.
-func countEmptyVoteRecords(wal *testutil.TestWAL, round uint64) (int, error) {
+// countEmptyVoteRecords returns how many empty vote records the WAL holds for the round.
+func countEmptyVoteRecords(wal *testutil.TestWAL) (int, error) {
 	rawRecords, err := wal.ReadAll()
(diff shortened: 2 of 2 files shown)


simplex/epoch.go
```diff
@@ -2680,6 +2680,14 @@ func (e *Epoch) triggerEmptyBlockNotarization(round uint64) {
 		return
 	}
 
+	// Several paths trigger the empty block agreement for the same round, so everything below,
```
yacovm, line 2691: production code LGTM, will look at test tomorrow

simplex/epoch_failover_test.go
```diff
@@ -1375,3 +1376,87 @@ func TestNodeIsNotSuspectedAfterSigningALaterNotarization(t *testing.T) {
...
+	require.NoError(t, err)
+
+	require.NoError(t, e.HandleMessage(&Message{
+		BlockMessage: &BlockMessage{
+			Vote:  *vote,
+			Block: block,
+		},
+	}, leader))
+
+	// The epoch appends to the WAL after Verify returns and after it re-acquires the lock, so
```
yacovm: no need to explain the internal locking stuff

simplex/epoch_failover_test.go
```diff
@@ -1375,3 +1376,87 @@ func TestNodeIsNotSuspectedAfterSigningALaterNotarization(t *testing.T) {
 			suspectIndex)
 	}
 }
+
+// countEmptyVoteRecords returns how many empty vote records the WAL holds for the round. It
+// returns an error rather than asserting, so it is safe to call from the condition goroutine
```
yacovm: I don't think require.Never spawns a goroutine and even if it doesn't it doesn't add any useful information to the test.

simplex/epoch_failover_test.go
```diff
@@ -1375,3 +1376,87 @@ func TestNodeIsNotSuspectedAfterSigningALaterNotarization(t *testing.T) {
...
+// require.Never spawns.
+func countEmptyVoteRecords(wal *testutil.TestWAL, round uint64) (int, error) {
```
yacovm: I think that in the test we only have a single round - round 0, so we don't actually need this `round` parameter here.

samliok: updated, although I think a better solution is to just have our test WAL throw an error whenever we append a duplicate record. Maybe for another pr

simplex/epoch_failover_test.go
```diff
@@ -1374,3 +1375,74 @@ func TestNodeIsNotSuspectedAfterSigningALaterNotarization(t *testing.T) {
...
+
+	require.NoError(t, e.HandleMessage(&Message{
+		BlockMessage: &BlockMessage{
+			Vote:  *vote,
+			Block: block,
+		},
+	}, nodes[0]))
+
+	// The epoch appends to the WAL after Verify returns
+	require.Never(t, func() bool {
```
yacovm: this can succeed if error is always != nil

yacovm (approved):

PR #490, opened 2026-08-04
Don't re-store a finalization the round already holds

samliok:
`processFinalizedBlock` stored the finalization unconditionally. When the round already held one, restored from the WAL or received in a finalization message, `storeFinalization` returned `round N already has a finalization`, that error reached `HandleMessage`, and the block was never indexed.

Store only when `round.finalization == nil`, then index as before.

`TestReplicationRedeliversRestoredFinalization` fails without the guard with `round 1 already has a finalization`.

This random long running test failed CI that caught the issue. [run 30950633369](https://github.com/ava-labs/Simplex/actions/runs/30950633369/job/92131525911) (subtest `#01`, seed `1785877451019`):

### DIFF
--- a/simplex/replication_test.go
+++ b/simplex/replication_test.go
@@ -1947,9 +1947,9 @@ func replicateSeq(block common.VerifiedFinalizedBlock) *common.Message {
 	}}
 }
 
-// TestReplicationRedeliversRestoredFinalization asserts that we can index a finalized block that already has
-// a finalization in the rounds map. This is possible if the WAL recovered the finalization or if it was sent via
-// a finalization message.
+// TestReplicationRedeliversRestoredFinalization asserts that a node commits a block when a replication
+// response redelivers a finalization the node already knows of. This is possible if the finalization was
+// recovered from the WAL or previously received in a finalization message.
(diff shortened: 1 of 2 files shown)


simplex/replication_test.go
```diff
@@ -1937,3 +1937,41 @@ func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
...
+	return &common.Message{ReplicationResponse: &common.ReplicationResponse{
+		Data: []common.QuorumRound{{
+			Block:        block.VerifiedBlock.(common.Block),
+			Finalization: &block.Finalization,
+		}},
+	}}
+}
+
+// TestReplicationRedeliversRestoredFinalization asserts that we can index a finalized block that already has
+// a finalization in the rounds map. This is possible if the WAL recovered the finalization or if it was sent via
```
yacovm: > asserts that we can index a finalized block that already has a finalization in the rounds map.

We should not reference internal data structures in tests, only behavior or scenarios.

yacovm (approved):

simplex/replication_test.go
samliok: updated by claude 1c0436e

PR #491, opened 2026-08-04
Index potential future finalizations when replication commits a block

samliok:
`createFinalizedBlockVerificationTask` indexed only the block it had just verified. A later round holding a future finalization, stored while its sequence was ahead of the next one to commit, was never revisited, so the node stalled while holding the block and finalization it was waiting for.

Index from the verified block's round forward with `indexFinalizations`, which stops as soon as a round is missing a finalization or is not next to commit.

`TestReplicationIndexesFutureFinalization` fails without this: seq 2 is never indexed.

Note: this and #490 both add the `replicateSeq` helper, so whichever merges second needs a small conflict fixup.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2236,8 +2236,6 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block common.Block, finaliz
 			zap.Uint64("seq", md.Seq),
 			zap.Stringer("digest", md.Digest))
 
-		// Sweep from this round, not just this block. Later rounds may already hold a future
-		// finalization, stored while its sequence was ahead of the next one to commit.
 		if err := e.indexFinalizations(md.Round); err != nil {
 			e.haltedError = err
 			e.Logger.Error("Failed to index finalization", zap.Error(err))
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2236,7 +2236,9 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block common.Block, finaliz
 			zap.Uint64("seq", md.Seq),
 			zap.Stringer("digest", md.Digest))
 
-		if err := e.indexFinalization(verifiedBlock, *finalization); err != nil {
+		// Sweep from this round, not just this block. Later rounds may already hold a future
+		// finalization, stored while its sequence was ahead of the next one to commit.
+		if err := e.indexFinalizations(md.Round); err != nil {
 			e.haltedError = err
 			e.Logger.Error("Failed to index finalization", zap.Error(err))
 			return md.Digest
--- a/simplex/replication_test.go
+++ b/simplex/replication_test.go
@@ -1965,7 +1965,7 @@ func setupFutureFinalization(t *testing.T, nodes []common.NodeID) (*simplex.Epoc
 	return e, storage, blocks
 }
 
-func replicateSeq(block common.VerifiedFinalizedBlock) *common.Message {
+func replicatedBlock(block common.VerifiedFinalizedBlock) *common.Message {
 	return &common.Message{ReplicationResponse: &common.ReplicationResponse{
 		Data: []common.QuorumRound{{
 			Block:        block.VerifiedBlock.(common.Block),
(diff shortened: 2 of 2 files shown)


simplex/epoch.go
```diff
@@ -2236,7 +2236,9 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block common.Block, finaliz
 			zap.Uint64("seq", md.Seq),
 			zap.Stringer("digest", md.Digest))
 
-		if err := e.indexFinalization(verifiedBlock, *finalization); err != nil {
+		// Sweep from this round, not just this block. Later rounds may already hold a future
```
yacovm: I don't think there is a need for a comment here, but all right.

simplex/replication_test.go
```diff
@@ -1937,3 +1937,66 @@ func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
...
+	e, err := simplex.NewEpoch(conf)
+	require.NoError(t, err)
+	t.Cleanup(e.Stop)
+	require.NoError(t, e.Start())
+	require.Equal(t, uint64(1), storage.NumBlocks())
+
+	return e, storage, blocks
+}
+
+func replicateSeq(block common.VerifiedFinalizedBlock) *common.Message {
```
yacovm: The method name is an action but it returns an object. Maybe `replicatedBlock` ?

simplex/epoch.go
samliok: removed

simplex/replication_test.go
samliok: updated

simplex/replication_test.go
```diff
@@ -1937,3 +1937,66 @@ func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
 		})
 	}
 }
+
+// setupFutureFinalization returns a node whose storage holds seq 0 and whose WAL holds the
+// block and finalization for round 2 (seq 2), which recovery restores into the rounds map
+// without indexing, since seq 2 is a future finalization. Only seq 1 is genuinely missing.
+func setupFutureFinalization(t *testing.T, nodes []common.NodeID) (*simplex.Epoch, *InMemStorage, []common.VerifiedFinalizedBlock) {
```
yacovm, line 1944: while this works, I think we should strive to create state with natural means instead of artificial means.

Can't we achieve the same state by simply injecting messages into the epoch?

simplex/replication_test.go
```diff
@@ -1937,3 +1937,66 @@ func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
...
+func replicatedBlock(block common.VerifiedFinalizedBlock) *common.Message {
+	return &common.Message{ReplicationResponse: &common.ReplicationResponse{
+		Data: []common.QuorumRound{{
+			Block:        block.VerifiedBlock.(common.Block),
+			Finalization: &block.Finalization,
+		}},
+	}}
+}
+
+// TestReplicationIndexesFutureFinalization asserts a node indexes a future finalization once nextSeqToCommit
```
yacovm, line 1977: we should not have reference to internal data structures in unit tests.

simplex/replication_test.go
```diff
@@ -1937,3 +1937,66 @@ func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
 		})
 	}
 }
+
+// setupFutureFinalization returns a node whose storage holds seq 0 and whose WAL holds the
+// block and finalization for round 2 (seq 2), which recovery restores into the rounds map
```
yacovm, line 1942: I think it's better to generate the `blocks` outside, because it's not clear which blocks we reference when evaluating: `blocks[1]), nodes[1])`

simplex/replication_test.go
```diff
@@ -1937,3 +1937,66 @@ func TestLeaderStartsRoundAfterReplicatedQuorumRound(t *testing.T) {
...
+	require.NoError(t, e.HandleMessage(replicatedBlock(blocks[1]), nodes[1]))
+	storage.WaitForBlockCommit(1)
+	require.Eventually(t, func() bool { return storage.NumBlocks() == 3 },
+		5*time.Second, 10*time.Millisecond,
+		"seq 2 never indexed, though the node holds its block and finalization")
+}
+
+// TestReplicationRedeliversStoredFinalization asserts a replication response with a
+// finalization the rounds map already holds is not an error.
+func TestReplicationRedeliversStoredFinalization(t *testing.T) {
```
yacovm, line 1993: why do we need this test? Can't we just move over:

```
	// a peer answering with seq 2, whose finalization round 2 already holds, must not fail
	require.NoError(t, e.HandleMessage(replicateSeq(blocks[2]), nodes[1]))
```

to the previous test `TestReplicationIndexesFutureFinalization` ?

yacovm (commented):
Fix LGTM but tests can use a little fix up.

samliok:
closed because this isn't actually possible in reality

because how can we have a future VERIFIED block + finalization in the rounds map? in order to verify a block we must have verified its child. if we verified its child then we must have stored it in the wal, and therefore `createFinalizedBlockVerificationTask` would never have been called

PR #492, opened 2026-08-05
Disable replication on Close

samliok:
Making this change means that if we call `Close` on replication state, we don't need to worry about calling `AdvanceTime` or frankly any of the other methods on it.

### DIFF
--- a/simplex/replication_state.go
+++ b/simplex/replication_state.go
@@ -6,6 +6,7 @@ package simplex
 import (
 	"math/rand/v2"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/ava-labs/simplex/common"
@@ -30,7 +31,7 @@ type finalizedQuorumRound struct {
 }
 
 type ReplicationState struct {
-	enabled  bool
+	enabled  *atomic.Bool
 	logger   common.Logger
 	myNodeID common.NodeID
 	rand     *rand.Rand // Random number generator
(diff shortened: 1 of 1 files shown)


yacovm:
I don't understand what goes wrong without this change. 

Is there any scenario that can be demonstrated that fails or gets stuck without this change?

samliok:
> I don't understand what goes wrong without this change.
> 
> Is there any scenario that can be demonstrated that fails or gets stuck without this change?

I was concerned that our shutdown for `non-validator` and `validator` could break depending on the ordering of the `Close` calls. Haven't written a test to show it breaking, but the potential issue is what if we close replication and then the epoch/non-validator is still processing replication messages. It will be trying to update the state of a closed struct, so maybe this would cause an ungraceful shutdown. 


```go
func (n *NonValidator) Stop() {
	n.Logger.Info("Shutting down non-validator", zap.Stringer("ID", n.ID))
	n.cancelCtx()
	n.sequenceReplicator.Close()
	n.verifier.Close()
}
```


```go
func (e *Epoch) Stop() {
	e.Logger.Info("Shutting down node")
	e.finishFn()
	e.monitor.Close()
	e.blockVerificationScheduler.Close()
	e.buildBlockScheduler.Close()
	e.timeoutHandler.Close()
	e.replicationState.Close()
}
```

simplex/replication_state.go
```diff
@@ -409,20 +409,21 @@ func (r *ReplicationState) DeleteRound(round uint64) {
...
-	if !r.enabled {
+	if !r.enabled.Load() {
 		return
 	}
 
 	r.digestTimeouts.Close()
 	r.emptyRoundTimeouts.Close()
 	r.roundRequestor.close()
 	r.finalizationRequestor.close()
+	r.enabled.Store(false)
```
yacovm, line 428: why not put this first?

PR #494, opened 2026-08-05
Add Golangci-lint job to CI

yacovm:


simplex/epoch.go
```diff
@@ -2590,12 +2590,16 @@ func (e *Epoch) createBlockBuildingTask(metadata common.ProtocolMetadata, blackl
...
 			case <-context.Done():
 			default:
-				e.Logger.Warn("Failed building block")
+				e.Logger.Debug("Failed building block")
 			}
 			return common.Digest{}
 		}
 
-		e.proposeBlock(block)
+		if err := e.proposeBlock(block); err != nil {
```
yacovm, line 2598: That's the main thing to notice in the PR

simplex/epoch_failover_test.go
```diff
@@ -449,7 +449,6 @@ func TestEpochLeaderFailoverInLeaderRound(t *testing.T) {
 		case msg := <-recordedMessages:
 			if msg.VerifiedBlockMessage != nil {
 				sentBlockMessage = true
```
yacovm, line 451: The loop condition exits the loop; a break here would only leave the select.

samliok (approved):

PR #496, opened 2026-08-06
Log genuine block build failures instead of treating them as cancellations

samliok:
In createBlockBuildingTask (simplex/epoch.go), the task called cancel() before the select that distinguishes a cancelled build from a failed one. cancel() closes the context's done channel synchronously, so the <-context.Done() case was always ready and the default branch logging "Failed building block" at Warn was unreachable.


simplex/epoch_test.go
```diff
@@ -2068,3 +2068,57 @@ func TestQuorum(t *testing.T) {
...
+}
+
+func (f *failingBlockBuilder) WaitForPendingBlock(ctx context.Context) {
+	<-ctx.Done()
+}
+
+// TestEpochWarnsOnBlockBuildFailure asserts that a BuildBlock failure with no
+// context cancellation is logged at Warn. createBlockBuildingTask used to call
+// cancel() before checking context.Done(), so the warn was never emitted.
+func TestEpochWarnsOnBlockBuildFailure(t *testing.T) {
```
samliok: i dont think we need a test for this

simplex/epoch.go
```diff
@@ -2585,11 +2585,12 @@ func (e *Epoch) createBlockBuildingTask(metadata common.ProtocolMetadata, blackl
 		e.lock.Lock()
 		defer e.lock.Unlock()
 
+		// Record cancellation before cancel() closes the context,
```
yacovm: we don't need to write this comment.

simplex/epoch_test.go
```diff
@@ -2068,3 +2068,57 @@ func TestQuorum(t *testing.T) {
 		})
 	}
 }
+
+// failingBlockBuilder fails every BuildBlock call without the context being cancelled.
```
yacovm: I think it's a waste of test. Let's just remove this.

simplex/epoch.go
```diff
@@ -2585,11 +2585,12 @@ func (e *Epoch) createBlockBuildingTask(metadata common.ProtocolMetadata, blackl
...
+		// otherwise a genuine build failure is indistinguishable from a cancelled one.
+		cancelled := context.Err() != nil
 		cancel()
 		if !ok {
-			select {
-			case <-context.Done():
-			default:
+			if !cancelled {
 				e.Logger.Warn("Failed building block")
```
samliok, line 2596: in the block builders we return false if the context was cancelled or an error was returned. if possible can we similarly log the error there so we don't lose visibility on why the block building failed

yacovm: we can just log the error in the block builder itself though. why does Simplex care why the block build failed?

simplex/epoch_test.go
samliok: updated 81820bb

simplex/epoch_test.go
samliok: updated 81820bb

simplex/epoch.go
samliok: updated ee7ec04

simplex/epoch.go
samliok: updated e6d02c1

adapters.go
```diff
@@ -217,6 +219,9 @@ func (bw *BlockBuilderWaiter) WaitForPendingBlock(ctx context.Context) {
 func (bw *BlockBuilderWaiter) BuildBlock(ctx context.Context, metadata common.ProtocolMetadata, blacklist common.Blacklist) (common.VerifiedBlock, bool) {
 	block, err := bw.msm.BuildBlock(ctx, metadata, blacklist)
 	if err != nil {
+		if ctx.Err() == nil {
```
yacovm, line 222: I don't think we should log this with a warning. 

The EVM can fail building a block from many reasons and we shouldn't spam the logs. 

If the EVM wants to spam the logs, it is welcome to, but we should log this in debug.

samliok: updated by claude 0b72ca3

PR #497, opened 2026-08-06
Serve verified uncommitted blocks to seq-only lookups in CachedStorage

samliok:
CachedStorage keeps verified-but-not-yet-finalized blocks in a cache keyed by digest. The msm sealing-block checks (`buildBlockOrTransitionEpoch`, `areWeReadyToTransitionEpoch`) look blocks up by sequence only and pass a zero digest, so they can never hit the cache. The lookup falls through to committed storage, and while the sealing block is notarized but not yet indexed, block building fails with `failed to retrieve sealing block for previous epoch`. The failed build is not retried, so the leader misses its round and the round empty-notarizes.

### DIFF
--- a/adapters.go
+++ b/adapters.go
@@ -113,26 +113,33 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
 
 func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.VerifiedBlock, *common.Finalization, error) {
 	cs.lock.RLock()
+
+	// A cached block is not finalized yet, since indexing removes it from the cache.
 	item, exists := cs.cache[digest]
 	if exists {
 		cs.lock.RUnlock()
-		// If the block is cached, it means it's not finalized yet, because upon finalizing the block (indexing)
-		// we also remove it from the cache. Therefore, we return nil for the finalization.
 		return item.ParsedBlock, nil, nil
+	}
+
+	if digest != (common.Digest{}) {
+		cs.lock.RUnlock()
+		return nil, nil, common.ErrBlockNotFound
+	}
+
+	for _, cb := range cs.cache {
+		if cb.BlockHeader().Seq == seq {
+			cs.lock.RUnlock()
+			return cb.ParsedBlock, nil, nil
+		}
 	}
 	cs.lock.RUnlock()
 
 	// We don't populate the cache here because we populate it externally.
-
 	block, finalization, err := cs.GetBlock(seq)
-	if err != nil {
-		return nil, nil, err
-	}
-
 	return &ParsedBlock{
 		StateMachineBlock: block,
 		msm:               cs.msm,
-	}, finalization, nil
+	}, finalization, err
 }
 
 func (cs *CachedStorage) Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error {
--- a/adapters.go
+++ b/adapters.go
@@ -236,8 +236,8 @@ func (bw *BlockBuilderWaiter) BuildBlock(ctx context.Context, metadata common.Pr
 }
 
 type blockDeserializer struct {
-	vm  VM
-	msm *metadata.StateMachine
+	vm VM
+	cs *CachedStorage
 }
 
 func (bd *blockDeserializer) DeserializeBlock(ctx context.Context, bytes []byte) (common.Block, error) {
@@ -250,11 +250,14 @@ func (bd *blockDeserializer) DeserializeBlock(ctx context.Context, bytes []byte)
 	if err != nil {
 		return nil, err
 	}
-	return &ParsedBlock{
-		StateMachineBlock: metadata.StateMachineBlock{
-			InnerBlock: block,
-			Metadata:   rawBlock.Metadata,
+	return &cachedBlock{
+		ParsedBlock: &ParsedBlock{
+			StateMachineBlock: metadata.StateMachineBlock{
+				InnerBlock: block,
+				Metadata:   rawBlock.Metadata,
+			},
+			msm: bd.cs.msm,
 		},
-		msm: bd.msm,
+		cache: bd.cs,
 	}, nil
 }
--- a/adapters.go
+++ b/adapters.go
@@ -60,7 +60,9 @@ func (e *EpochAwareStorage) Index(ctx context.Context, block common.VerifiedBloc
 		// This is a Telock from a previous epoch, so we ignore it and do not index it.
 		return nil
 	}
-	if err := e.Storage.Index(ctx, block, certificate); err != nil {
+	// e.Storage would resolve to the raw storage promoted through CachedStorage,
+	// skipping the cache eviction in CachedStorage.Index.
+	if err := e.CachedStorage.Index(ctx, block, certificate); err != nil {
 		return err
 	}
 	// This is a sealing block, and it is not the zero block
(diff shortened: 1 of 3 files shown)


adapters.go
```diff
@@ -114,6 +114,16 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
 func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.VerifiedBlock, *common.Finalization, error) {
 	cs.lock.RLock()
 	item, exists := cs.cache[digest]
+	if !exists && digest == (common.Digest{}) {
+		// Seq-only lookups pass a zero digest, so scan the cache by seq.
+		// Otherwise a verified but not yet finalized block is invisible to them.
+		for _, cb := range cs.cache {
```
yacovm: can we iterate with index instead of this?

yacovm:
```
	if !exists && digest == (common.Digest{}) {
		// Seq-only lookups pass a zero digest, so scan the cache by seq.
		// Otherwise a verified but not yet finalized block is invisible to them.
		for i := range cs.cache {
			cb := cs.cache[i].ParsedBlock
			md := &cb.Metadata
			protocolMD := &md.SimplexProtocolMetadata
			if protocolMD.Seq == seq {
				item, exists = cs.cache[i], true
				break
			}
		}
	}
```

samliok: updated by claude 76738bc

adapters.go
```diff
@@ -114,6 +114,20 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
 func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.VerifiedBlock, *common.Finalization, error) {
 	cs.lock.RLock()
 	item, exists := cs.cache[digest]
+	if !exists && digest == (common.Digest{}) {
```
yacovm: Let's add a test that tests the added lines.

samliok: this should be 

	if !exists || digest == (common.Digest{}) {

samliok: updated by claude abf52e5

samliok: updated by claude c9de1c6

adapters.go
```diff
@@ -113,22 +113,34 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
 
 func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.VerifiedBlock, *common.Finalization, error) {
 	cs.lock.RLock()
+	defer cs.lock.RUnlock()
+
 	item, exists := cs.cache[digest]
 	if exists {
-		cs.lock.RUnlock()
```
yacovm, line 120: why the locking change?

adapters.go
```diff
@@ -113,22 +113,34 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
...
 	// We don't populate the cache here because we populate it externally.
-
 	block, finalization, err := cs.GetBlock(seq)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// If we are requesting an empty digest, and the sequence is not found in storage,
+	// check if we have verified but not indexed this sequence
+	if err == common.ErrBlockNotFound && digest == (common.Digest{}) {
```
yacovm: shouldn't this be: `f err == common.ErrBlockNotFound || digest == (common.Digest{})` ?

samliok: no in this case it should be &&.

because || would mean we grab and return from the cache even if we returned a block & finalization above. this shouldn't happen because if we finalize a block, it shouldn't be in the cache but still ranging through the cache seems unnecessary.

yacovm: After reading the code again - I just noticed that we do the check after we check the storage with `cs.GetBlock(seq)` in line 126. 

Why do we do that? Shouldn't we iterate the cache *before* we ask the storage? 

What's the point of asking the storage first?

yacovm: Can't we do this?


```
func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.VerifiedBlock, *common.Finalization, error) {
	cs.lock.RLock()
	defer cs.lock.RUnlock()

	// A cached block is not finalized yet, since indexing removes it from the cache.
	if item, exists := cs.cache[digest]; exists {
		return item.ParsedBlock, nil, nil
	}

	for _, cb := range cs.cache {
		if cb.BlockHeader().Seq == seq && digest == (common.Digest{}) {
			return cb.ParsedBlock, nil, nil
		}
	}

	// We don't populate the cache here because we populate it externally.
	block, finalization, err := cs.GetBlock(seq)
	if err == nil {
		return &ParsedBlock{
			StateMachineBlock: block,
			msm:               cs.msm,
		}, finalization, nil
	}

	return nil, nil, err
}
```

samliok: yep, i updated the logic + re-added the locking stuff from before

adapters.go
```diff
@@ -113,26 +113,33 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
 
 func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.VerifiedBlock, *common.Finalization, error) {
 	cs.lock.RLock()
+
+	// A cached block is not finalized yet, since indexing removes it from the cache.
```
yacovm: I am not sure this comment is beneficial. 

Yes, indexing removes the cached block from the cache, but how is that relevant to retrieval?

adapters.go
```diff
@@ -113,26 +113,33 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
...
 	if exists {
 		cs.lock.RUnlock()
-		// If the block is cached, it means it's not finalized yet, because upon finalizing the block (indexing)
-		// we also remove it from the cache. Therefore, we return nil for the finalization.
 		return item.ParsedBlock, nil, nil
 	}
-	cs.lock.RUnlock()
 
-	// We don't populate the cache here because we populate it externally.
+	if digest != (common.Digest{}) {
```
yacovm: why do we return here?

adapters.go
```diff
@@ -113,26 +113,33 @@ func (cs *CachedStorage) RetrieveBlock(seq uint64, digest common.Digest) (metada
...
+		cs.lock.RUnlock()
+		return nil, nil, common.ErrBlockNotFound
+	}
 
-	block, finalization, err := cs.GetBlock(seq)
-	if err != nil {
-		return nil, nil, err
+	for _, cb := range cs.cache {
+		if cb.BlockHeader().Seq == seq {
```
yacovm, line 127: we should only return if the digest either matches or the digest we requested to match is empty.

adapters_test.go
```diff
@@ -0,0 +1,151 @@
...
+				Payload: []byte(payload),
+			},
+		},
+	}
+}
+
+// TestCachedStorageRetrieveBySeq asserts that Retrieve serves a verified but
+// not yet indexed block when the caller passes a zero digest or a digest that
+// misses the cache, by matching on seq alone.
+func TestCachedStorageRetrieveBySeq(t *testing.T) {
```
yacovm: I think we should add tests that cover the corner cases I pointed out above

adapters.go
samliok: i guess its relevant because we don't return a finalization if we get a cache hit. i can remove the comment though

adapters.go
samliok: fixed, also fixed after cs.GetBlock() is returned

adapters.go
samliok: removed

adapters_test.go
samliok: add some nice table tests

yacovm (approved):

PR #499, opened 2026-08-08
Skip verifying MSM when replicating blocks

yacovm:
This commit adds a VerifyOptions to the internal API of Verify() in order for replicated blocks to not verify the MSM because there is no need to verify something that has a quorum certificate.

### DIFF
--- a/common/api.go
+++ b/common/api.go
@@ -91,7 +91,7 @@ type Block interface {
 	Blacklist() Blacklist
 
 	// Verify verifies the block by speculatively executing it on top of its ancestor.
-	Verify(ctx context.Context) (VerifiedBlock, error)
+	Verify(context.Context, ...VerifyOptions) (VerifiedBlock, error)
 
 	// non nil only for sealing blocks & first ever simplex block
 	SealingBlockInfo() *SealingBlockInfo
@@ -212,3 +212,13 @@ func F(n int) int {
 
 // SignatureAggregatorCreator creates a SignatureAggregator from a list of nodes and their weights.
 type SignatureAggregatorCreator func([]Node) SignatureAggregator
+
+type VerifyOptions func(*VerifyOpts)
+
+func OnlyVMVerifyOpt(vo *VerifyOpts) {
+	vo.OnlyVerifyVM = true
+}
+
+type VerifyOpts struct {
+	OnlyVerifyVM bool
+}
--- a/common/api.go
+++ b/common/api.go
@@ -213,12 +213,12 @@ func F(n int) int {
 // SignatureAggregatorCreator creates a SignatureAggregator from a list of nodes and their weights.
 type SignatureAggregatorCreator func([]Node) SignatureAggregator
 
-type VerifyOptions func(*VerifyOpts)
+type VerifyOptions func(*VerifyConfig)
 
-func OnlyVMVerifyOpt(vo *VerifyOpts) {
-	vo.OnlyVerifyVM = true
+func OnlyVMVerifyOpt(vc *VerifyConfig) {
+	vc.OnlyVM = true
 }
 
-type VerifyOpts struct {
-	OnlyVerifyVM bool
+type VerifyConfig struct {
+	OnlyVM bool
 }
--- a/external.go
+++ b/external.go
@@ -48,14 +48,14 @@ func (p *ParsedBlock) Blacklist() common.Blacklist {
 }
 
 func (p *ParsedBlock) Verify(ctx context.Context, verifyOpts ...common.VerifyOptions) (common.VerifiedBlock, error) {
-	var verifyOptions common.VerifyOpts
+	var verifyOptions common.VerifyConfig
 	for _, opt := range verifyOpts {
 		opt(&verifyOptions)
 	}
 
 	// When replicating a block that carries a notarization or a finalization, the quorum certificate
 	// already attests to the state machine transition, so we only verify the inner block.
-	if verifyOptions.OnlyVerifyVM {
+	if verifyOptions.OnlyVM {
 		if p.InnerBlock == nil {
 			return p, nil
 		}
--- a/external.go
+++ b/external.go
@@ -47,7 +47,21 @@ func (p *ParsedBlock) Blacklist() common.Blacklist {
 	return p.Metadata.SimplexBlacklist.Clone()
 }
 
-func (p *ParsedBlock) Verify(ctx context.Context) (common.VerifiedBlock, error) {
+func (p *ParsedBlock) Verify(ctx context.Context, verifyOpts ...common.VerifyOptions) (common.VerifiedBlock, error) {
+	var verifyOptions common.VerifyOpts
+	for _, opt := range verifyOpts {
+		opt(&verifyOptions)
+	}
+
+	// When replicating a block that carries a notarization or a finalization, the quorum certificate
+	// already attests to the state machine transition, so we only verify the inner block.
+	if verifyOptions.OnlyVerifyVM {
+		if p.InnerBlock == nil {
+			return p, nil
+		}
+		return p, p.InnerBlock.Verify(ctx, p.Metadata.ICMEpochInfo.PChainEpochHeight)
+	}
+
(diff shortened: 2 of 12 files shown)


simplex/util.go
```diff
@@ -161,7 +161,7 @@ func (block *oneTimeVerifiedBlock) Verify(ctx context.Context) (common.VerifiedB
 		return result.vb, result.err
```
samliok, line 161: so i think this could break. Imagine we failed verification due to not having the state diffs the first time we receive a block, then we try and verify again(this time the msm verification should be disabled) however we will still return an error due to the cache.

I think this is worth writing a test for?

external.go
```diff
@@ -47,7 +47,21 @@ func (p *ParsedBlock) Blacklist() common.Blacklist {
...
+		opt(&verifyOptions)
+	}
+
+	// When replicating a block that carries a notarization or a finalization, the quorum certificate
+	// already attests to the state machine transition, so we only verify the inner block.
+	if verifyOptions.OnlyVerifyVM {
+		if p.InnerBlock == nil {
+			return p, nil
+		}
+		return p, p.InnerBlock.Verify(ctx, p.Metadata.ICMEpochInfo.PChainEpochHeight)
```
samliok, line 62: should we return `nil` instead of `p` if p.InnerBlock.Verify() fails?

instance_test.go
```diff
@@ -505,27 +493,29 @@ func TestInstanceZeroBlockUsesLastNonSimplexPChainHeight(t *testing.T) {
...
-	zeroBlock, ok := storage.blockAt(1)
-	require.True(t, ok)
-	require.Equal(t, metadata.BlockTypeZero, zeroBlock.Type())
-	require.Equal(t, basePChainHeight, zeroBlock.Metadata.PChainHeight)
-	require.Equal(t, basePChainHeight, zeroBlock.Metadata.SimplexEpochInfo.PChainReferenceHeight)
+	require.ErrorContains(t, inst.Start(t.Context()), "instance already started")
 }
 
-func TestInstanceDoubleStartFails(t *testing.T) {
+func TestNonValidatorSkipsMSMVerification(t *testing.T) {
```
samliok, line 552: why is this test in the `Instance` code? I think it would make more sense to write the test closer to the non-validator? The instance tests should just be about transitioning between non-validator & validator, rather than the internals of how verification works if you are a validator/non-validator.

Thoughts?

instance_test.go
```diff
@@ -539,14 +529,249 @@ func TestInstanceDoubleStartFails(t *testing.T) {
...
+
+	nonValidatorInstance := newInstance(t, nonValidatorNodeID, nonValidatorStorage, newInMemNetwork(t), pChain, cops, genesisBlock)
+	require.NoError(t, nonValidatorInstance.Start(t.Context()))
+	t.Cleanup(nonValidatorInstance.Stop)
+
+	// The MSM that built the chain - the validator has stopped, so nothing else uses it - accepts
+	// the block, so we know the block is valid.
+	msm := validatorInstance.msm
+	_, err := (&ParsedBlock{StateMachineBlock: replicated.Clone(), msm: msm}).Verify(context.Background())
+	require.NoError(t, err)
```
samliok: why do we do this verification and error checking?

instance_test.go
```diff
@@ -539,14 +529,249 @@ func TestInstanceDoubleStartFails(t *testing.T) {
...
+	require.True(t, ok)
+	require.Equal(t, tampered.Digest(), committed.Digest())
+
+	// ... so the two ledgers are the same height but disagree on their last block: the
+	// non-validator committed a block that exists nowhere in the validator's storage.
+	require.Equal(t, storage.NumBlocks(), nonValidatorStorage.NumBlocks())
+	require.NotEqual(t, replicated.Digest(), committed.Digest())
+}
+
+func TestValidatorSkipsMSMVerificationWhenReplicating(t *testing.T) {
```
samliok, line 660: my other concern for these tests is that they are way to long and you need to keep track of so many components to write them. I feel like writing a test like this is going to be pretty hard without the help of claude.

simplex/epoch.go
```diff
@@ -2205,7 +2201,7 @@ func (e *Epoch) createFinalizedBlockVerificationTask(block common.Block, finaliz
 			return md.Digest
 		}
 
-		verifiedBlock, err := block.Verify(context.Background())
+		verifiedBlock, err := block.Verify(context.Background(), common.OnlyVMVerifyOpt)
```
samliok, line 2208: especially since this is a change in the epoch code, i don't think we need to create a large instance test for this.

common/api.go
```diff
@@ -212,3 +212,13 @@ func F(n int) int {
...
+
+type VerifyOptions func(*VerifyOpts)
+
+func OnlyVMVerifyOpt(vo *VerifyOpts) {
+	vo.OnlyVerifyVM = true
+}
+
+type VerifyOpts struct {
+	OnlyVerifyVM bool
+}
```
samliok: thoughts on these names?

```suggestion
type VerifyOption func(*VerifyConfig)

type VerifyConfig struct {
    VMOnly bool
}

// OnlyVerifyVM skips MSM verification and verifies only the inner VM block.
// Used when the block is already backed by a quorum certificate.
func OnlyVerifyVM(c *VerifyConfig) {
    c.VMOnly = true
}
```

samliok: or we can make it more clear and have the option be `SkipsMSM`. The issue with that is now the epoch needs to know about the MSM, but I guess its the same as now because the epoch needs to know the block has an inner VM block which is also weird 🤷

external.go
yacovm: In Golang, if the error isn't nil, you shouldn't look at the other returned argument.

simplex/util.go
yacovm: Interesting idea. 

Luckily, we only cache if the block verification succeeded. 

I think it might be worth to write a test regardless, but as a separate PR once we move using the sealing block for the validator set instead of the P-chain state for the current validator set.

instance_test.go
yacovm: The wiring is in this package though. The non-validator package just invokes the API blindly. What if the implementation of the API is wrong? The implementation is in this package, not in the non-validator package.

The test cannot be in a package that is beneath where the implementation is.

yacovm: > The instance tests should just be about transitioning between non-validator & validator

I agree that this test is exercising a lot more than just checking the implementation of how the wiring is done, however the tests prove without any doubt that the implementation is correct.

instance_test.go
yacovm: Because we later clone , temper it, it and verify it again and see that it fails verifying. 

I wanted to show that before tempering with it, it verifies successfully. I can remove this though.

instance_test.go
yacovm: The reason the tests take a long time is because the block production in the tests is based on time and not on explicit block counting. 

`waitForNumBlocks` just waits for blocks to be produced. 

We can just change the implementation of the `testVM` to receive instructions from the test instead of relying on time:

```
func (vm *testVM) WaitForPendingBlock(ctx context.Context) {
	if vm.paused.Load() {
		<-ctx.Done() // no pending block while paused
		return
	}
	select {
	case <-ctx.Done():
	case <-time.After(100 * time.Millisecond):
	}
}
```

If we do it, the tests will then take milliseconds only.

As for the complexity - I agree the tests are complex to write, but we can refactor later and make common initialization code.

simplex/epoch.go
yacovm: Just verifying in the epoch code that we passed the correct flag doesn't prove anything, as the implementation is not in the epoch code but in a higher layer.

common/api.go
yacovm: https://github.com/ava-labs/Simplex/pull/499/commits/2bd80b5851178af1a3074b9b246778c32963fc6c

instance_test.go
yacovm: https://github.com/ava-labs/Simplex/pull/499/commits/5c667f712f668290972842bb834e18c98b32ccd6

samliok (approved):

PR #503, opened 2026-08-13
Stop indexing and gate messages once the epoch is sealed

samliok:
Split out of the cleanup-instance branch.

Two related fixes for behavior after an epoch seals:
- `HandleMessage` now treats every sender like a non-validator once the epoch is sealed, so only replication requests are served.
- `indexFinalization` no longer indexes blocks after the epoch is sealed, though it still tracks the last block and progresses rounds.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -166,7 +166,7 @@ func (e *Epoch) HandleMessage(msg *common.Message, from common.NodeID) error {
 	}
 
 	_, known := e.validatorsToPKs[string(from)]
-	if !known {
+	if !known || e.isEpochSealed() {
 		e.Logger.Debug("Received message from a non-validator node", zap.Stringer("nodeID", from))
 		switch {
 		case msg.ReplicationRequest != nil && e.ReplicationEnabled:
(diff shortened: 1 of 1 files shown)


simplex/epoch.go
```diff
@@ -1466,13 +1466,17 @@ func (e *Epoch) indexFinalizations(startRound uint64) error {
 }
 
 func (e *Epoch) indexFinalization(block common.VerifiedBlock, finalization common.Finalization) error {
-	if err := e.Storage.Index(e.finishCtx, block, finalization); err != nil {
-		return err
+	// index only if the epoch is not sealed
+	if !e.isEpochSealed() {
```
yacovm: but don't we already enforce that at the instance level? Why do this here too?

simplex/epoch.go
```diff
@@ -166,7 +166,7 @@ func (e *Epoch) HandleMessage(msg *common.Message, from common.NodeID) error {
 	}
 
 	_, known := e.validatorsToPKs[string(from)]
```
yacovm, line 168: I think we should enforce this externally, not internally.

simplex/epoch.go
samliok: yea agreed it can be removed

simplex/epoch.go
samliok: agreed, should this be done in a separate pr though?

yacovm: but if we do that in a separate PR then we don't do it here...

samliok:
closed with https://github.com/ava-labs/Simplex/pull/513

PR #504, opened 2026-08-13
Find verified unfinalized blocks on seq-only cache lookups

samliok:
Split out of the cleanup-instance branch.

`CachedStorage.Retrieve` only looked the cache up by digest, so seq-only lookups (which pass a zero digest) missed blocks that were verified but not yet finalized. Scan the cache by seq for those lookups.

Related to the upcoming auxiliary info dissemination work: the aux history traversal retrieves blocks by seq with a zero digest, so this should merge before or alongside that stack.


samliok:
Duplicate of #497, which carries the same fix plus a unit test.

PR #505, opened 2026-08-13
Refactor instance: extract util helpers, split BlockDeserializer, rework epoch transitions

samliok:
First PR of the cleanup-instance stack (1/4).

- Extract `LastBlock` and the epoch/validator-set derivation out of instance.go into util.go, with unit tests.
- Split `ParseBlock` out of the `VM` interface into a `BlockDeserializer` interface, and move ICM epoch computation from the VM to a `Config.ICMETransition` field.
- Replace `EpochAwareStorage` with `InstanceStorage`, which skips Telocks by block type instead of epoch counting and exposes a generic `onIndex` hook (a no-op here, consumed by the transition listener later in the stack).
- Signal epoch changes through `simplex.EpochConfig.OnSealingBlockIndex` for validators and `nonvalidator.Config.TransitionToValidator` for non-validators instead of the storage wrapper. The non-validator trigger now checks membership in the highest replicated epoch's validator set rather than the P-chain tip.
- Rework `processEpochChange` to hold the lock for the whole transition and honor `Stop` for validators too, and make `notifyEpochChange` keep only the newest pending change instead of blocking the indexing thread.
- Remove the `nodeRole` dispatch, export `testutil.TestControlledBlockBuilder`, add an `errAlreadyStarted` sentinel.

The existing instance tests are kept with mechanical adjustments; the full test refactor lands at the top of the stack.


simplex/epoch.go
```diff
@@ -1490,6 +1492,9 @@ func (e *Epoch) indexFinalization(block common.VerifiedBlock, finalization commo
 		e.broadcast(finalizationMsg)
 
 		e.epochSealed.Store(true)
+		if e.OnSealingBlockIndex != nil {
```
samliok, line 1500: can we add a test that ensures the `OnSealingBlockIndex` function is called?

testutil/controlled.go
```diff
@@ -91,7 +91,7 @@ func (n *ControlledInMemoryNetwork) AdvanceWithoutLeader(round uint64, laggingNo
 
 type ControlledNode struct {
 	*BasicNode
-	bb      *testControlledBlockBuilder
+	bb      *TestControlledBlockBuilder
```
samliok, line 94: can we separate this out into another pr?

config.go
```diff
@@ -53,12 +53,11 @@ type VM interface {
...
 	// or when the VM signals that a block should be built.
 	WaitForPendingBlock(ctx context.Context)
+}
 
+type BlockDeserializer interface {
 	// ParseBlock parses the given block bytes into a VMBlock.
 	ParseBlock(context.Context, []byte) (avalanchego.VMBlock, error)
-
-	// ComputeICMEpoch computes the ICM epoch transition given the input parameters.
-	ComputeICMEpoch(input metadata.ICMEpochInput) metadata.ICMEpochInfo
```
samliok: can we separate this out into a separate PR?

simplex/epoch.go
samliok: updated by claude d02a4a6

testutil/controlled.go
samliok: updated by claude c2b0c10

config.go
samliok: split into #512, this PR is now rebased on it

testutil/controlled.go
samliok: moved to #511, the revert here is 6841c66 after rebasing onto #512

simplex/epoch.go
samliok: now 508e9a8 after rebasing onto #512

PR #507, opened 2026-08-13
Wire the epoch transition listener into the instance

samliok:
Adds `epochTransitionListener` and hooks it into the instance:

- `NewInstance` constructs the listener and wires its `onIndex` into both the validator and non-validator storage paths (replacing the no-op hook).
- When a transition block is indexed, the listener either generates and broadcasts auxiliary info, or once the history is sufficient, signs and broadcasts an epoch transition approval. Validators also record their own approval locally; non-validators pass a nil handler.
- `HandleMessage` routes incoming `AuxiliaryInfo` and `EpochTransitionApproval` messages into the MSM.

### DIFF
--- a/instance.go
+++ b/instance.go
@@ -313,6 +313,14 @@ func (i *Instance) HandleMessage(msg *common.Message, from common.NodeID) error 
 	if i.e != nil {
 		switch {
 		case msg.AuxiliaryInfo != nil:
+			if msg.AuxiliaryInfo.Epoch != i.e.Epoch {
+				i.Config.Logger.Debug(
+					"Received an auxiliary info from an old epoch",
+					zap.Uint64("Aux Info Epoch", msg.AuxiliaryInfo.Epoch),
+					zap.Uint64("Our Epoch", i.e.Epoch),
+					zap.Stringer("From", from))
+				return nil
+			}
 			i.msm.HandleAuxiliaryInfo(*msg.AuxiliaryInfo, avalanchego.NodeID(from))
 		case msg.EpochTransitionApproval != nil:
 			// TODO: pass in time.Now() rather than uint64
--- a/instance.go
+++ b/instance.go
@@ -332,7 +332,7 @@ func (i *Instance) HandleMessage(msg *common.Message, from common.NodeID) error 
 			i.msm.HandleAuxiliaryInfo(*msg.AuxiliaryInfo, avalanchego.NodeID(from))
 		case msg.EpochTransitionApproval != nil:
 			// TODO: pass in time.Now() rather than uint64
-			i.msm.HandleApproval(msg.EpochTransitionApproval, uint64(time.Now().Unix()))
+			i.msm.HandleApproval(msg.EpochTransitionApproval, uint64(time.Now().UnixMilli()))
 		}
 		return i.e.HandleMessage(msg, from)
 	}
--- a/instance.go
+++ b/instance.go
@@ -56,10 +56,6 @@ type epochChange struct {
 	epoch      uint64
 	validators common.Nodes
 }
-
-// TODO: replace noop-index with a function that checks whether we need to send approvals or auxiliary information
-func noopOnIndex(*ParsedBlock) error { return nil }
-
 type timeAdvancer interface {
 	AdvanceTime(t time.Time)
 }
@@ -67,24 +63,40 @@ type timeAdvancer interface {
 type Instance struct {
 	Config Config
 
-	lock         sync.Mutex
-	started      bool
-	cs           *CachedStorage
-	wal          *wal.GarbageCollectedWAL
-	msm          *metadata.StateMachine
-	e            *simplex.Epoch
-	nv           *nonvalidator.NonValidator
-	epochOrNV    timeAdvancer
-	epochChanges chan epochChange
-	stopCh       chan struct{}
+	lock               sync.Mutex
+	started            bool
+	cs                 *CachedStorage
+	transitionListener *epochTransitionListener
+	wal                *wal.GarbageCollectedWAL
+	msm                *metadata.StateMachine
+	e                  *simplex.Epoch
+	nv                 *nonvalidator.NonValidator
+	epochOrNV          timeAdvancer
+	epochChanges       chan epochChange
+	stopCh             chan struct{}
 }
 
 func NewInstance(config Config) *Instance {
+	cs := NewCachedStorage(config.Storage)
+	// Non-validators have no block builder, so they pass a nil approval handler:
+	// they broadcast approvals but do not need to record their own locally.
+	transitionListener := newEpochTransitionListener(
+		config.Logger,
+		config.Broadcaster,
+		avalanchego.NodeID(config.ID),
+		config.PlatformChain.GetValidatorSet,
+		cs.RetrieveBlock,
+		config.CryptoOps,
+		&NoopAuxiliaryInfoApp{}, // TODO: set this in the config
+		nil,
+	)
+
 	return &Instance{
-		Config:       config,
-		stopCh:       make(chan struct{}),
-		epochChanges: make(chan epochChange, 1),
-		cs:           NewCachedStorage(config.Storage),
+		Config:             config,
+		stopCh:             make(chan struct{}),
+		epochChanges:       make(chan epochChange, 1),
+		cs:                 cs,
+		transitionListener: transitionListener,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


instance.go
```diff
@@ -299,6 +311,21 @@ func (i *Instance) HandleMessage(msg *common.Message, from common.NodeID) error
...
+				i.Config.Logger.Debug(
+					"Received an auxiliary info from an old epoch",
+					zap.Uint64("Aux Info Epoch", msg.AuxiliaryInfo.Epoch),
+					zap.Uint64("Our Epoch", i.e.Epoch),
+					zap.Stringer("From", from))
+				return nil
+			}
+			i.msm.HandleAuxiliaryInfo(*msg.AuxiliaryInfo, avalanchego.NodeID(from))
+		case msg.EpochTransitionApproval != nil:
+			// TODO: pass in time.Now() rather than uint64
```
yacovm, line 334: This should be milliseconds, but `Unix()` is seconds.

instance.go
```diff
@@ -463,6 +490,9 @@ func (i *Instance) createEpochConfig(epoch uint64, validators common.Nodes) (*ep
 	comm := newCommunication(i.Config.Sender, i.Config.Broadcaster, validators)
 
 	instanceStorage := NewCallbackStorage(i.cs, msm, func(block *ParsedBlock) error {
+		if err := i.transitionListener.onIndex(block); err != nil {
```
yacovm:
```
		if block.Type() == metadata.BlockTypeTransitioning {
			if err := i.transitionListener.onIndex(block); err != nil {
				return err
			}
		}
```

We only need to send either approvals or aux info if we're transitioning to a new epoch. We currently handle this in `onIndex` but I think it's cleaner to do it here since we anyway do it for `BlockTypeSealing`, this will let us collapse `handleTransitionBlock` into `onIndex`.

transition_listener.go
```diff
@@ -0,0 +1,152 @@
...
+			Version: auxInfoHistory.OldestVersionID,
+			Data:    generatedAuxInfo,
+		},
+	}
+
+	a.broadcaster.Broadcast(auxInfoMessage)
+	return nil
+}
+
+// TODO: use common.Digest
```
yacovm, line 127: Is this TODO too hard to do in this PR?

transition_listener.go
```diff
@@ -0,0 +1,152 @@
...
+		PChainHeight:  nextEpochPChainReference,
+		AuxInfoDigest: auxInfoDigest,
+		Signature:     sig,
+	}
+
+	approvalMessage := common.Message{
+		EpochTransitionApproval: &approval,
+	}
+
+	a.broadcaster.Broadcast(&approvalMessage)
```
yacovm: why not send it to the validators only? iterate through them and send it in unicast.

transition_listener_test.go
```diff
@@ -0,0 +1,236 @@
+package simplex
```
yacovm, line 4: license header

transition_listener.go
```diff
@@ -0,0 +1,152 @@
+package simplex
```
yacovm, line 4: license header

instance.go
```diff
@@ -56,35 +56,47 @@ type epochChange struct {
...
+	// they broadcast approvals but do not need to record their own locally.
+	transitionListener := newEpochTransitionListener(
+		config.Logger,
+		config.Broadcaster,
+		avalanchego.NodeID(config.ID),
+		config.PlatformChain.GetValidatorSet,
+		cs.RetrieveBlock,
+		config.CryptoOps,
+		&NoopAuxiliaryInfoApp{}, // TODO: set this in the config
+		nil,
```
yacovm, line 91: why is this `nil` ?

instance.go
```diff
@@ -299,6 +311,21 @@ func (i *Instance) HandleMessage(msg *common.Message, from common.NodeID) error
...
+					"Received an auxiliary info from an old epoch",
+					zap.Uint64("Aux Info Epoch", msg.AuxiliaryInfo.Epoch),
+					zap.Uint64("Our Epoch", i.e.Epoch),
+					zap.Stringer("From", from))
+				return nil
+			}
+			i.msm.HandleAuxiliaryInfo(*msg.AuxiliaryInfo, avalanchego.NodeID(from))
+		case msg.EpochTransitionApproval != nil:
+			// TODO: pass in time.Now() rather than uint64
+			i.msm.HandleApproval(msg.EpochTransitionApproval, uint64(time.Now().Unix()))
```
yacovm: shouldn't we return here? Why do we continue further to `return i.e.HandleMessage(msg, from)` ?

instance.go
```diff
@@ -56,35 +56,47 @@ type epochChange struct {
...
-	started      bool
-	cs           *CachedStorage
-	wal          *wal.GarbageCollectedWAL
-	msm          *metadata.StateMachine
-	e            *simplex.Epoch
-	nv           *nonvalidator.NonValidator
-	epochOrNV    timeAdvancer
-	epochChanges chan epochChange
-	stopCh       chan struct{}
+	lock               sync.Mutex
```
yacovm, line 66: We added code to `instance.go` but no tests were added. Are the tests in a future PR in the chain after the test refactoring?

transition_listener_test.go
```diff
@@ -0,0 +1,236 @@
...
+	msg := env.broadcaster.messages[0]
+	require.Nil(t, msg.EpochTransitionApproval)
+	require.NotNil(t, msg.AuxiliaryInfo)
+	require.Equal(t, auxApp.DefaultVersionID(), msg.AuxiliaryInfo.Version)
+	require.Equal(t, auxApp.generated, msg.AuxiliaryInfo.Data)
+}
+
+// TestNonValidatorContributesApproval asserts that once the auxiliary info history is
+// sufficient, a non-validator that belongs to the next validator set broadcasts its
+// epoch transition approval. It does not record the approval locally (nil handleApproval),
```
yacovm: > // epoch transition approval. It does not record the approval locally (nil handleApproval),
> // since it has no block builder to include it.

I don't understand, how is the block builder relevant here?

transition_listener_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	require.Len(t, env.broadcaster.messages, 1)
+	msg := env.broadcaster.messages[0]
+	require.Nil(t, msg.AuxiliaryInfo)
+	require.NotNil(t, msg.EpochTransitionApproval)
+
+	approval := msg.EpochTransitionApproval
+	require.Equal(t, testNodeID, approval.NodeID)
+	require.Equal(t, nextPChainRef, approval.PChainHeight)
+	require.Equal(t, []byte("signature"), approval.Signature)
```
yacovm: where is `[]byte("signature")` coming from?

transition_listener_test.go
```diff
@@ -0,0 +1,236 @@
...
+	require.Len(t, env.broadcaster.messages, 1)
+	msg := env.broadcaster.messages[0]
+	require.Nil(t, msg.EpochTransitionApproval)
+	require.NotNil(t, msg.AuxiliaryInfo)
+	require.Equal(t, auxApp.DefaultVersionID(), msg.AuxiliaryInfo.Version)
+	require.Equal(t, auxApp.generated, msg.AuxiliaryInfo.Data)
+	require.Empty(t, env.approvals)
+}
+
+func TestTransitionBroadcastsApproval(t *testing.T) {
```
yacovm: Isn't this test contained in other tests below?

transition_listener_test.go
```diff
@@ -0,0 +1,236 @@
...
+	env := newListenerTestEnv(t, otherValidator, auxApp)
+
+	block := newTransitionBlock(t, 100)
+
+	require.NoError(t, env.listener.onIndex(block))
+	require.Empty(t, env.broadcaster.messages)
+	require.Empty(t, env.approvals)
+}
+
+func TestTransitionNotEnoughAuxiliary(t *testing.T) {
```
yacovm, line 130: I don't understand what this test is doing, can you add some comments?

instance.go
samliok: yep, the tests are in the next PR

instance.go
samliok: yea good cleanup 👍

transition_listener.go
samliok: no, for another because we don't use digest in many places

instance.go
samliok: i set it when we transition to a validator, so that this points the the `msm.HandleApproval` function

transition_listener_test.go
samliok: added

transition_listener_test.go
samliok: yea im not sure either. removed this comment.

transition_listener_test.go
samliok: removed

transition_listener_test.go
samliok: from the stub signer, i moved this to a shared variable so its more clear

yacovm (approved):

PR #508, opened 2026-08-13
Refactor instance tests around a network harness

### LINKED ISSUE #467: Test flake in TestInstanceRestartAcrossEpochs
```
2026-07-27T21:24:20.9419803Z === RUN   TestInstanceRestartAcrossEpochs
2026-07-27T21:24:20.9420974Z [07-27|21:20:38.584] INFO simplex/epoch.go:231 Starting Simplex Epoch {"ID": "2d58e26a0240fc78", "nodes": "[2d58e26a0240fc78]"}
2026-07-27T21:24:20.9422471Z [07-27|21:20:38.584] INFO common/notarization.go:45 Collected Quorum of votes {"round": 1, "votes": 1}
2026-07-27T21:24:20.9424115Z [07-27|21:20:38.585] INFO simplex/epoch.go:2913 Moving to a new round {"prev round": 1, "next round": 2, "prev leader": "2d58e26a0240fc78", "next leader": "2d58e26a0240fc78"}
2026-07-27T21:24:20.9426144Z [07-27|21:20:38.585] INFO simplex/epoch.go:1419 Committed block {"round": 1, "sequence": 1, "digest": "99e2a4c95e55f511646d..."}
2026-07-27T21:24:20.9427775Z [07-27|21:20:39.086] INFO common/notarization.go:45 Collected Quorum of votes {"round": 2, "votes": 1}
2026-07-27T21:24:20.9429928Z [07-27|21:20:39.086] INFO simplex/epoch.go:2913 Moving to a new round {"prev round": 2, "next round": 3, "prev leader": "2d58e26a0240fc78", "next leader": "2d58e26a0240fc78"}
2026-07-27T21:24:20.9431840Z [07-27|21:20:39.087] INFO simplex/epoch.go:1419 Committed block {"round": 2, "sequence": 2, "digest": "2f09830c80e939a294eb..."}
2026-07-27T21:24:20.9433383Z [07-27|21:20:39.588] INFO common/notarization.go:45 Collected Quorum of votes {"round": 3, "votes": 1}
2026-07-27T21:24:20.9435068Z [07-27|21:20:39.588] INFO simplex/epoch.go:2913 Moving to a new round {"prev round": 3, "next round": 4, "prev leader": "2d58e26a0240fc78", "next leader": "2d58e26a0240fc78"}
2026-07-27T21:24:20.9436838Z [07-27|21:20:39.589] INFO simplex/epoch.go:1419 Committed block {"round": 3, "sequence": 3, "digest": "c61d2ad4d28a6a04dad5..."}
2026-07-27T21:24:20.9438539Z [07-27|21:20:39.589] INFO simplex/epoch.go:1431 Committed a sealing block, epoch is sealed {"round": 3, "sequence": 3, "digest": "c61d2ad4d28a6a04dad5..."}
2026-07-27T21:24:20.9440291Z [07-27|21:20:39.589] INFO simplex/epoch.go:731 Shutting down node
2026-07-27T21:24:20.9441639Z [07-27|21:20:39.589] INFO Simplex/instance.go:523 Last block is a sealing block, garbage collecting all WALs preceding it to start a new epoch
2026-07-27T21:24:20.9443377Z [07-27|21:20:39.590] INFO simplex/epoch.go:231 Starting Simplex Epoch {"ID": "2d58e26a0240fc78", "nodes": "[2d58e26a0240fc78]"}
2026-07-27T21:24:20.9444664Z [07-27|21:20:39.684] INFO simplex/epoch.go:731 Shutting down node
2026-07-27T21:24:20.9446095Z [07-27|21:20:39.685] INFO Simplex/instance.go:523 Last block is a sealing block, garbage collecting all WALs preceding it to start a new epoch
2026-07-27T21:24:20.9447914Z [07-27|21:20:39.685] INFO simplex/epoch.go:231 Starting Simplex Epoch {"ID": "2d58e26a0240fc78", "nodes": "[2d58e26a0240fc78]"}
2026-07-27T21:24:20.9449322Z     instance_test.go:377: 
2026-07-27T21:24:20.9450532Z         	Error Trace:	/home/runner/work/Simplex/Simplex/instance_test.go:509
2026-07-27T21:24:20.9452337Z         	            				/home/runner/work/Simplex/Simplex/instance_test.go:377
2026-07-27T21:24:20.9453462Z         	Error:      	Condition never satisfied
2026-07-27T21:24:20.9454379Z         	Test:       	TestInstanceRestartAcrossEpochs
2026-07-27T21:24:20.9455273Z --- FAIL: TestInstanceRestartAcrossEpochs (21.10s)
```

samliok:
Final PR of the cleanup-instance stack (4/4), on top of the transition listener PR.

- All scaffolding moves to `instance_testhelpers_test.go`, rebuilt around a `network` harness: `addNode`, `acceptNewBlock`, `waitUntilSealingBlock`. The queue-based `inMemNetwork` is replaced by `instanceComm`, which delivers messages directly and re-parses blocks so recipients do not share mutable canoto state.
- `instance_test.go` becomes short scenario tests. Epoch sealing tests now rely on production approval/aux info dissemination instead of hand-injecting approvals (removing the old TODO).
- New coverage: validator set unchanged/decreased, a validator skipping an epoch, and signer-set assertions proving a promoted non-validator contributes its approval.
- `TestParseBlockSizeMatchesBytes` moves verbatim to `external_test.go` next to `external.go`.
- `TestNonValidator_StaysNonValidator` (replacing `TestInstanceNonValidatorBootstraps`) is skipped pending offline-node timeouts, and `TestInstanceRestartAcrossEpochs` is dropped (the `node.restart()` helper remains unused).


instance_test.go
```diff
@@ -4,1506 +4,361 @@
...
 
-	// Both storages start with only the genesis block.
-	storage := newStorageWithGenesis(t, genesisBlock)
-	storage2 := newStorageWithGenesis(t, genesisBlock)
+// TestNonValidator_BecomesValidator tests that an upcoming validator becomes a validator
+// when an epoch change they are following is sealed. The non-validator must contribute it's apporval in
+// order to do so.
+// We then check  does so by checking they participated in signing
+// Equivalent test as the previous TestInstanceMixedNodeType.
+func TestNonValidator_BecomesValidator(t *testing.T) {
```
samliok, line 110: equivalent to the previous: TestInstanceMixedNodeType

instance_test.go
```diff
@@ -4,1506 +4,361 @@
...
+	sealing := chain.waitUntilSealingBlock()
+	assertExpectedNodeIds(t, sealing.SealingBlockInfo().ValidatorSet.NodeIDs(), newValidatorSet.NodeIDs())
 }
 
-func TestInstanceDoubleStartFails(t *testing.T) {
-	const basePChainHeight = uint64(1)
+// TestNonValidator_StaysNonValidator ensures that a non-validator does not restart when it is processing
+// previous epoch changes.
+// Equivalent to TestInstanceNonValidatorBootstraps
+func TestNonValidator_StaysNonValidator(t *testing.T) {
```
samliok, line 198: previously was `TestInstanceNonValidatorBootstraps`

instance_test.go
```diff
@@ -4,1506 +4,361 @@
...
 }
 
-func TestNonValidatorSkipsMSMVerification(t *testing.T) {
-	t.Skip("skipping until test instance refactor")
+// TestInstanceValidatorSkipsAnEpoch tests that a validator stops and starts being a validator
+// It boots up as a non-validator then syncs to the highest epoch where it is a validator,
+// then it is no longer a validator, and finally it is
+// Equivalent to: TestInstanceValidatorSkipsAnEpoch. This also starts a validator when the tip is a sealing block so it covers an edge
+// case previously caught from the logging test.
+func TestInstanceValidatorSkipsAnEpoch(t *testing.T) {
```
samliok, line 253: Equivalent to: TestInstanceValidatorSkipsAnEpoch. This also starts a validator when the tip is a sealing block so it covers an edge case previously caught from the logging test.

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
+	pChain.advanceHeight(10)
+
+	sealing := network.waitUntilSealingBlock()
+	assertExpectedNodeIds(t, sealing.SealingBlockInfo().ValidatorSet.NodeIDs(), newValidatorSet.NodeIDs())
+}
 
-	var id [20]byte
-	rand.Read(id[:])
-	nodeID := common.NodeID(id[:])
+// TestInstance_OfflineDuringTransition asserts an epoch transition completes while a
```
yacovm:
```
2026-08-31T14:40:37.4305288Z --- PASS: TestValidator_ValidatorSetDecreased (3.02s)
2026-08-31T14:40:37.4306101Z === RUN   TestInstance_OfflineDuringTransition
2026-08-31T14:40:37.4307634Z [08-31|14:30:53.602] INFO Simplex/instance.go:554 Last block is a sealing block, garbage collecting all WALs preceding it to start a new epoch
2026-08-31T14:40:37.4309724Z [08-31|14:30:53.602] INFO simplex/epoch.go:243 Starting Simplex Epoch {"ID": "0100000000000000", "nodes": "[0100000000000000 0200000000000000 0300000000000000 0400000000000000]"}
2026-08-31T14:40:37.4311664Z [08-31|14:30:53.603] INFO Simplex/instance.go:554 Last block is a sealing block, garbage collecting all WALs preceding it to start a new epoch
2026-08-31T14:40:37.4313525Z [08-31|14:30:53.603] INFO simplex/epoch.go:243 Starting Simplex Epoch {"ID": "0200000000000000", "nodes": "[0100000000000000 0200000000000000 0300000000000000 0400000000000000]"}
2026-08-31T14:40:37.4315668Z [08-31|14:30:53.604] INFO Simplex/instance.go:554 Last block is a sealing block, garbage collecting all WALs preceding it to start a new epoch
2026-08-31T14:40:37.4317770Z [08-31|14:30:53.604] INFO simplex/epoch.go:243 Starting Simplex Epoch {"ID": "0300000000000000", "nodes": "[0100000000000000 0200000000000000 0300000000000000 0400000000000000]"}
2026-08-31T14:40:37.4319304Z [08-31|14:30:56.605] INFO simplex/epoch.go:2851 It is time to build a block {"round": 2}
2026-08-31T14:40:37.4320605Z [08-31|14:30:56.606] INFO simplex/epoch.go:2851 It is time to build a block {"round": 2}
2026-08-31T14:40:37.4321576Z panic: test timed out after 10m0s
2026-08-31T14:40:37.4322204Z 	running tests:
2026-08-31T14:40:37.4322937Z 		TestInstance_OfflineDuringTransition (9m44s)
2026-08-31T14:40:37.4323445Z 
2026-08-31T14:40:37.4323634Z goroutine 2635 [running]:
2026-08-31T14:40:37.4324308Z testing.(*M).startAlarm.func1()
```

yacovm:
```
2026-09-01T00:01:27.0910080Z panic: test timed out after 10m0s
2026-09-01T00:01:27.0910683Z 	running tests:
2026-09-01T00:01:27.0911291Z 		TestInstance_OfflineDuringTransition (9m44s)
```

external_test.go
```diff
@@ -0,0 +1,112 @@
+package simplex
```
yacovm, line 4: license header

yacovm (dismissed):

instance_helpers_test.go
```diff
@@ -0,0 +1,747 @@
...
+	}
+	return testutil.TestQC(qc), nil
+}
+
+type MockStorage struct {
+	t *testing.T
+	*testutil.InMemStorage
+	bd *testInnerBlockDeserializer
+
+	snapLock sync.Mutex
```
yacovm: why is this called snapLock?

instance_helpers_test.go
```diff
@@ -0,0 +1,747 @@
...
+	genesis := &ParsedBlock{StateMachineBlock: metadata.StateMachineBlock{InnerBlock: genesisBlock}}
+	require.NoError(t, s.Index(context.Background(), genesis, common.Finalization{}))
+	return s
+}
+
+func (m *MockStorage) Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error {
+	// We serialized the block so that the original reference isn't shared with other goroutines that may concurrently mutate it.
+	encoded := block.Bytes()
+	seq := m.NumBlocks()
+	m.snapLock.Lock()
```
yacovm: can we just lock and defer unlock for simplicity?

instance_helpers_test.go
```diff
@@ -0,0 +1,747 @@
...
+		InMemStorage: testutil.NewInMemStorage(),
+		blocks:       make(map[uint64]storedBlock),
+		bd:           bd,
+	}
+}
+
+func NewMockStorageWithGenesis(t *testing.T, bd *testInnerBlockDeserializer) *MockStorage {
+	s := &MockStorage{
+		t:            t,
+		InMemStorage: testutil.NewInMemStorage(),
```
yacovm: aren't we saving the blocks twice?

instance_helpers_test.go
samliok: i think u named it that 😅

samliok: just renamed

instance_helpers_test.go
samliok: sure, i copied this over from the previous changes though

instance_helpers_test.go
```diff
@@ -85,32 +87,22 @@ func (pc *testPlatformChain) currentHeight() uint64 {
...
+	pc.lock.Lock()
+	defer pc.lock.Unlock()
 
-	var lastCheckpoint uint64
-	for _, h := range heights {
-		if h > height {
-			break
-		}
-		lastCheckpoint = h
+	set, ok := pc.validatorSetAtHeight[height]
```
yacovm, line 109: how is this correct? 

If I have a validator set at height `100` and the next `setValidatorSetAt` call I get is at height `200` shouldn't `GetValidatorSet(150)` work?

instance_helpers_test.go
```diff
@@ -121,32 +113,48 @@ func (pc *testPlatformChain) GetCurrentHeight() uint64 {
...
+
+	pc.validatorSetAtHeight[height] = validatorSet
+}
+
+// advanceHeight bumps the P-chain height and wakes every WaitForProgress waiter.
+func (pc *testPlatformChain) advanceHeight(height uint64) {
+	pc.lock.Lock()
+	defer pc.lock.Unlock()
+
+	if height <= pc.height {
```
yacovm, line 180: shouldn't == be a no-op?

instance_helpers_test.go
```diff
@@ -54,6 +57,15 @@ func (ibd *testInnerBlockDeserializer) ParseBlock(_ context.Context, buff []byte
 	return b, nil
 }
 
+const genesisPChainHeight uint64 = 0
+
+var genesisBlock = &testInnerBlock{Height_: genesisPChainHeight, TS: time.Now(), Payload: []byte("genesis")}
```
yacovm:
```
var (


)
```

instance_helpers_test.go
```diff
@@ -260,6 +285,34 @@ func (m *MockStorage) parseStored(encoded []byte) metadata.StateMachineBlock {
...
 
+// newChainStorage builds and indexes the minimum chain a node can start from: genesis plus
+// epoch 1's defining block, which carries the descriptor naming the epoch's validator set.
+// It returns the storage and the epoch-defining block at its tip.
+func newChainStorage(t *testing.T, validators metadata.NodeBLSMappings) (*MockStorage, metadata.StateMachineBlock) {
+	storage := NewMockStorageWithGenesis(t, &testInnerBlockDeserializer{})
+	genesis, ok := storage.blockAt(0)
+	require.True(t, ok)
+
+	epochBlock := metadata.StateMachineBlock{
```
yacovm: why are we doing this? Won't this make us unable to test the code paths for the first ever simplex block?

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+}
+
+func (c *instanceComm) Send(msg *common.Message, destination common.NodeID) {
+	// loop through all nodes in chain directly send to handle message directly via but do it in a separate go routine
+	for _, n := range c.c.nodesSnapshot() {
+		if !bytes.Equal(n.id, destination) {
+			continue
+		}
+
+		go func(dst *Instance) {
```
yacovm: No, let's not do this. 

This is an uncontrolled spawning of a goroutine and we should have something more structured like a dispatching goroutine and a queue or even an array of callbacks like here: https://github.com/ava-labs/avalanchego/pull/5799/changes/5323655dc6da358d51420edd1284676ba5fbb679#diff-99aaeb3bd0cf52fb818243017f8eb5044ca8f25202b09c3ec3a7a8595e5234f5R1-R36

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+}
+
+func (c *instanceComm) Broadcast(msg *common.Message) {
+	// send to every node in the chain but ourselves, each on its own go routine
+	for _, n := range c.c.nodesSnapshot() {
+		if bytes.Equal(n.id, c.id) {
+			continue
+		}
+
+		go func(dst *Instance) {
```
yacovm: similar comment - let's not spin up goroutines in an uncontrolled manner.

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+}
+
+func newBlockBuilderVM(bb *testutil.TestControlledBlockBuilder, storage *MockStorage, pending *pendingBlockSignal) *blockBuilderVM {
+	return &blockBuilderVM{bb: bb, storage: storage, pending: pending}
+}
+
+func (vm *blockBuilderVM) BuildBlock(ctx context.Context, pChainHeight uint64) (avalanchego.VMBlock, error) {
+	// The builder gates when a block is built; the block it returns is not an inner block, so
+	// it is thrown away.
+	if _, ok := vm.bb.BuildBlock(ctx, common.ProtocolMetadata{}, common.Blacklist{}); !ok {
```
yacovm: why are we invoking a BB that the epoch is invoking here?

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+	}
+
+	// ensure a unique id; snapshot because nodes may be added concurrently
+	for _, node := range n.nodesSnapshot() {
+		require.NotEqual(n.t, node.id, id)
+	}
+
+	comm := newInstanceComm(n, id)
+
+	vm := newBlockBuilderVM(testutil.NewTestControlledBlockBuilder(n.t), storage, n.pending)
```
yacovm: is it me or is the pending signal completely disconnected from the block builder...?

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+	epoch  uint64
+
+	// pending wakes every VM blocked in WaitForPendingBlock when index creates a block.
+	pending *pendingBlockSignal
+
+	validatorSets map[uint64]common.Nodes // epoch -> sorted validators
+
+	// lock guards nodes, which comm goroutines read while addNode appends.
+	lock  sync.Mutex
+	nodes []node
```
yacovm, line 596: Does this include validators and non validators too? 

If so then why do we have `acceptNewBlock` care about non validators?

instance_helpers_test.go
samliok: yea but why would our implementation of MSM/Instance ever need to call  `GetValidatorSet(150)`? the tests explicitly define the validator sets and the heights so i'm not sure interpolating is necessary

yacovm: It may do so according to what is going on in the test. If the test increments the P-chain height to 150, it will call that.

instance_helpers_test.go
samliok: i can make it a no-op, but only the tests will call this function, and there is no benefit of a test calling advanceHeight with the same height.

instance_helpers_test.go
samliok: something that made creating these tests a bit tricky and not deadlock-y is the fact that when a node starts from a fresh network, and it is the leader for the first round, it will try and build the first ever simplex block immediately. 

This is annoying to test because now we need to ensure that every other node is added before the leader of the first round (so that these nodes can receive and finalize the first ever simplex block). Otherwise, the added node will hang in `addNode` since it can't receive a finalization because no other nodes have been added to the network yet.

Therefore, I created this helper so that multi-node tests can start running without building the first ever simplex block and therefore never hang when added to the network. 

But you can still write a test without using this helper. By default, the nodes use a storage that only has the genesis block so the default behavior is building the first ever simplex block.

instance_helpers_test.go
samliok: removed

instance_helpers_test.go
samliok: because we want to ensure validators and non-validators are able to commit blocks

yacovm: but isn't that a problem if we want to make a test in which the non-validator is disconnected or something?

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+	return block, finalization
+}
+
+// waitUntilSealingBlock waits until every node commits the block at the current seq,
+// repeating until that block is a sealing block. It then advances the network into
+// the new epoch and returns the sealing block.
+// This is useful for when we are transitioning epochs because blocks will be built impatiently
+// without a notification from the mempool.
+func (n *network) waitUntilSealingBlock() common.VerifiedBlock {
+	for {
```
yacovm, line 775: what is the reason for this loop to iterate several times? Is it the fact that we build blocks automatically once we start the epoch transition?

instance_helpers_test.go
samliok: yea when we add those tests, we can just add a flag to the `node` struct, and skip `WaitForBlockCommit` for nodes that are offline

instance_helpers_test.go
samliok: yep, because we don't know how many blocks until a sealing block is produced

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+}
+
+// acceptNewBlock blocks until every node has accepted a newly indexed block.
+func (n *network) acceptNewBlock() (common.VerifiedBlock, common.Finalization) {
+	nodes, ok := n.validatorSets[n.epoch]
+	require.True(n.t, ok, fmt.Sprintf("epoch is not set epoch: %d. trying to index seq: %d", n.epoch, n.seq))
+
+	// no nodes have indexed this sequence yet
+	for _, node := range n.nodes {
+		node.storage.EnsureNoBlockCommit(n.t, n.seq)
```
yacovm: This takes way too long. 

Why not:

```
	for _, node := range n.nodes {
		//node.storage.EnsureNoBlockCommit(n.t, n.seq)
		require.Less(n.t, node.storage.NumBlocks() - 1, n.seq)
	}
```

?

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+	}
+
+	leaderID := simplex.LeaderForRound(nodes.NodeIDs(), n.seq)
+	for _, node := range n.nodes {
+		if bytes.Equal(node.id, leaderID) {
+			node.vm.bb.TriggerNewBlock()
+		}
+	}
+
+	// wake every VM blocked in WaitForPendingBlock
```
yacovm: I don't understand what's going on here. Why are we doing this?

instance_helpers_test.go
```diff
@@ -318,3 +371,377 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+	require.NoError(n.t, node.inst.Start(ctx))
+	n.t.Cleanup(node.inst.Stop)
+
+	instance.Config.Logger.Debug("Added a node to the test network", zap.Uint64("Seq", n.seq), zap.Uint64("num block", node.storage.NumBlocks()))
+	return &node
+}
+
+// acceptNewBlock blocks until every node has accepted a newly indexed block.
+func (n *network) acceptNewBlock() (common.VerifiedBlock, common.Finalization) {
+	nodes, ok := n.validatorSets[n.epoch]
```
yacovm: why do we have a per epoch validator set and we use it to iterate over nodes to accept a block?

instance_helpers_test.go
yacovm: That's because we:

1. Don't have an ingress channel for messages, so messages are lost
2. Don't start the network explicitly, but start a node once we add it.

If we do both then we won't have this problem.

yacovm: The first ever simplex block is a fragile part of the MSM because it doesn't have to be finalized before we continue further, so I think it should be exercised as much as possible.

instance_test.go
```diff
@@ -136,47 +41,43 @@ func (r *emptyVoteRecorder) Broadcast(msg *common.Message) {
...
-	// is cancelled: the MSM must decide to build on its own for the round to make progress.
-	vm := newTestVM()
-	vm.pause()
-
-	inst := newInstanceWithVM(t, ourNode, storage, net, pChain, cops, genesisBlock, vm)
-
-	// Capture the empty vote the node broadcasts once it gives up waiting for the leader.
-	recorder := &emptyVoteRecorder{Broadcaster: inst.Config.Broadcaster, got: make(chan struct{}, 1)}
-	inst.Config.Broadcaster = recorder
+// TestEpochInvokesMSMWaitForPendingBlock verifies the epoch drives the MSM's WaitForPendingBlock:
```
yacovm: I don't understand the description. Why would we want to make an empty vote in such a case?

instance_test.go
```diff
@@ -136,47 +41,43 @@ func (r *emptyVoteRecorder) Broadcast(msg *common.Message) {
...
+	leader := newBLSMapping(2)
+	genesisSet := []metadata.NodeBLSMapping{ourValidator, leader}
+
+	pChain := newTestPChain(genesisSet)
+	nodes := pChain.GenesisValidatorSet().Nodes()
+	common.SortNodes(nodes)
+	require.NotEqual(t, common.NodeID(ourValidator.NodeID[:]), simplex.LeaderForRound(nodes.NodeIDs(), 1))
+
+	storage := NewMockStorageWithGenesis(t, &testInnerBlockDeserializer{})
+	// the VM blocks in WaitForPendingBlock until a block is indexed, which never happens here
```
yacovm: I don't think we should comment about internals in the test, but regardless - I don't understand this comment.

instance_helpers_test.go
```diff
@@ -0,0 +1,822 @@
...
+		// add the validator set to the networks memory for block building
+		n.epoch = n.seq
+		newValidatorSet := block.SealingBlockInfo().ValidatorSet
+		common.SortNodes(newValidatorSet)
+		n.validatorSets[n.epoch] = newValidatorSet
+		return block
+	}
+}
+
+// newBLSMapping creates a mapping with a nodeID, BLSKey and Weight with a given [id].
```
yacovm: I think we should probably give it a different name, seeing how we use this function. 

Maybe `newNodeMapping`?

instance_test.go
yacovm: can we have names without _?

yacovm: can we have names without _?

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
+	network := newNetwork(t, pChain)
+	network.addNode(validator.NodeID[:])
 
-	genesisBlock := &testInnerBlock{Height_: 0, TS: time.Now(), Payload: []byte("genesis")}
+	network.acceptNewBlock()
 
-	net := newInMemNetwork(t)
-	t.Cleanup(net.stop)
+	nonValidator := newBLSMapping(2)
+	network.addNode(nonValidator.NodeID[:])
```
yacovm, line 103: It's non-trivial that when we add a node, we also wait for it to sync with the rest of the network. 

Perhaps add another: `network.acceptNewBlock()` at the end?

instance_helpers_test.go
```diff
@@ -0,0 +1,822 @@
...
+
+	return block, finalization
+}
+
+// waitUntilSealingBlock waits until every node commits the block at the current seq,
+// repeating until that block is a sealing block. It then advances the network into
+// the new epoch and returns the sealing block.
+// This is useful for when we are transitioning epochs because blocks will be built impatiently
+// without a notification from the mempool.
+func (n *network) waitUntilSealingBlock() common.VerifiedBlock {
```
yacovm: can we just pass the required/expected validator set as a parameter and assert that it's equal to the sealing block at the end?

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
-		t.Fatal("non-validator did not transition to validator")
-	}
+	// initiate an epoch change
+	pChain.setValidatorSetAt(10, []metadata.NodeBLSMapping{validator})
+	pChain.advanceHeight(10)
 
-	// The newly promoted validator now participates in extending the chain.
-	require.Equal(t, nonValidatorInstance.Config.ID, latestValidatorID(t, storage))
+	// potential time to propose blocks (if any)
+	time.Sleep(3 * time.Second)
```
yacovm:
```
		return node.storage.NumBlocks() != prevBlockCount	
	}, time.Second, 100*time.Millisecond, "no new block should have been proposed")
```
?

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
-	require.Equal(t, logEpochFromSealingStorage, *lastEpochBranch.Load())
+	pChain := newTestPChain(genesisSet)
+	network := newNetwork(t, pChain)
+	wg := sync.WaitGroup{}
 
-	// The restarted node keeps extending the chain.
-	waitForNumBlocks(t, storage, storage.NumBlocks()+2)
-}
+	wg.Go(func() {
+		// add node is a synchronous call.
```
yacovm: why async? I don't understand. 

We can just add an addNodes(...) method which will add them and then start them instead of doing this async.

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
+	block, _ := network.acceptNewBlock()
+	require.Equal(t, uint64(2), block.BlockHeader().Round)
+
+	// initiate an epoch change
+	newValidatorSet := metadata.NodeBLSMappings{validator}
+	pChain.setValidatorSetAt(10, newValidatorSet)
+	pChain.advanceHeight(10)
+
+	sealing := network.waitUntilSealingBlock()
+	assertExpectedNodeIds(t, sealing.SealingBlockInfo().ValidatorSet.NodeIDs(), newValidatorSet.NodeIDs())
```
yacovm: Maybe add:

```
	block, _ = network.acceptNewBlock()
	require.Equal(t, sealing.BlockHeader().Seq+1, block.BlockHeader().Seq)
```

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
-	cops := &testCryptoOps{}
-	genesisBlock := &testInnerBlock{Height_: 0, TS: time.Now(), Payload: []byte("genesis")}
+	v1Storage, _ := newChainStorage(t, genesisSet)
+	v2Storage, _ := newChainStorage(t, genesisSet)
+	v3Storage, _ := newChainStorage(t, genesisSet)
 
-	net := newInMemNetwork(t)
-	t.Cleanup(net.stop)
+	// addNode blocks until the node syncs, which needs a quorum online,
+	// so the first nodes are added concurrently
```
yacovm: this is not concurrently though

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
+	network.addNodeWithStorage(v1.NodeID[:], v1Storage)
+	network.addNodeWithStorage(v2.NodeID[:], v2Storage)
+	network.addNodeWithStorage(v3.NodeID[:], v3Storage)
+
+	// using seq should be fine since we have no empty blocks
+	leader := simplex.LeaderForRound(pChain.GenesisValidatorSet().NodeIDs(), network.seq)
+	for !leader.Equals(offline.NodeID[:]) {
+		network.acceptNewBlock()
+		leader = simplex.LeaderForRound(nodeIDs, network.seq)
```
yacovm: `leader = simplex.LeaderForRound(nodeIDs, node1.inst.e.Metadata().Round)` ?

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
 
-	storage := newStorageWithGenesis(t, genesisBlock)
+	network.waitUntilSealingBlock()
+}
 
-	inst := newInstance(t, nodeID, storage, net, pChain, cops, genesisBlock)
-	net.register(nodeID, inst)
-	require.NoError(t, inst.Start(t.Context()))
-	t.Cleanup(inst.Stop)
+// TestNonValidator_StaysNonValidator ensures that a non-validator does not restart when it is processing
```
yacovm: what does "does not restart" mean?

instance_test.go
```diff
@@ -188,1388 +89,371 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
+	network.acceptNewBlock()
 
-	require.NoError(t, inst.Start(t.Context()))
-	t.Cleanup(inst.Stop)
+	pChain.advanceHeight(30)
+	sealingBlock = network.waitUntilSealingBlock()
+	assertExpectedNodeIds(t, sealingBlock.SealingBlockInfo().ValidatorSet.NodeIDs(), targetNodeInHighestEpoch.NodeIDs())
 
-	require.ErrorIs(t, inst.Start(t.Context()), errAlreadyStarted)
+	// the target node should join now
```
yacovm, line 274: where is the check that it had caught up with the rest of the network? Can we make an explicit check in the test?

yacovm: also where is the check that it functions as a validator?

instance_helpers_test.go
samliok: i think this is for an old commit

instance_helpers_test.go
samliok: not sure I understand this comment -> the map is just used to ensure validators are ready before accepting a new block. Otherwise i was running into a weird race

instance_test.go
samliok: removed

instance_test.go
samliok: old comment, with `addNodeWithStorage` we don't need to add nodes concurrently anymore

instance_test.go
samliok: in the previous tests i was trying to test that the node doesn't start a new non-validator instance when it doesn't need to. the way to do this was kind of gross, and i'm not sure how to test the non-validator is not "restarting" aka being recreated. 

I mean i could ensure this by looking at the logs 🤔

instance_test.go
samliok: by calling `addNode` and returning it has caught up with the rest of the network. 

```go
// addNode adds a node to the network and blocks until it catches up with the latest tip
func (n *network) addNode(id common.NodeID) *node {
```

I'll add a check that ensures it contributes its votes

yacovm: It's not possible to check that it contributes its votes because we need only `n-f` out of `n`

yacovm: That's why I think we should issue another block afterwards which proves it functions well if it's a validator.

instance_test.go
yacovm: It's fine that we have a test for this, it's just that the comment throws the reader off, because the restart is an internal thing and should not be related to in the test

instance_helpers_test.go
samliok: fixed

instance_helpers_test.go
samliok: hmm, seems like it. but this was also here from before. i think we can just remove most of the mock storage implementation now that i look at it

samliok: yep a lot could have been removed

https://github.com/ava-labs/Simplex/pull/508/commits/c2970687b0cd48e431d565da587802c33d2c0c5a

instance_helpers_test.go
samliok: done https://github.com/ava-labs/Simplex/pull/508#:~:text=start%20all%20instance%20tests%20from%20zero%20block

all instance tests now start from genesis, and exercise the first ever simplex block path

instance_helpers_test.go
```diff
@@ -395,42 +309,101 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+		id:     id,
+		queue:  make(chan inflightMessage, maxInFlightMessages),
+		closed: make(chan struct{}),
+	}
+}
+
+// start allows messages to be processed
+func (i *instanceComm) start() {
+	i.wg.Add(1)
+	i.wg.Go(func() {
```
yacovm, line 344:
```
func (wg *WaitGroup) Go(f func()) {
	wg.Add(1)
	go func() {
		defer wg.Done()
		f()
	}()
}
```

we don't need the Add, Done

instance_helpers_test.go
```diff
@@ -395,42 +309,101 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+}
+
+// run sends queued messages one at a time until stop is called.
+func (i *instanceComm) run() {
+	for {
+		select {
+		case <-i.closed:
+			return
+		case m := <-i.queue:
+			require.NoError(i.n.t, m.to.HandleMessage(m.msg, m.from))
```
yacovm: `require.NoError` won't fail the test because it's from a different goroutine. 

We can just panic here, it's brutal but works.

instance_helpers_test.go
```diff
@@ -395,42 +309,101 @@ func verifiedQuorumRoundToQuorumRound(t *testing.T, vqr common.VerifiedQuorumRou
...
+	i.closeOnce.Do(func() { close(i.closed) })
+	i.wg.Wait()
+}
+
+// enqueue adds a message to be sent
+func (i *instanceComm) enqueue(m inflightMessage) {
+	select {
+	case i.queue <- m:
+	default:
+		i.n.t.Errorf("node %x dropped a message, queue is full at %d", i.id, maxInFlightMessages)
```
yacovm, line 374: Why not just block until the message is enqueued?

instance_helpers_test.go
```diff
@@ -618,7 +607,7 @@ func newNetwork(t *testing.T, pChain *testPlatformChain) *network {
 	validatorSets := make(map[uint64]common.Nodes)
 	genesisNodes := pChain.GenesisValidatorSet().Nodes()
 	common.SortNodes(genesisNodes)
-	validatorSets[1] = genesisNodes
+	validatorSets[firstEverEpoch] = genesisNodes
```
yacovm: I still don't understand why we need an epoch-->validator set mapping.

Why is having a single validator set not sufficient? The network can only advance to a single epoch at a time.

instance_test.go
```diff
@@ -188,1388 +88,374 @@ func TestEpochInvokesMSMWaitForPendingBlock(t *testing.T) {
...
+	newValidatorSet := []metadata.NodeBLSMapping{v1, v2, v3}
+	pChain.setValidatorSetAt(10, newValidatorSet)
+	pChain.advanceHeight(10)
+
+	network.waitUntilSealingBlock(metadata.NodeBLSMappings(newValidatorSet).Nodes())
 }
 
-// TestInstanceZeroBlockUsesLastNonSimplexPChainHeight asserts that the first ever Simplex block
-// references the P-chain height of the last non-Simplex block.
-func TestInstanceZeroBlockUsesLastNonSimplexPChainHeight(t *testing.T) {
```
yacovm, line 549: do we have a test equivalent to this test?

instance_helpers_test.go
samliok: because blocking would probably mean a deadlock, since its being called from someones `Instance` lock. I think we should just set a reasonable capacity on the channel (in this case `10000` seems plenty), and if we drop one because the channel is full it probably means we are sending too many messages.

instance_helpers_test.go
samliok: yea removed

instance_helpers_test.go
yacovm: I know it'll be a deadlock, but since we have a reasonable capacity, wouldn't that mean that reaching this capacity and having a deadlock is a sign that our tests or production code is wrong? 

Just logging an error and continuing is IMO just avoiding a problem instead of facing it. 

I think we should perhaps even panic if we are out of capacity.

yacovm (approved):

PR #512, opened 2026-08-13
Extract ComputeICMEpoch interface from VM, and some BlockDeserializer nits

samliok:
Replaces the VM ComputeICMEpoch method with a `Config.ICMETransition` function field.

### DIFF
--- a/config.go
+++ b/config.go
@@ -56,9 +56,6 @@ type VM interface {
 
 	// ParseBlock parses the given block bytes into a VMBlock.
 	ParseBlock(context.Context, []byte) (avalanchego.VMBlock, error)
-
-	// ComputeICMEpoch computes the ICM epoch transition given the input parameters.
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -53,12 +53,11 @@ type VM interface {
 	// WaitForPendingBlock returns when either the given context is cancelled,
 	// or when the VM signals that a block should be built.
 	WaitForPendingBlock(ctx context.Context)
+}
 
+type BlockDeserializer interface {
 	// ParseBlock parses the given block bytes into a VMBlock.
 	ParseBlock(context.Context, []byte) (avalanchego.VMBlock, error)
```
yacovm, line 58: But in practice, deserializing blocks is a function of the VM. 

I am fine with extracting out the ComputeICMEpoch, but I think the ParseBlock should stay.

yacovm (approved):

PR #513, opened 2026-08-14
Gate messages to epoch once the epoch is sealed

samliok:


adapters.go
```diff
@@ -153,23 +154,28 @@ func (cs *CachedStorage) Retrieve(seq uint64, digest common.Digest) (common.Veri
...
+
+	// We also delete all blocks that are older than the indexed block, because they are now finalized and persisted.
+	for digest, cachedBlock := range cs.cache {
+		if cachedBlock.BlockHeader().Seq < block.BlockHeader().Seq {
+			delete(cs.cache, digest)
 		}
 	}
 
-	return err
+	if block.SealingBlockInfo() != nil && block.SealingBlockInfo().PrevSealingBlockHash != [32]byte{} {
```
yacovm, line 174: we have the method `onEpochChange` in `instance.go`, can't we just define the variable there?

instance.go
```diff
@@ -325,6 +318,28 @@ func (i *Instance) HandleMessage(msg *common.Message, from common.NodeID) error
 	return nil
 }
 
+func (i *Instance) handleValidatorMessage(msg *common.Message, from common.NodeID) error {
+	// we only want to process replication requests if the epoch is sealed
+	if i.cs.lastSealedEpoch == i.e.Metadata().Epoch {
```
yacovm, line 323: why `==`? What if it's a replication request for a previous epoch? Shouldn't we serve it as well?

instance.go
```diff
@@ -325,6 +318,28 @@ func (i *Instance) HandleMessage(msg *common.Message, from common.NodeID) error
...
+		if msg.ReplicationRequest != nil && i.e.ReplicationEnabled {
+			return i.e.HandleMessage(msg, from)
+		}
+
+		i.Config.Logger.Debug("Receive a message for an epoch that was sealed", zap.Uint64("epoch", i.cs.lastSealedEpoch), zap.Any("Message", msg))
+		return nil
+	}
+
+	switch {
```
yacovm, line 332: if the message handles one of the cases then shouldn't we return?

yacovm, line 332: why do we have auxiliary info disseminated separately from the approval?

instance.go
samliok: this is not checking the epoch noted in the message, its strictly checking the epoch instance's metadata. So the contents of the replication request have no bearing on this check

instance.go
samliok: not sure i understand. We handle auxiliary info differently than approvals because one comes after the other. They also have different message structures

PR #514, opened 2026-08-14
Build block immediately when transitioning epoch

yacovm:


msm/msm.go
```diff
@@ -271,6 +271,103 @@ func (sm *StateMachine) maybeInitializeApprovalStore(validatorSet NodeBLSMapping
...
+	}
+
+	// In case the parent block is a Telock, we want to look at the sealing block's metadata instead.
+	if parentBlock.Type() == BlockTypeTelock {
+		sealingBlockSeq := parentBlock.Metadata.SimplexEpochInfo.SealingBlockSeq
+		var sealingBlock StateMachineBlock
+		sealingBlock, finalization, err = sm.GetBlock(sealingBlockSeq, common.Digest{})
+		// The only two reasons why we would not be able to retrieve the sealing block are:
+		// (1) The sealing block is finalized and there is a storage failure, or
+		// (2) The sealing block is not finalized yet, in which case we are still transitioning to a new epoch.
```
samliok: this comment is not true, if we have advanced to a round caused by a telock then we have verified the sealing block. This means it's in the cached storage and `GetBlock` should return it regardless of finalization status.

msm/msm.go
```diff
@@ -271,6 +271,103 @@ func (sm *StateMachine) maybeInitializeApprovalStore(validatorSet NodeBLSMapping
...
+		// Either way and in case the sealing block isn't finalized, we shouldn't wait indefinitely for the VM to build a block.
```
samliok: why shouldn't we wait indefinetly? if the sealing block is not finalized, the epoch will either keep broadcasting empty votes to advance the round or it will eventually notarize/finalize it. 

If we empty notarize the round, then the next block that will be proposed may be another sealing block. So I think waiting like normal is fine here, the `BlockBuilder` & epoch should properly cancel the `WaitForPendingBlock` context?

msm/msm.go
```diff
@@ -271,6 +271,103 @@ func (sm *StateMachine) maybeInitializeApprovalStore(validatorSet NodeBLSMapping
...
+		if err != nil || sealingBlock.Type() != BlockTypeSealing || finalization == nil {
+			if err != nil {
```
samliok: can we move this error outside of the if?

msm/msm.go
```diff
@@ -271,6 +271,103 @@ func (sm *StateMachine) maybeInitializeApprovalStore(validatorSet NodeBLSMapping
...
+	switch currentState {
+	case stateFirstSimplexBlock, stateBuildCollectingApprovals:
+		// In case of the first simplex block we need to persist the validator set,
+		// and if we're collecting approvals then we're in the process of an epoch change already.
+		shouldObviouslyBuildBlockImmediately = true
+	case stateBuildBlockEpochSealed:
+		// The parent is a sealing block, whose finalization we already fetched above.
+		// If it isn't finalized we must build a Telock immediately to extend the epoch;
+		// if it is finalized, we fall through to the normal epoch-transition check below.
+		if finalization == nil {
```
samliok, line 332: i dont get how we can ever reach this case. we check if the finalization == nil before and return if it is.

msm/msm.go
yacovm: The block is in the cache, correct. However, notice that we fetch it with a wildcard digest (any digest) becauseit might be buried several Telocks under.

When we fetch by a wildcard digest, we iterate over the cache and we might return a block that is on a different chain than we intended to. 

That's why I said that only if it's finalized this does not happen, because then we will delete the block from the cache.

msm/msm.go
yacovm: > why shouldn't we wait indefinitely? if the sealing block is not finalized, the epoch will either keep broadcasting empty votes to advance the round or it will eventually notarize/finalize it.

If we wait indefinitely and the leader is stuck, we will not send an empty vote, right?

msm/msm.go
yacovm: No, we check if the finalization is nil *and* `parentBlock.Type() == BlockTypeTelock`

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/514/commits/8d8d5e3bbabb07f3796d655651552603cf566b02

msm/msm.go
yacovm: Slightly changed the code, please check, I think it's more clear now: https://github.com/ava-labs/Simplex/pull/514/commits/8d8d5e3bbabb07f3796d655651552603cf566b02

msm/msm.go
```diff
@@ -271,6 +271,101 @@ func (sm *StateMachine) maybeInitializeApprovalStore(validatorSet NodeBLSMapping
...
+		var sealingBlock StateMachineBlock
+		sealingBlock, finalization, err = sm.GetBlock(sealingBlockSeq, common.Digest{})
+		if err != nil {
+			sm.Logger.Debug("Failed retrieving sealing block for previous epoch", zap.Uint64("seq", sealingBlockSeq), zap.Error(err))
+			sm.BlockBuilder.WaitForPendingBlock(ctx)
+			return
+		}
+
+		// If the sealing block isn't finalized, we need to build a Telock immediately to extend the epoch,
+		// so we wait up to MaxBlockBuildingWaitTime and then return.
```
samliok: this comment doesn't make sense. It says we need to build a telock immediately, but then we wait up to MaxBlockBuildingWaitTime.

Also, why do we need to wait? Telocks don't have any inner blocks

msm/msm.go
```diff
@@ -271,6 +271,101 @@ func (sm *StateMachine) maybeInitializeApprovalStore(validatorSet NodeBLSMapping
...
+		}
+		// Else, err is nil and the sealing block is finalized,
+		// so we can use the sealing block's metadata to determine whether we should build a block immediately or not.
+		parentBlock = sealingBlock
+	}
+
+	currentState := parentBlock.Metadata.SimplexEpochInfo.NextState()
+
+	// We first check if we have obvious signs that we need to build a block immediately:
+	var shouldObviouslyBuildBlockImmediately bool
```
samliok, line 321: no action but seems weird naming a variable with `shouldObviously....`. It should be obvious via the code, not the variable name 🤷

msm/msm_test.go
```diff
@@ -1957,3 +1957,196 @@ func TestCollectAuxiliaryInfo(t *testing.T) {
 		})
 	}
 }
+
+// blockingBlockBuilder waits in WaitForPendingBlock until it is handed a pending block, the way a
+// real VM waits on its mempool. The notification is consumed by a single waiter.
+type blockingBlockBuilder struct {
```
samliok, line 1963: i think we have this in testutil.TestBlockBuilder

msm/msm.go
```diff
@@ -271,6 +271,101 @@ func (sm *StateMachine) maybeInitializeApprovalStore(validatorSet NodeBLSMapping
...
+	}
+
+	// Otherwise, we might need to build a block if we detect that we should transition to a new epoch,
+	// so we initialize a blockBuildingDecider and listen while waiting for the VM.
+	// We return when either the VM signals that a block is ready to be built,
+	// or that the blockBuildingDecider detects that we should transition to a new epoch.
+	pChainReferenceHeight := parentBlock.Metadata.SimplexEpochInfo.PChainReferenceHeight
+	if parentBlock.Type() == BlockTypeSealing {
+		// We've moved to a new epoch, so we need to use the next P-chain reference height of the sealing block,
+		// because the P-chain reference height is of the next epoch is inherited from the P-chain reference height of the sealing block.
```
```suggestion
		// because the P-chain reference height of the next epoch is inherited from the P-chain reference height of the sealing block.
```

msm/msm_test.go
```diff
@@ -1957,3 +1957,196 @@ func TestCollectAuxiliaryInfo(t *testing.T) {
...
+
+func (b *blockingBlockBuilder) WaitForPendingBlock(ctx context.Context) {
+	select {
+	case <-b.pending:
+	case <-ctx.Done():
+	}
+}
+
+// TestMSMWaitForPendingBlock checks when WaitForPendingBlock stops waiting for the VM. It must
+// return on its if BuildBlock would have produced a block.
```
samliok:
```suggestion
// TestMSMWaitForPendingBlock checks when WaitForPendingBlock stops waiting for the VM. It must
// return on its own if BuildBlock would have produced a block.
```

samliok (approved):
looks good to me, the function seems a bit complex & can stray away if we ever change things about `BuildBlock`, but i'm not sure how to make it cleaner

msm/msm.go
yacovm: yeah you're right we don't need to wait in such a case.

msm/msm.go
yacovm: so, anything that isn't `shouldObviouslyBuildBlockImmediately == true` means we might need to epoch change but we might not.

msm/msm_test.go
yacovm: they have a different interface, we can't re-use it.

yacovm:
```
BuildBlock(context.Context, uint64) (avalanchego.VMBlock, error)
```

vs 

```
BuildBlock(_ context.Context, metadata common.ProtocolMetadata, blacklist common.Blacklist) (common.VerifiedBlock, bool)
```

msm/msm.go
yacovm: https://github.com/ava-labs/Simplex/pull/514/changes/03de0fc5c142244c24a10e6425c2f47386eb93f4

PR #516, opened 2026-08-14
Evict non-validator replicated blocks from the block cache once finalized

samliok:
## What was wrong
The non-validator's EpochAwareStorage wrapped the raw storage instead of the instance's CachedStorage, unlike the validator path. Blocks replicated by a non-validator are cached as unfinalized when verified, and CachedStorage.Index is the only place cache entries are deleted, so the non-validator's indexing never evicted them.

## Consequence
While a node runs as a non-validator, the cache grows by one entry per replicated block and is never pruned, holding the entire replicated chain in memory. Retrieving one of these finalized blocks by digest returns it with a nil finalization, violating the documented cache invariant. 

```go
// If the block is cached, it means it's not finalized yet, because upon finalizing the block (indexing)
// we also remove it from the cache. Therefore, we return nil for the finalization.
```


instance.go
```diff
@@ -156,8 +156,9 @@ func (i *Instance) createNonValidatorConfig(epochNum uint64, validators common.N
 	comm.SetValidators(validators)
 
 	epochAwareStorage := &EpochAwareStorage{
-		epoch:   epochNum,
-		Storage: i.Config.Storage,
+		epoch: epochNum,
+		// Index through the cache so blocks inserted upon verification are evicted once finalized.
```
yacovm: > // Index through the cache so blocks inserted upon verification are evicted once finalized.

This is a weird place to place this comment. 

I think if we want to comment this, it should be above the struct description and not here.

instance_test.go
```diff
@@ -518,6 +518,71 @@ func TestInstanceZeroBlockUsesLastNonSimplexPChainHeight(t *testing.T) {
 	require.Equal(t, basePChainHeight, zeroBlock.Metadata.SimplexEpochInfo.PChainReferenceHeight)
 }
 
+// TestInstanceNonValidatorEvictsFinalizedBlocks asserts that a non-validator serves replicated
+// blocks with their finalization once they are finalized. Blocks are cached as unfinalized upon
+// verification, so indexing must evict them (instance.go:161) or they are served without a
```
yacovm: we should not reference line numbers in comments. There is no way line numbers won't change.

yacovm: This comment is for all line numbers going forward

instance_test.go
```diff
@@ -518,6 +518,71 @@ func TestInstanceZeroBlockUsesLastNonSimplexPChainHeight(t *testing.T) {
...
+	nonValidatorInstance.Stop()
+
+	// Every replicated block below the tip is finalized in storage, so retrieving it
+	// by digest must return its finalization.
+	numBlocks := storage2.NumBlocks()
+	for seq := uint64(1); seq < numBlocks; seq++ {
+		block, ok := storage2.blockAt(seq)
+		require.True(t, ok)
+		digest := (&ParsedBlock{StateMachineBlock: block}).BlockHeader().Digest
+		_, finalization, err := nonValidatorInstance.cs.Retrieve(seq, digest)
```
yacovm: What does this prove? We could still be retrieving this from the storage right?

I think we can just empty the dependency of the storage of the non-validator and see that it can still retrieve blocks.

adapters.go
```diff
@@ -35,10 +35,11 @@ func (c *Communication) Validators() common.Nodes {
 // Upon an epoch change, it will ignore blocks from previous epochs
 // and will call the onEpochChange callback when a new epoch is detected.
 type EpochAwareStorage struct {
+	*CachedStorage
```
yacovm, line 39: Just because we wire the `CachedStorage` doesn't mean we can't use the `Storage` interface here, what am I missing? We're not using anything specific in the `CachedStorage`'s methods.

samliok: i wanted to force us to use `CachedStorage` since we rely on calling `CachedStorage.Index()` rather than `Storage.Index`.

yacovm (approved):

PR #517, opened 2026-08-17
Separate Instance Helpers into utils.go

samliok:
also adds tests for these utils

### DIFF
--- a/instance.go
+++ b/instance.go
@@ -101,14 +101,7 @@ func (i *Instance) Start(ctx context.Context) error {
 
 	context.AfterFunc(ctx, i.Stop)
 
-	lastBlock, numBlocks, err := i.lastBlock()
-	if err != nil {
-		return fmt.Errorf("error retrieving last block: %w", err)
-	}
-
-	lastNonSimplexHeight := i.Config.LastNonSimplexInnerBlock.Height()
-	genesisValidatorSet := i.Config.PlatformChain.GenesisValidatorSet()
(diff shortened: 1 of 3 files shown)


util.go
```diff
@@ -0,0 +1,111 @@
...
+
+	lastBlock, _, err := storage.GetBlock(numBlocks - 1)
+	if err != nil {
+		return metadata.StateMachineBlock{}, 0, fmt.Errorf("error retrieving last block from storage: %w", err)
+	}
+
+	return lastBlock, numBlocks, nil
+}
+
+// getLastAcceptedEpoch determines the epoch the instance should start at based on
```
yacovm: getLastAcceptedEpochAndValidatorSet

util.go
```diff
@@ -0,0 +1,111 @@
...
+// the last block in storage. If the ledger only contains non-Simplex blocks, the
+// epoch is the first Simplex height. If the last block is a sealing block, the
+// epoch it seals has ended, so the next epoch is returned. Otherwise, the epoch
+// of the last block is returned.
+func getLastAcceptedEpochAndValidatorSet(config *Config) (common.Nodes, uint64, error) {
```
yacovm, line 40: shouldn't we remove `constructEpochAndValidatorSet` then?

samliok: ah yea, this forgot to remove when i splitting this PR up

instance.go
```diff
@@ -418,14 +411,13 @@ func (i *Instance) iAmValidator(nodes common.Nodes) bool {
 }
 
 func (i *Instance) createEpochConfig() (simplex.EpochConfig, error) {
-	lastBlock, numBlocks, err := i.lastBlock()
+	lastBlock, _, err := i.lastBlock()
```
samliok, line 414: in a future pr i remove this call completely as its only needed for garbage collecting the wal

yacovm (approved):

PR #521, opened 2026-08-18
Refactor Non-Validator and Epoch to have a transition callback

### LINKED ISSUE #445: Only start an epoch instance if has synced to the current epoch and is part of the latest validator set
We currently start an epoch instance if it is a validator set of the transitioning epoch(i.e. the epoch `onEpochChange` is called on)
https://github.com/ava-labs/Simplex/blob/main/instance.go#L552

But what if the epoch is an old epoch? Maybe the validator set has changed since then and we aren't even a validator anymore.

samliok:
The main feature of this PR is the callback in the `Epoch` and `NonValidator` structs that signal to parent components a transition is required. This fixes the non-validator bug of restarting in-between epochs when it shouldn't be. 

In the `Epoch` struct, when a sealing block is indexed we will now call the `OnSealingBlockIndex` which will notify the instance of an epoch change. Similarly, the `NonValidator` has a `TransitionToValidator` config that notifies the `Instance` when it should transition an epoch change. 

The main benefit is now the non-validator can signal to the `Instance` when it is ready to transition to a validator. Before, since the Storage component was handling this change, the Storage component needed to re-call into the non-validator to confirm whether the sealing block actually meant a transition was needed. Since the non-validator has all the info, that logic should stay within the non-validator. This also helps avoiding deadlock since we don't need to be concerned about re-grabbing the non-validator lock. 

This PR also simplifies and refactors some of the logic for transitioning an epoch. 
- No more `nodeRole` iota
- removes some methods that i felt weren't adding much such as `transitionEpochNonValidator` and `transitionEpochValidator`
- other small cleanups that were kind of hard to separate from the main changes

### DIFF
--- a/adapters.go
+++ b/adapters.go
@@ -41,9 +41,9 @@ func (c *Communication) Validators() common.Nodes {
 	return nodes
 }
 
-// InstanceStorage is a wrapper around Storage that skips indexing Telocks
+// CallbackStorage is a wrapper around Storage that skips indexing Telocks
 // and delegates post-index handling to a caller-provided onIndex hook.
-type InstanceStorage struct {
+type CallbackStorage struct {
 	// CachedStorage is used to ensure that we prune the cache on Index.
 	*CachedStorage
 
@@ -52,15 +52,15 @@ type InstanceStorage struct {
 	onIndex func(block *ParsedBlock) error
 }
 
-func NewInstanceStorage(storage *CachedStorage, msm *metadata.StateMachine, onIndex func(block *ParsedBlock) error) *InstanceStorage {
-	return &InstanceStorage{
+func NewCallbackStorage(storage *CachedStorage, msm *metadata.StateMachine, onIndex func(block *ParsedBlock) error) *CallbackStorage {
+	return &CallbackStorage{
 		CachedStorage: storage,
 		msm:           msm,
 		onIndex:       onIndex,
 	}
 }
 
-func (s *InstanceStorage) Retrieve(seq uint64) (common.VerifiedBlock, common.Finalization, error) {
+func (s *CallbackStorage) Retrieve(seq uint64) (common.VerifiedBlock, common.Finalization, error) {
 	block, finalization, err := s.GetBlock(seq)
 	if err != nil {
 		return nil, common.Finalization{}, err
@@ -72,7 +72,7 @@ func (s *InstanceStorage) Retrieve(seq uint64) (common.VerifiedBlock, common.Fin
 	return parsedBlock, *finalization, nil
 }
 
-func (s *InstanceStorage) Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error {
+func (s *CallbackStorage) Index(ctx context.Context, block common.VerifiedBlock, certificate common.Finalization) error {
 	pb, ok := block.(*ParsedBlock)
 	if !ok {
 		return fmt.Errorf("expected ParsedBlock, got %T", block)
--- a/adapters.go
+++ b/adapters.go
@@ -20,6 +20,15 @@ type Communication struct {
 	Broadcaster
 }
 
+func newCommunication(sender Sender, broadcaster Broadcaster, validators common.Nodes) *Communication {
+	c := &Communication{
+		Sender:      sender,
+		Broadcaster: broadcaster,
+	}
+	c.SetValidators(validators)
+	return c
+}
+
 func (c *Communication) SetValidators(nodes common.Nodes) {
 	c.nodes.Store(nodes)
 }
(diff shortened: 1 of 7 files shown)


nonvalidator/non_validator.go
```diff
@@ -266,6 +270,16 @@ func (n *NonValidator) newFinalizedBlockTask(block common.Block, finalization *c
 			return md.Digest
 		}
 
+		if block.SealingBlockInfo() != nil {
+			// are we the highest validator
+			highestEpoch, highestValidatorSet := n.epochs.highestEpoch()
```
yacovm, line 275:
```
highestEpoch == md.Seq {
```

So this basically means we have replicated the sealing block of the last epoch?

And the reason this works is that we replicate backwards?

yacovm: If so, can we add some comments?

samliok: yep, if we have an epoch in the `epochs` map, that means we have verified it's accepted by enough validators. Therefore, if the block we are indexing the sealing block that is part of the highest epoch we should check if we need to transition.

will add a comment

simplex/epoch.go
```diff
@@ -1495,6 +1497,9 @@ func (e *Epoch) indexFinalization(block common.VerifiedBlock, finalization commo
 		e.broadcast(finalizationMsg)
 
 		e.epochSealed.Store(true)
+		if e.OnSealingBlockIndex != nil {
```
yacovm, line 1505: Why do we need to add this? This can be implemented in `instance.go`, that's what it's for.

nonvalidator/non_validator.go
yacovm: I'm not sure this is good enough, on second thought. 

If we start the non-validator we initialize this map with the last thing we have in the ledger and then if we commit a sealing block we will transition, won't we? 

I think we should peek at the P-chain and compare the validator set and only if they're the same, then trigger it. 

Thoughts?

samliok: > I think we should peek at the P-chain and compare the validator set and only if they're the same, then trigger it.

Hmm not sure this works either, because what if we are in the middle of a transition? and the highest pchain validator set is different than the current epoch. Maybe our validator is down, and it needs to catch up in order to sign off on the approval that will transition the epoch.

yacovm: ok so i think as long as we do https://github.com/ava-labs/Simplex/issues/530 we should be able to know for sure that we have the latest epoch for real, thoughts?

adapters.go
```diff
@@ -32,48 +41,53 @@ func (c *Communication) Validators() common.Nodes {
 	return nodes
 }
 
-// EpochAwareStorage is a wrapper around Storage that is aware of epoch changes.
-// Upon an epoch change, it will ignore blocks from previous epochs
-// and will call the onEpochChange callback when a new epoch is detected.
-type EpochAwareStorage struct {
+// InstanceStorage is a wrapper around Storage that skips indexing Telocks
+// and delegates post-index handling to a caller-provided onIndex hook.
+type InstanceStorage struct {
```
yacovm, line 46: why the rename? The name doesn't say anything about what it is doing, but just says now who is consuming it.

instance.go
```diff
@@ -360,68 +369,41 @@ func (i *Instance) listenForEpochChanges() {
...
-	return epoch.Start()
-}
+	var err error
 
-func (i *Instance) lastBlock() (metadata.StateMachineBlock, uint64, error) {
-	numBlocks := i.Config.Storage.NumBlocks()
-	if numBlocks == 0 {
-		return metadata.StateMachineBlock{}, 0, fmt.Errorf("no genesis block found in storage")
+	switch {
+	case i.nv != nil:
```
yacovm, line 384: can we be more explicit about what `i.nv != nil ` and `i.e != nil`mean?

instance.go
```diff
@@ -258,9 +258,18 @@ func (i *Instance) stopNonValidator() {
 	}
 }
 
-func (i *Instance) stopValidator() {
+func (i *Instance) stopValidator(garbageCollectWAL bool) {
 	if i.e != nil {
 		i.e.Stop()
+		// Wipe out the WALs from the config so we won't try to load them again
+		if garbageCollectWAL {
+			i.Config.WALs = nil
```
yacovm, line 266: isn't this a problem? How are we going to re-create the WAL now?

instance.go
```diff
@@ -533,56 +510,23 @@ func (i *Instance) maybeGarbageCollectWAL(lastBlock metadata.StateMachineBlock)
...
-	defer i.lock.Unlock()
-
-	// Stop the epoch before doing anything else, so that we don't process any more messages while we are changing epochs.
-	i.stopValidator()
-	// Wipe out the WALs from the config so we won't try to load them again
-	i.Config.WALs = nil
-	// On epoch change, garbage collect the WAL to remove all entries from previous epochs.
-	if err := i.wal.GarbageCollect(math.MaxUint64); err != nil {
-		i.Config.Logger.Error("Error garbage collecting epoch config on epoch change", zap.Error(err))
+func GetHighestValidatorSet(platform PlatformChain) (common.Nodes, error) {
```
yacovm, line 522: This function is only used once, why is it needed?

adapters.go
samliok: i think `EpochAware` was no longer relevant, i renamed to `CallbackStorage` 🤷

instance.go
samliok: i think this is a problem, but i just copied the code from before. Probably should address the WAL issues in a separate PR. 

I had this [PR](https://github.com/ava-labs/Simplex/pull/481), maybe we can add it on to that

instance.go
samliok: removed

nonvalidator/non_validator.go
samliok: yea i think this is sufficient

nonvalidator/non_validator_test.go
```diff
@@ -319,6 +320,154 @@ func TestHandleMessages_DuplicateBlock(t *testing.T) {
...
+// validator set.
+func TestNonValidator_CallsTransition(t *testing.T) {
+	newValidatorID := common.NodeID{5}
+	joinedSet := append(slices.Clone(testNodes), common.Node{Id: newValidatorID, Weight: 1})
+	otherSet := append(slices.Clone(testNodes), common.Node{Id: common.NodeID{6}, Weight: 1})
+
+	tests := []struct {
+		name          string
+		setup         func(t *testing.T) (*testChain, []*messageInfo)
+		expectedCalls []transitionCall
```
yacovm: why is this a slice and not a single element?

instance.go
```diff
@@ -144,67 +151,62 @@ func (i *Instance) startNonValidator(epochNum uint64, validators common.Nodes) e
...
-	case <-i.stopCh:
-		// If the instance is stopped, we don't need to notify about epoch changes.
-		return
+	}
+
+	for {
+		select {
+		case i.epochChanges <- ec:
+			return
+		// The slot holds a stale epoch change: take it, keep the newer of the two and retry.
```
yacovm, line 201: I'm not strictly opposed to this, but - do we really need it now? Do we have a test that fails if this isn't done?

yacovm (approved):

instance.go
samliok: I don't think it can happen because we aren't acquiring the lock so we shouldn't deadlock. I still think we should have it and maybe I can log a warning so our test fails if it ever does?

nonvalidator/non_validator_test.go
samliok: fixed

PR #524, opened 2026-08-18
auxiliary info disemination msm only

samliok:
Blocks now embed an AuxiliaryInfoBatch holding one entry per node. The builder collects legal entries from a new auxInfoStore mempool (populated by HandleAuxiliaryInfo) instead of calling Generate itself.
SignApproval is exported so the transition listener (next PR) can sign and broadcast approvals, and GetAuxiliaryHistory exposes the aux history traversal.

In future prs, the `HandleAuxiliaryInfo` and `HandleApproval` methods will be called by the instance that feeds in messages.


instance_test.go
```diff
@@ -30,6 +30,8 @@ import (
 )
 
 func TestInstanceMixedNodeType(t *testing.T) {
+	t.Skip("skipping until test instance refactor")
```
samliok, line 33: im skipping these because i didn't want to spend time updating the tests since i have them passing here

https://github.com/ava-labs/Simplex/pull/508

common/msg.go
```diff
@@ -432,6 +436,23 @@ type BlockDigestRequest struct {
 // VersionID is an identifier for applications that care about epoch changes.
 type VersionID uint32
 
+//go:generate go run github.com/StephenButtolph/canoto/canoto msg.go
+
+// AuxiliaryInfo defines application-specific information for applications that might care about epoch change,
+// such as threshold distributed public key generation.
```
yacovm: > threshold distributed public key generation.

Just: distributed key generation, without the threshold and the public

(yes I know that this is originally my mistake)

common/msg.go
```diff
@@ -432,6 +436,23 @@ type BlockDigestRequest struct {
...
+type AuxiliaryInfo struct {
+	// VersionID is an identifier that identifies the application.
```
yacovm: Version is the field name

common/msg.go
```diff
@@ -432,6 +436,23 @@ type BlockDigestRequest struct {
...
+	// Can be used for backward-compatibility and upgrade purposes.
+	Version VersionID `canoto:"uint,1"`
+
+	// Info is opaque bytes that can be used by applications to encode any information that describes
```
yacovm: Data, not Info

yacovm: > Info is opaque bytes that can be used by applications to encode any information that describes the current state for the application.

That's the most ambiguous definition I've read for quite some time, but I can only blame myself.

msm/auxiliary.go
```diff
@@ -0,0 +1,177 @@
+package metadata
```
yacovm, line 4: missing license header

msm/auxiliary.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/ava-labs/simplex/avalanchego"
+	"github.com/ava-labs/simplex/common"
+)
+
+//go:generate go run github.com/StephenButtolph/canoto/canoto auxiliary.go
+
+// AuxiliaryInfoBatch is a batch of AuxiliaryInfos to be included in a block
+type AuxiliaryInfoBatch struct {
+	// data is how we expect the order being appended. 0 index is appended first, then data[len()-1] is last
```
yacovm: > // data is how we expect the order being appended. 0 index is appended first, then data[len()-1] is last

I am not sure I understand. Isn't that for any programming language except maybe Matlab? 

I also don't think it's that worthwhile to explain what this means.

msm/auxiliary.go
```diff
@@ -0,0 +1,177 @@
...
+	data []common.AuxiliaryInfo `canoto:"repeated value,1"`
+	// PrevAuxInfoSeq is a sequence number that applications can use to find previous AuxiliaryInfo in the chain.
+	// It is zero if this is the first AuxiliaryInfoBatch for this epoch.
+	PrevAuxInfoSeq uint64 `canoto:"uint,2"`
```
yacovm: This was previously just `AuxiliaryInfo` right? I don't think we should move this out of `encoding.go`. 

I'm fine renaming this but can we please move it back where it was before?

msm/msm.go
```diff
@@ -236,10 +236,18 @@ func NewStateMachine(config *Config) (*StateMachine, error) {
 	if config.TimeSkewLimit == 0 {
 		config.TimeSkewLimit = maxSkew
 	}
-	sm := StateMachine{Config: config}
+	sm := StateMachine{Config: config, auxInfoStore: newAuxInfoStore(config.AuxiliaryInfoApp)}
```
yacovm: Notice that we don't have an epoch in the auxiliary info. 

This will leak aux info across epochs.

We need to re-create the aux info store every time we change epochs.

common/msg.go
```diff
@@ -432,6 +436,23 @@ type BlockDigestRequest struct {
 // VersionID is an identifier for applications that care about epoch changes.
 type VersionID uint32
 
+//go:generate go run github.com/StephenButtolph/canoto/canoto msg.go
+
+// AuxiliaryInfo defines application-specific information for applications that might care about epoch change,
+// such as threshold distributed public key generation.
+type AuxiliaryInfo struct {
```
yacovm, line 443: We should add an epoch field here, otherwise we may get a misunderstanding if a node is lagging and sends information regarding the wrong epoch.

msm/auxiliary.go
```diff
@@ -0,0 +1,177 @@
...
+	// just set the nodes Auxiliary info to the most recent one they sent
+	a.sentInfo[from] = info
+}
+
+// collectAuxInfo returns the stored entries that are legal appends to the given history.
+func (a *auxInfoStore) collectAuxInfo(history AuxInfoHistory, validators NodeBLSMappings) []common.AuxiliaryInfo {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	// Iterate in node ID order so the returned entries are deterministic.
```
yacovm, line 119: why do we need it to be deterministic? So that anyone verifying this would iterate in the same order when applying `IsLegalAppend`?

msm/msm.go
```diff
@@ -1118,6 +1136,19 @@ func assembleApprovalToBeSigned(pChainHeight uint64, auxInfoDigest [32]byte) ([]
 	return asn1.Marshal(signedMsg)
 }
 
+func SignApproval(signer common.Signer, nextPChainReferenceHeight uint64, auxInfoDigest [32]byte) ([]byte, error) {
```
yacovm, line 1140: what's the reason this is exported? This was `createSelfApproval` before.

common/msg.go
```diff
@@ -30,6 +30,10 @@ type Message struct {
 	// Verified Messages
 	VerifiedBlockMessage        *VerifiedBlockMessage
 	VerifiedReplicationResponse *VerifiedReplicationResponse
+
+	// Epoch Transition Messages
```
yacovm, line 34: I'm missing the part where we send these two...?

yacovm (commented):
Made a first pass. 

The most important comment is regarding resetting the aux info store between epochs.

msm/msm.go
samliok: isn't this already done? the msm is created after every epoch(for validators), so `newAuxInfoStore` should also be re-created

yacovm: ah, good point 👍

common/msg.go
samliok: shouldn't this be handled by `Data` and the auxiliary info application? 

I think the only source of truth for whether AuxiliaryInfo is valid should be the `IsLegalAppend` function call.  It shouldn't be up to the msm/instance to filter aux infos. If the application cared about epochs, they can encode that information in `Data` or maybe encode a timestamp.

msm/auxiliary.go
samliok: it's helpful for testing. i don't think it matters for production, because this is only called by the block builder.

msm/msm.go
samliok: it's used in a future PR

samliok: https://github.com/ava-labs/Simplex/pull/507/changes

common/msg.go
samliok: also in a future PR 😉

https://github.com/ava-labs/Simplex/pull/507/changes

common/msg.go
yacovm: But there is no point in putting stuff into memory that if that stuff belongs to the wrong epoch. 

The framework should do its best to provide reliable data delivery. 

If the node is correct but it's just behind, we should exclude its contribution at the protocol layer, not at the app layer.

samliok: ok, will add

i still think a malicious node can bypass this by encoding a different epoch. Therefore, we would still need some sort of check in the application layer

samliok: added the epoch to the struct, but I'll set up blocking old auxiliary info for when I wire it into the `Instance` since the MSM isn't aware of the current epoch.

yacovm: > i still think a malicious node can bypass this by encoding a different epoch. Therefore, we would still need some sort of check in the application layer

Yeah but we should offload protection of malicious input to the application, but not let correct but stale input be regarded as malicious

msm/auxiliary.go
```diff
@@ -0,0 +1,144 @@
...
+	var legalAppends []common.AuxiliaryInfo
+	legalHistory := append([][]byte{}, history.Data...)
+
+	for _, nodeID := range nodeIDs {
+		info := a.sentInfo[nodeID]
+		if history.OldestVersionID != info.Version {
+			continue // keep consistent versions throughout epoch transition
+		}
+
+		if err := a.app.IsLegalAppend(info.Version, validators, legalHistory, info.Data); err != nil {
```
yacovm, line 137: can we add a debug log here?

msm/auxiliary.go
```diff
@@ -0,0 +1,144 @@
...
+			// we don't remove this info from the mempool. maybe it can be added in a different block
```
yacovm, line 138: This is a good decision, because DKG material is linear in the size of the validator set, so we may not be able to add more than a single entry per block.

msm/msm.go
```diff
@@ -1488,106 +1445,86 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
 
-	if err := sm.AuxiliaryInfoApp.IsLegalAppend(versionID, validators, auxInfoHistory.data, proposedAuxInf); err != nil {
-		return nil, [32]byte{}, false, fmt.Errorf("proposed auxiliary info is not a legal append to the history for application %d: %w", versionID, err)
+	// go through all the collected data and return whether proposed datum are legal
+
+	for _, info := range proposedAuxInfos {
+		if auxInfoHistory.OldestVersionID != info.Version {
+			return nil, [32]byte{}, false, fmt.Errorf("proposed auxiliary info does not have the proper version %d: %w", auxInfoHistory.OldestVersionID, err)
+		}
+		if err := sm.AuxiliaryInfoApp.IsLegalAppend(auxInfoHistory.OldestVersionID, validators, auxInfoHistory.Data, info.Data); err != nil {
```
yacovm: This just checks that adding `info` is allowed if we just add this `info`. 

But it might be that adding the first `info` is allowed on its own, adding the second `info` is allowed on its own, but adding the first `info` and then the second `info` is forbidden.

For example, if `proposedAuxInfos` contains several `info` that are the same one, we should only be able to add the first one, not the rest.

We need to accumulate the history as we go through the auxiliary info we check are legal appends.

common/msg.go
```diff
@@ -432,6 +436,26 @@ type BlockDigestRequest struct {
 // VersionID is an identifier for applications that care about epoch changes.
 type VersionID uint32
 
+//go:generate go run github.com/StephenButtolph/canoto/canoto msg.go
+
+// AuxiliaryInfo defines application-specific information for applications that might care about epoch change,
+// such as distributed key generation.
+type AuxiliaryInfo struct {
+	// The epoch this Auxiliary info is associated with
+	Epoch uint64
```
yacovm: where is the `canoto` tag for this?

msm/encoding.go
```diff
@@ -50,7 +50,7 @@ func (smm *StateMachineMetadata) Clone() StateMachineMetadata {
 		PChainHeight:            smm.PChainHeight,
 		Timestamp:               smm.Timestamp,
 		ICMEpochInfo:            smm.ICMEpochInfo.Clone(),
-		AuxiliaryInfo:           smm.AuxiliaryInfo.Clone(),
+		AuxiliaryInfoBatch:      smm.AuxiliaryInfoBatch,
```
yacovm: `.Clone()` ?

msm/msm.go
```diff
@@ -1488,106 +1445,86 @@ func (sm *StateMachine) verifyBlockEpochSealed(ctx context.Context, parentBlock
...
 		}
 	}
 
-	if err := sm.AuxiliaryInfoApp.IsLegalAppend(versionID, validators, auxInfoHistory.data, proposedAuxInf); err != nil {
-		return nil, [32]byte{}, false, fmt.Errorf("proposed auxiliary info is not a legal append to the history for application %d: %w", versionID, err)
+	// go through all the collected data and return whether proposed datum are legal
+
+	for _, info := range proposedAuxInfos {
+		if auxInfoHistory.OldestVersionID != info.Version {
+			return nil, [32]byte{}, false, fmt.Errorf("proposed auxiliary info does not have the proper version %d: %w", auxInfoHistory.OldestVersionID, err)
```
yacovm: `%w` will be always `nil` no? what does `err` stand for here?

yacovm (commented):
Some more comments. Will make another pass after they're addressed.

msm/msm.go
samliok: ah good catch, we need to have some better testing with the auxiliary info stuff. maybe in some future prs or when DKG is added we can create some better aux tests

samliok: i think below i'm also checking `IsSufficient` with the old history, rather than the new appended history

msm/encoding.go
```diff
@@ -458,3 +425,55 @@ func (vsa ValidatorSetApprovals) UniqueByNodeID() ValidatorSetApprovals {
...
+	if ai == nil {
+		return nil
+	}
+	cloned := &AuxiliaryInfoBatch{
+		PrevAuxInfoSeq: ai.PrevAuxInfoSeq,
+	}
+	if ai.data != nil {
+		cloned.data = make([]common.AuxiliaryInfo, len(ai.data))
+		for i, entry := range ai.data {
+			entry.Data = slices.Clone(entry.Data)
```
yacovm: There is no reason to clone the data. We are cloning to avoid cloning the canoto fields.

msm/encoding.go
```diff
@@ -458,3 +425,55 @@ func (vsa ValidatorSetApprovals) UniqueByNodeID() ValidatorSetApprovals {
...
+func (ai *AuxiliaryInfoBatch) Clone() *AuxiliaryInfoBatch {
+	if ai == nil {
+		return nil
+	}
+	cloned := &AuxiliaryInfoBatch{
+		PrevAuxInfoSeq: ai.PrevAuxInfoSeq,
+	}
+	if ai.data != nil {
+		cloned.data = make([]common.AuxiliaryInfo, len(ai.data))
+		for i, entry := range ai.data {
```
yacovm: `for i, entry := range ai.data {` - this is racy. 

We should only use `i` to access otherwise we copy `canotoData` too.

yacovm (commented):
Two important comments and then LGTM

yacovm (approved):

PR #525, opened 2026-08-21
Security Hardening Changes

GeneralZero:
Here are a list of the changes. 

260dc20  Reject a second vote from a node that already voted in the round
09b47c7  Unify msm on the common SignatureVerifier argument order
dbd6d0b  Count Bitmask bits in linear time and add BitLen
64ce9db  Reject an approvals bitmask wider than the validator set
9cf4a8d  Compute WAL record lengths in 64 bits. Prevent Overflow issue
6d8ce20  Rename WALMaxEntryCount to WALMaxSizeBytes. It was counting bytes so just renamed the variable name


yacovm:
Hey, appreciate the code contribution 👍 

But can we please have each commit in its own PR?

GeneralZero:
Yep can do.

PR #533, opened 2026-08-30
Build an empty block if tip is notarized but not finalized

yacovm:
This commit makes the epoch detect that the tip of the chain is only notarized but not finalized for a prolonged period of time, and then cancel the block building or waiting for a need to build a block by passing a ErrShouldBuildEmptyBlock cause to the context.

Then, the MSM returns an empty block.

Instance tests are deferred until https://github.com/ava-labs/Simplex/pull/508 is merged.

### DIFF
--- a/common/api.go
+++ b/common/api.go
@@ -48,6 +48,7 @@ type BlockBuilder interface {
 	// When the given context is cancelled by the caller:
 	// returns an empty block and true, if the context was cancelled with ErrShouldBuildEmptyBlock
 	// returns nil, false otherwise.
+	// The returned boolean indicates whether the block is nil or not.
 	BuildBlock(ctx context.Context, metadata ProtocolMetadata, blacklist Blacklist) (VerifiedBlock, bool)
 
 	// WaitForPendingBlock returns when either the given context is cancelled,
--- a/common/api.go
+++ b/common/api.go
@@ -6,6 +6,7 @@ package common
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 
@@ -36,11 +37,17 @@ type Logger interface {
 	Verbo(msg string, fields ...zap.Field)
 }
 
+var (
+	ErrShouldBuildEmptyBlock = errors.New("should build empty block")
+)
+
 type BlockBuilder interface {
 	// BuildBlock blocks until some transactions are available to be batched into a block,
-	// in which case a block and true are returned.
-	// When the given context is cancelled by the caller, returns false.
-	// The given metadata and blacklist are encoded into the built block.
+	// and the given metadata and blacklist are encoded into the built block.
+	// Returns a block and true unless the given context is cancelled by the caller.
+	// When the given context is cancelled by the caller:
+	// returns an empty block and true, if the context was cancelled with ErrShouldBuildEmptyBlock
+	// returns nil, false otherwise.
 	BuildBlock(ctx context.Context, metadata ProtocolMetadata, blacklist Blacklist) (VerifiedBlock, bool)
 
 	// WaitForPendingBlock returns when either the given context is cancelled,
--- a/common/api.go
+++ b/common/api.go
@@ -41,6 +41,9 @@ var (
 	// ErrShouldBuildEmptyBlock is returned when a block needs to be built even though the VM has no transactions to include in the block.
 	// This is used to advance the chain when the tip is notarized but not finalized.
 	ErrShouldBuildEmptyBlock = errors.New("should build empty block")
+
+	// ErrBlockNotFound is returned when a block cannot be found in storage.
+	ErrBlockNotFound = fmt.Errorf("block not found")
 )
 
 type BlockBuilder interface {
@@ -57,8 +60,6 @@ type BlockBuilder interface {
 	// or when the application signals that a block should be built.
 	WaitForPendingBlock(ctx context.Context)
 }
-
-var ErrBlockNotFound = fmt.Errorf("block not found")
 
 type Storage interface {
 	NumBlocks() uint64
--- a/common/api.go
+++ b/common/api.go
@@ -38,6 +38,8 @@ type Logger interface {
 }
 
 var (
+	// ErrShouldBuildEmptyBlock is returned when a block needs to be built even though the VM has no transactions to include in the block.
+	// This is used to advance the chain when the tip is notarized but not finalized.
 	ErrShouldBuildEmptyBlock = errors.New("should build empty block")
 )
 
--- a/msm/msm.go
+++ b/msm/msm.go
@@ -552,6 +552,12 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
 	blockBuildingDecider := sm.createBlockBuildingDecider(newSimplexEpochInfo.PChainReferenceHeight)
 	decisionToBuildBlock, err := blockBuildingDecider.shouldBuildBlock(ctx)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


common/api.go
```diff
@@ -36,11 +37,17 @@ type Logger interface {
 	Verbo(msg string, fields ...zap.Field)
 }
 
+var (
```
samliok, line 40: no action: but i don't think we need the () for just one variable

common/api.go
```diff
@@ -36,11 +37,17 @@ type Logger interface {
...
+	ErrShouldBuildEmptyBlock = errors.New("should build empty block")
```
samliok, line 43: can we note the edge case that made us come up with this change somewhere in the code? i think as a comment above this err seems like a good place

common/api.go
```diff
@@ -36,11 +37,17 @@ type Logger interface {
...
+)
+
 type BlockBuilder interface {
 	// BuildBlock blocks until some transactions are available to be batched into a block,
-	// in which case a block and true are returned.
-	// When the given context is cancelled by the caller, returns false.
-	// The given metadata and blacklist are encoded into the built block.
+	// and the given metadata and blacklist are encoded into the built block.
+	// Returns a block and true unless the given context is cancelled by the caller.
+	// When the given context is cancelled by the caller:
```
samliok, line 53: As an aside: can we note was the second boolean represents? It's not obvious from looking at the method/comments what false & true return values actually mean

msm/msm.go
```diff
@@ -574,6 +580,9 @@ func (sm *StateMachine) buildBlockOrTransitionEpoch(ctx context.Context, parentB
 	if decisionToBuildBlock.buildInnerBlock {
 		innerBlock, err = sm.BlockBuilder.BuildBlock(ctx, icmEpochInfo.PChainEpochHeight)
```
samliok, line 593: why would this ever be cancelled with ErrShouldBuildEmpty block? If we have gottento this point doesn't it mean should either build an inner block, or a transition block. In the former, we will already build a block impatiently. In the latter, we already have an inner block that should be built so why not include it?

msm/msm_test.go
```diff
@@ -2194,3 +2195,113 @@ func TestMSMWaitForPendingBlock(t *testing.T) {
...
+	cancel      context.CancelCauseFunc
+	cancelCause error // if non-nil, BuildBlock cancels the context with this cause before failing
+	err         error // error returned by BuildBlock
+}
+
+func (b *emptyBlockRequestingBuilder) BuildBlock(context.Context, uint64) (avalanchego.VMBlock, error) {
+	if b.cancelCause != nil {
+		b.cancel(b.cancelCause)
+	}
+	return nil, b.err
```
samliok, line 2224: why does this return nil always?

msm/msm_test.go
```diff
@@ -2194,3 +2195,113 @@ func TestMSMWaitForPendingBlock(t *testing.T) {
...
+}
+
+// TestMSMBuildBlockBuildsEmptyBlockWhenInnerBlockBuildingCancelled covers the branch in
+// buildBlockOrTransitionEpoch where the inner (VM) block builder fails: when the context
+// was cancelled with ErrShouldBuildEmptyBlock, an empty block is returned instead of
+// propagating the error.
+func TestMSMBuildBlockBuildsEmptyBlockWhenInnerBlockBuildingCancelled(t *testing.T) {
+	errVMBuildFailed := errors.New("vm failed to build inner block")
+
+	t.Run("empty-block cause yields an empty block", func(t *testing.T) {
```
samliok, line 2273: this seems very similar to `TestMSMBuildBlockBuildsEmptyBlockWhenBlockBuildingCancelled`

samliok: ah its checking the inner vm block builder, but it still feels weird that the inner vm block builder would need this context to be cancelled. building a block shouldn't take that long especially if its not waiting for an event

yacovm: This is just for test coverage

msm/msm_test.go
yacovm: because this mocking the VM, and in all test cases in which we use this struct, the VM returns a nil block.

msm/msm.go
yacovm: In this place in the code, we do not build the block impatiently. 

In the places where we build the block impatiently, I didn't add this handling.

common/api.go
yacovm: https://github.com/ava-labs/Simplex/pull/533/changes/6be04f12d5366c6cf6941a013ad4d4623bc0595b

common/api.go
yacovm: https://github.com/ava-labs/Simplex/pull/533/changes/1acfe1a927940f2185be9505e433dc4dbb043c26

common/api.go
yacovm: https://github.com/ava-labs/Simplex/pull/533/changes/2979611decb40a5fdefc2f190513e81fe86892a1

simplex/epoch.go
```diff
@@ -2595,15 +2613,15 @@ func (e *Epoch) createBlockBuildingTask(metadata common.ProtocolMetadata, blackl
...
 
 		e.lock.Lock()
 		defer e.lock.Unlock()
 
 		canceled := context.Err() != nil
 		cancel()
-		if !ok {
-			if !canceled {
+		if !ok || canceled {
+			if !ok && !canceled {
```
samliok: something like this to simplify the logic?
```suggestion
		canceled := context.Err() != nil
		cancel()
		if canceled {
			return common.Digest{}
		}
		if !ok {
			e.Logger.Debug("Failed building block")
			return common.Digest{}
		}
```

simplex/epoch.go
```diff
@@ -313,6 +319,18 @@ func (e *Epoch) Start() error {
 	return nil
 }
 
+func (e *Epoch) haveUnFinalizedButNotarizedSuffix(ctx context.Context) bool {
+	<-ctx.Done()
+
+	_, ok := e.haveNotFinalizedNotarizedRound()
```
samliok: this function returns the lowest round number that has a notarization but no finalization. How does this mean the suffix is unFinalized? wouldn't this be more like a prefix 

aside: can we add a function comment on this even though we aren't changing i?, had to go over the logic of the function to understand what it was doing

samliok: would be interesting to have a test where we have multiple notarized blocks

yacovm: > this function returns the lowest round number that has a notarization but no finalization. How does this mean the suffix is unFinalized? wouldn't this be more like a prefix

If we have enough nodes to form a quorum, then if we have any block that is notarized but not finalized, it is either part of a suffix of only non-finalized (but notarized) blocks, or is not part of such a suffix.

 If it is not part of such a suffix of un-finalized blocks, it will be finalized through the mechanism of `CheckForNotFinalizedNotarizedBlocks`. 

Therefore if we have a block that is notarized but not finalized, if it is not finalized after a while, it is part of a suffix of only non-finalized blocks. 

Having said that, I simplified the code to only look at the last round and made the test more comprehensive.

simplex/epoch.go
yacovm: https://github.com/ava-labs/Simplex/pull/533/changes/780ba9156819b8f013b68d3879d41ccb7b4c1527

simplex/epoch.go
yacovm: https://github.com/ava-labs/Simplex/pull/533/changes/7f2ba2caab8660af9f02ea27541b80c9f610ae88

samliok (approved):

PR #534, opened 2026-08-30
Do not query P-chain for current validator set

yacovm:
This commit makes the MSM not need to query the P-chain for the current validator set when transitioning epochs.

Instead, it just looks at the block validation descriptor of the sealing block.

### DIFF
--- a/msm/msm.go
+++ b/msm/msm.go
@@ -296,24 +296,24 @@ func (sm *StateMachine) WaitForPendingBlock(ctx context.Context, currentRoundMet
 	// In order to know whether we're transitioning to a new epoch, or should transition to one,
 	// we need to look at the previous block's metadata.
 
-	prevBlockSeq := currentRoundMetadata.Seq - 1
+	lastBlockSeq := currentRoundMetadata.Seq - 1
 
-	parentBlock, finalization, err := sm.GetBlock(prevBlockSeq, currentRoundMetadata.Prev)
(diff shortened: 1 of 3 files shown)


msm/msm.go
```diff
@@ -809,11 +848,6 @@ func (sm *StateMachine) createBlockBuildingDecider(pChainReferenceHeight uint64)
 			// and the new validator set defined by the given pChainHeight.
 			// If they are different, then we should transition to a new epoch.
 
-			currentValidatorSet, err := sm.GetValidatorSet(pChainReferenceHeight)
```
samliok, line 806: 👍

msm/msm.go
```diff
@@ -721,9 +763,17 @@ func (sm *StateMachine) verifyNextPChainRefHeightNormal(ctx context.Context, pre
 	// It might be that this block is the first block that has set the next P-chain reference height for the epoch,
 	// so check if it has done so correctly by observing whether the validator set has indeed changed.
 
-	currentValidatorSet, err := sm.GetValidatorSet(prevMD.SimplexEpochInfo.PChainReferenceHeight)
+	// Else, !currentValidatorSet.Equal(newValidatorSet) || next.NextPChainReferenceHeight == 0
```
samliok, line 760: is this relevant?

samliok (approved):

msm/msm.go
yacovm: yes, this is not code, this is just an explanation.

yacovm: i moved it from 740 to this line

PR #538, opened 2026-09-01
Cache self-built blocks in CachedStorage

### LINKED ISSUE #537: Proposer Should Add Block to Cache When Building the Block
Otherwise, if we have only notarized a block and then try and build/verify the next block we will not be able to fetch it's parent.

Some logs from an instance test that was flaking
```
[09-01|10:45:02.507] DEBUG msm/msm.go:370 Error while deciding whether to build a block {"test": "TestInstance_OfflineDuringTransition", "myNodeID": 3, "error": "context canceled"}
[09-01|10:45:02.507] DEBUG simplex/epoch.go:2847 Not starting monitoring for block notarization because we were cancelled while waiting for block to be built {"test": "TestInstance_OfflineDuringTransition", "myNodeID": 3, "epoch round": 2, "monitored round": 2}
[09-01|10:45:02.507] DEBUG simplex/monitor.go:89 Task executed {"test": "TestInstance_OfflineDuringTransition", "myNodeID": 3, "taskID": 30}
[09-01|10:45:02.507] DEBUG simplex/monitor.go:87 Executing f {"test": "TestInstance_OfflineDuringTransition", "myNodeID": 3, "taskID": 31}
[09-01|10:45:02.507] DEBUG msm/msm.go:303 WaitForPendingBlock failed to get block {"test": "TestInstance_OfflineDuringTransition", "myNodeID": 3, "seq": 2, "error": "block not found"}
```

samliok:
Addreses #537

### DIFF
--- a/adapters.go
+++ b/adapters.go
@@ -211,15 +211,24 @@ func (n *NoopAuxiliaryInfoApp) DefaultVersionID() common.VersionID {
 	return 0
 }
 
-type BlockBuilderWaiter struct {
+type blockBuilderWaiter struct {
 	lock   sync.Mutex
 	cancel context.CancelFunc
 	msm    *metadata.StateMachine
+	cs     *CachedStorage
 	e      *simplex.Epoch
 	vm     VM
 }
 
-func (bw *BlockBuilderWaiter) stop() {
+func newBlockBuilderWaiter(msm *metadata.StateMachine, cs *CachedStorage, vm VM) *blockBuilderWaiter {
+	return &blockBuilderWaiter{
+		msm: msm,
+		cs:  cs,
+		vm:  vm,
+	}
+}
+
+func (bw *blockBuilderWaiter) stop() {
 	bw.lock.Lock()
 	defer bw.lock.Unlock()
 	if bw.cancel != nil {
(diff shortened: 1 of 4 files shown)


adapters.go
```diff
@@ -248,12 +249,16 @@ func (bw *BlockBuilderWaiter) BuildBlock(ctx context.Context, metadata common.Pr
...
-	pb := ParsedBlock{
+	pb := &ParsedBlock{
 		StateMachineBlock: *block,
 		msm:               bw.msm,
 	}
 
-	return &pb, true
+	// A leader never verifies its own proposal, so nothing else populates the cache with it.
+	// Without this the MSM cannot retrieve the block between notarization and indexing.
+	bw.cs.insertBlock(pb)
```
samliok: i dont like this implementation, going to do it myself.

adapters.go
```diff
@@ -211,15 +211,24 @@ func (n *NoopAuxiliaryInfoApp) DefaultVersionID() common.VersionID {
 	return 0
 }
 
-type BlockBuilderWaiter struct {
+type blockBuilderWaiter struct {
```
samliok, line 214: i dont think this should be exported from this package

PR #570, opened 2026-09-01
Protect against leaders that send several distinct blocks for the same round

### LINKED ISSUE #541: Leader equivocation floods block verification queue, silently dropping genuine proposals
## Details
The Simplex epoch schedules asynchronous block verification through a shared BlockDependencyManager/BasicScheduler pair capped at DefaultProcessingBlocks (500) tasks. handleBlockMessage (simplex/epoch.go:1839-1943) accepts any block for the current round whose header metadata is valid and whose vote is signed by the round's leader, and schedules a verification task for it. There is no per-round or per-leader deduplication before scheduling: the one-proposal-per-(leader,round) buffering at 1885-1902 applies only to future rounds (isWithinMaxRoundWindow requires round > e.round strictly), the proposalBeingProcessed flag is set (1933-1941) but never checked in handleBlockMessage, and the existence of e.rounds[md.Round] is not checked either. Duplicate suppression happens only inside the verification task, after the expensive block.Verify call (storeProposal at 3138-3149; OneTimeVerifier caches only per-digest).

A malicious validator that is the leader of the current round can therefore equivocate: craft up to ~500 distinct blocks (same round/seq/prev — all pinned by verifyProposalMetadataAndBlacklist — but different payloads, hence different digests), sign a proposal vote for each, and send them to every honest node. Each victim schedules one full VM verification per block, filling the 500-slot queue. Each queued task performs block.Verify with no staleness check (2123-2152), so the entire flood is fully verified serially even after the round advances, and the queue drains only at VM-verification speed.

While the queue is full, every subsequent ScheduleTaskWithDependencies call fails admission (common/block_scheduler.go:151-155). In the proposal path the error is silently swallowed (epoch.go:1919-1921, return nil): honest leaders' genuine proposals for the following rounds are discarded with no retry or re-request, so victims cannot verify or vote on them. Rounds then advance only via the MaxProposalWait (5s) empty-vote timeout path, producing empty rounds instead of blocks. Because the attacker sent the flood by broadcast, all honest nodes are starved simultaneously, so no quorum can notarize any block until the flood drains.

The starvation is renewable: leadership rotates round-robin (LeaderForRound = nodes[r % n]), so the attacker refills the queue every n rounds. The blacklist does not mitigate: a leader is marked for blacklisting only if its round ends in an empty notarization (1585, 1595-1609), which the attacker avoids by sending one valid block first (it verifies, stores, and notarizes normally) followed by ~499 junk blocks; and handleBlockMessage never rejects proposals from blacklisted leaders — blacklisting only removes the 5s wait for their proposal (2812-2822). If 500 x VM-verify-latency exceeds n x empty-round duration, a single malicious validator sustains a network-wide liveness outage indefinitely; otherwise each leadership slot still buys a multi-round outage window. The flood also serves as the setup for turning transient backpressure into permanent replication stalls (reported separately).

## Evidence
1. [simplex/epoch.go:1839–1943](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L1839-L1943)
   handleBlockMessage has no deduplication for current-round proposals. For md.Round == e.round it checks only: round not in the past (1857), not too far ahead (1863), sender is the round leader (1870), vote signature valid (1878), and header metadata valid (1905). The one-proposal-per-(leader,round) buffering at 1885-1902 applies only to future rounds because isWithinMaxRoundWindow requires round > e.round strictly. It never checks whether a proposal for this round is already scheduled/being processed (proposalBeingProcessed is set at 1933-1941 but never consulted here, and e.rounds[md.Round] existence is not checked). A malicious leader can therefore submit an unbounded number of distinct, properly signed blocks for its round, each of which is scheduled for full verification at line 1919.
2. [simplex/epoch.go:1919–1921](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L1919-L1921)
   When ScheduleTaskWithDependencies fails with ErrTooManyPendingVerifications (queue full), the error is swallowed and handleBlockMessage returns nil. The genuine leader's proposal is silently discarded: it is not stored in futureMessages (that happens only for future rounds), there is no retry, no re-request, no proposal rebroadcast mechanism, and no signal to the sender. The victim cannot vote for that round until the timeout/replication fallback.
3. [common/block\_scheduler.go:151–155](https://github.com/ava-labs/Simplex/blob/main/common/block_scheduler.go#L151-L155)
   Admission control: any new verification task is rejected once len(dependencies)+scheduler.Size() >= maxDeps (500). The queue is shared by all block verification (proposals and replication), so spam tasks admitted earlier starve all later admissions.
4. [simplex/epoch.go:2123–2152](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L2123-L2152)
   createBlockVerificationTask calls block.Verify (full VM verification via msm.VerifyBlock -> innerBlock.Verify) before acquiring the epoch lock and without any check that the round is still current (only isEpochSealed is checked). Every queued spam block is therefore fully VM-verified even after the round has advanced, so a flood of N distinct blocks costs N full VM verifications and the queue drains only at VM-verify speed. Tasks run serially on the single BasicScheduler goroutine, and channel-queued tasks cannot be pruned (RemoveOldTasks only touches the dependency list). storeProposal-based dedup (epoch.go:3143) happens only after verification completes.
5. [simplex/epoch.go:212–213](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L212-L213)
   The verification scheduler and the dependency manager are both sized with DefaultProcessingBlocks (500, defined at epoch.go:31), fixing the queue capacity an attacker must fill; the constant is hardcoded, not configurable.
6. [simplex/epoch.go:3138–3149](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L3138-L3149)
   storeProposal rejects a second block for a round, but this dedup only runs inside the verification task after block.Verify has already completed, so it does not prevent queue exhaustion or wasted VM verifications.
7. [simplex/epoch.go:2812–2822](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L2812-L2822)
   The blacklist is not a sufficient protection: being blacklisted only causes honest nodes to skip waiting MaxProposalWait for the leader's proposal — handleBlockMessage never checks IsNodeSuspected, so even a blacklisted leader's proposals are still admitted to the verification queue during its rotational slot. Additionally, the leader is only marked for blacklisting when its round ends in an empty notarization (epoch.go:1585, 1595-1609), which the attacker avoids by letting one valid proposal notarize while the junk flood clogs the queue.

## Impact
Consensus liveness is lost network-wide while the flood drains: honest nodes silently drop all genuine proposals at queue admission, cannot vote, and rounds only advance as empty rounds via 5s timeouts. The attack is renewable every leadership rotation and, with realistic VM verification latency, sustainable indefinitely by a single validator. No confidentiality or safety (integrity) impact: blocks are never wrongly finalized, only not produced.

## Reproduction steps
1. Attacker is a member of the validator set (holds a validator key). It waits for its deterministic round-robin leadership slot (guaranteed every n rounds), then broadcasts up to ~500 distinct self-signed proposals for its round. All checks in handleBlockMessage pass since it is the legitimate leader; each message costs the attacker one signature and a small block. No victim interaction and no special deployment configuration are needed; the code path is the default message-handling path.

## Recommended fix
1. handleBlockMessage schedules a verification task for every leader-signed block for the current round with no per-round/per-leader deduplication, allowing one leader to consume the entire shared verification queue (equivocation is not detected until after full verification). Fix criteria: At most one proposal per (leader, round) may occupy the verification pipeline at a time; additional distinct proposals from the same leader for the same round must be rejected before scheduling (equivocation detected at admission). Verify that a leader sending many distinct valid-header blocks for its round results in only one scheduled verification and no queue growth.
2. When ScheduleTaskWithDependencies fails for the current round's genuine proposal, handleBlockMessage returns nil, discarding the proposal with no retry, re-request, or buffering, so a full queue translates directly into missed rounds. Fix criteria: A proposal that cannot be scheduled due to backpressure must be retained (e.g., buffered like future-round proposals) or re-requested once capacity frees, so the round can still be voted on without waiting for the empty-vote timeout. Verify that a proposal arriving while the queue is full is still eventually verified once the queue drains.
3. Verification tasks execute block.Verify without checking whether the round is stale, so flooded tasks keep consuming VM verification time long after the round advanced, extending the starvation window. Fix criteria: Before performing VM verification, a task must check that its block's round is still relevant (round not advanced past it / round not already holding a stored proposal) and skip verification otherwise. Verify that queued tasks for superseded rounds complete near-instantly.

---
**Severity:** HIGH
**Status:** Open
**Category:** Uncontrolled resource consumption
**Branch:** main
**Date created:** 2026-08-21

---

---
_Related existing issue: #37_

yacovm:
This commit limits block verification at any round to a single block at a time when receiving the block as a block message.

This is to prevent from a malicious leader from sending many distinct blocks and overwhelming the node.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -106,6 +106,7 @@ type Epoch struct {
 	validatorNodeIDs               common.NodeIDs
 	validators                     common.Nodes
 	validatorsToPKs                map[string][]byte
+	pendingRounds                  map[uint64]struct{} // block messages that are verified and may create a new round object soon
 	rounds                         map[uint64]*Round
 	emptyVotes                     map[uint64]*EmptyVoteSet
 	oldestNotFinalizedNotarization NotarizationTime
@@ -223,6 +224,7 @@ func (e *Epoch) init() error {
 	e.timedOutRounds = make(map[uint16]uint64, len(e.validatorNodeIDs))
 	e.redeemedRounds = make(map[uint16]uint64, len(e.validatorNodeIDs))
 	e.rounds = make(map[uint64]*Round)
+	e.pendingRounds = make(map[uint64]struct{})
 	e.emptyVotes = make(map[uint64]*EmptyVoteSet)
 	e.futureMessages = make(messagesFromNode, len(e.validatorNodeIDs))
 	e.replicationState = NewReplicationState(e.Logger, e.Comm, e.ID, e.MaxRoundWindow, e.ReplicationEnabled, e.StartTime, &e.lock, e.RandomSource)
@@ -1908,6 +1910,16 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
 		return nil
 	}
 
+	if e.isRoundPending(md.Round) {
+		e.Logger.Debug("Got block for a round that is pending", zap.Uint64("round", md.Round))
+		return nil
+	}
+
+	// Create a task that will verify the block in the future, after its predecessors have also been verified.
+	blockVerificationTask := e.createBlockVerificationTask(e.oneTimeVerifier.Wrap(block), from, vote)
+	// Mark the round as pending and wrap the task with a task that will cleanup the pending round after the block verification task is executed.
+	task := e.markPendingRoundAndCleanupAfter(md.Round, blockVerificationTask)
+
 	if !e.verifyProposalMetadataAndBlacklist(block) {
 		e.Logger.Debug("Got invalid block in a BlockMessage")
 		return nil
@@ -1918,9 +1930,6 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
 	if len(missingRounds) > 0 {
 		e.sendMissingRoundsRequest(from, missingRounds)
 	}
-
-	// Create a task that will verify the block in the future, after its predecessors have also been verified.
-	task := e.createBlockVerificationTask(e.oneTimeVerifier.Wrap(block), from, vote)
 
 	if err := e.blockVerificationScheduler.ScheduleTaskWithDependencies(task, md.Seq, prevBlockDependency, missingRounds); err != nil {
 		return nil
@@ -1947,6 +1956,24 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
 	}
 
 	return nil
+}
+
+func (e *Epoch) isRoundPending(round uint64) bool {
+	_, exists := e.pendingRounds[round]
+	return exists
+}
+
+// markPendingRoundAndCleanupAfter marks a round as pending and returns a function that will clean up the pending round after the task is executed.
+func (e *Epoch) markPendingRoundAndCleanupAfter(round uint64, task common.Task) common.Task {
+	e.pendingRounds[round] = struct{}{}
+	return func() common.Digest {
+		defer func() {
+			e.lock.Lock()
+			defer e.lock.Unlock()
+			delete(e.pendingRounds, round)
+		}()
+		return task()
+	}
 }
 
 func (e *Epoch) sendMissingRoundsRequest(to common.NodeID, missingRounds []uint64) {
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -106,7 +106,6 @@ type Epoch struct {
 	validatorNodeIDs               common.NodeIDs
 	validators                     common.Nodes
 	validatorsToPKs                map[string][]byte
-	pendingRounds                  map[uint64]struct{} // block messages that are verified and may create a new round object soon
 	rounds                         map[uint64]*Round
 	emptyVotes                     map[uint64]*EmptyVoteSet
 	oldestNotFinalizedNotarization NotarizationTime
@@ -224,7 +223,6 @@ func (e *Epoch) init() error {
 	e.timedOutRounds = make(map[uint16]uint64, len(e.validatorNodeIDs))
 	e.redeemedRounds = make(map[uint16]uint64, len(e.validatorNodeIDs))
 	e.rounds = make(map[uint64]*Round)
-	e.pendingRounds = make(map[uint64]struct{})
 	e.emptyVotes = make(map[uint64]*EmptyVoteSet)
 	e.futureMessages = make(messagesFromNode, len(e.validatorNodeIDs))
 	e.replicationState = NewReplicationState(e.Logger, e.Comm, e.ID, e.MaxRoundWindow, e.ReplicationEnabled, e.StartTime, &e.lock, e.RandomSource)
@@ -1915,15 +1913,14 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
 		return nil
 	}
 
-	if e.isRoundPending(md.Round) {
-		e.Logger.Debug("Got block for a round that is pending", zap.Uint64("round", md.Round))
+	// If we are already processing a block for this round, reject the block while it is being processed.
+	if msgForRound, exists := e.futureMessages[string(from)][md.Round]; exists && msgForRound.proposalBeingProcessed {
+		e.Logger.Debug("Got block for a round that is being processed", zap.Uint64("round", md.Round))
 		return nil
 	}
 
 	// Create a task that will verify the block in the future, after its predecessors have also been verified.
-	blockVerificationTask := e.createBlockVerificationTask(e.oneTimeVerifier.Wrap(block), from, vote)
-	// Mark the round as pending and wrap the task with a task that will cleanup the pending round after the block verification task is executed.
-	task := e.markPendingRoundAndCleanupAfter(md.Round, blockVerificationTask)
+	task := e.createBlockVerificationTask(e.oneTimeVerifier.Wrap(block), from, vote)
 
 	prevBlockDependency, missingRounds := e.blockDependencies(md)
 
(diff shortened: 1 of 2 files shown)


samliok:
for future claude fixes, can we link them under development of a PR? Then when we merge the PR, the issue should auto-close 

i did it for this one

samliok:
haven't looked into it but this test is failing `TestEpochStopsAtSealingBlock` via CI

yacovm:
> haven't looked into it but this test is failing `TestEpochStopsAtSealingBlock` via CI

It's due to an unrelated race bug. I will post a fix tomorrow.

yacovm:
> haven't looked into it but this test is failing `TestEpochStopsAtSealingBlock` via CI

Fixed the tests

simplex/epoch.go
```diff
@@ -1928,6 +1930,16 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
...
 
+	if e.isRoundPending(md.Round) {
+		e.Logger.Debug("Got block for a round that is pending", zap.Uint64("round", md.Round))
+		return nil
+	}
+
+	// Create a task that will verify the block in the future, after its predecessors have also been verified.
+	blockVerificationTask := e.createBlockVerificationTask(e.oneTimeVerifier.Wrap(block), from, vote)
+	// Mark the round as pending and wrap the task with a task that will cleanup the pending round after the block verification task is executed.
+	task := e.markPendingRoundAndCleanupAfter(md.Round, blockVerificationTask)
```
samliok: shouldnt we do this after the check below? otherwise if we receive an invalid block we will mark it as pending forever

simplex/epoch.go
```diff
@@ -1969,6 +1978,24 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
 	return nil
 }
 
+func (e *Epoch) isRoundPending(round uint64) bool {
+	_, exists := e.pendingRounds[round]
+	return exists
+}
+
+// markPendingRoundAndCleanupAfter marks a round as pending and returns a function that will clean up the pending round after the task is executed.
+func (e *Epoch) markPendingRoundAndCleanupAfter(round uint64, task common.Task) common.Task {
```
samliok: i feel like its more clear if we just add this in the original `createBlockVerificationTask` and just set it to pending in `handleBlockMessage` without making a new task & method for this

```go
		delete(e.pendingRounds, round)
```

this way we don't create the wrapper tasks, and also we already have the lock in `createBlockVerificationTask`

simplex/epoch_test.go
```diff
@@ -696,30 +708,105 @@ func TestEpochConsecutiveProposalsDoNotGetVerified(t *testing.T) {
...
+
+	require.NoError(t, e.Start())
+
+	md := e.Metadata()
+	require.Equal(t, uint64(0), md.Round)
+
+	var timesVerified atomic.Uint32
+
+	// More distinct, valid proposals than the 500-slot queue can hold.
+	const floodSize = DefaultProcessingBlocks + 50
```
samliok, line 770: no action: but maybe its worth setting this as a config param?
```go
func (e *Epoch) maybeAssignDefaultConfig() error {
```

we can assign it to default if not passed in.

This way we can just set default processing blocks to 1, or something much less than 500 in the test.

yacovm: I'm not sure I follow. This is a test... what config param?

simplex/epoch.go
yacovm: nice catch

simplex/epoch.go
yacovm: I did this intentionally because this way the `delete(e.pendingRounds, round)` will not look out of place in the verification task. If we refactor the verification task it may get silently deleted.

simplex/epoch.go
yacovm: https://github.com/ava-labs/Simplex/pull/570/changes/2847cbfe801df4ab53f536e31b25a6a38d063783

simplex/epoch_test.go
samliok: default processing blocks

samliok (approved):

samliok:
hmm weird failure on `TestEpochVoteSentTwiceKeepsVerifiedVote`

simplex/epoch_test.go
yacovm: I don't see a reason to set something configurable until we know when we would want to change it.

simplex/epoch.go
```diff
@@ -107,6 +106,7 @@ type Epoch struct {
 	validatorNodeIDs               common.NodeIDs
 	validators                     common.Nodes
 	validatorsToPKs                map[string][]byte
+	pendingRounds                  map[uint64]struct{} // block messages that are verified and may create a new round object soon
```
samliok: unnecessary now

simplex/epoch.go
```diff
@@ -1933,15 +1915,21 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
...
 	}
 
+	// If we are already processing a block for this round, reject the block while it is being processed.
+	if msgForRound, exists := e.futureMessages[string(from)][md.Round]; exists && msgForRound.proposalBeingProcessed {
+		e.Logger.Debug("Got block for a round that is being processed", zap.Uint64("round", md.Round))
+		return nil
+	}
+
+	// Create a task that will verify the block in the future, after its predecessors have also been verified.
+	task := e.createBlockVerificationTask(e.oneTimeVerifier.Wrap(block), from, vote)
```
samliok, line 1952: where do we set this block as being processing? if the leader just sends via the handleBlockMessage() then don't we still run into this issue?

samliok (dismissed):

simplex/epoch.go
yacovm: https://github.com/ava-labs/Simplex/pull/570/changes/BASE..62de7c47ff4ae4cc6163c9a0b7474f276b3e8d56#diff-b6a2eb766dc190b568ec16bd7277c8633013e63dcd6e590db5c72dcc0b69d220R1948

simplex/epoch.go
yacovm: fixed

simplex/epoch.go
yacovm: a few lines below

yacovm:
```
	// mark in future messages while we are verifying the block
	msgForRound, exists := e.futureMessages[string(from)][md.Round]
	if !exists {
		msgsForRound := &messagesForRound{}
		msgsForRound.proposalBeingProcessed = true
		e.futureMessages[string(from)][md.Round] = msgsForRound
	} else {
		msgForRound.proposalBeingProcessed = true
	}
```

samliok (approved):

PR #572, opened 2026-09-01
Drop mismatching epoch

yacovm:
This commit makes the epoch object drop messages containing mismatching epoch.

This is a defensive fix.

### DIFF
--- a/simplex/epoch_test.go
+++ b/simplex/epoch_test.go
@@ -2389,3 +2389,267 @@ func TestEpochRejectsReplicatedQuorumRoundWithMismatchedHeader(t *testing.T) {
 		}
 	}
 }
+
+// TestEpochDropsConsensusMessagesFromOtherEpochs ensures that every consensus message
+// type declaring an epoch other than ours is dropped before reaching its handler,
+// while the same message carrying our epoch is processed.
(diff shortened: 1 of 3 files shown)


simplex/epoch_test.go
```diff
@@ -2132,3 +2132,267 @@ func TestEpochVoteSentTwiceKeepsVerifiedVote(t *testing.T) {
...
+	testutil.InjectTestFinalization(t, e, &finalization, nodes[2])
+	committed := storage.WaitForBlockCommit(md.Seq)
+	require.Equal(t, block.BlockHeader().Digest, committed.BlockHeader().Digest)
+}
+
+// TestEpochIgnoresReplicatedQuorumRoundsFromOtherEpochs ensures that quorum rounds
+// carried in replication responses are ignored when they belong to a different
+// epoch, both for finalized blocks and for empty notarizations, and that the
+// same quorum rounds for our epoch are processed.
+func TestEpochIgnoresReplicatedQuorumRoundsFromOtherEpochs(t *testing.T) {
```
samliok, line 2589: nit: wouldn't it be better to just submit one replication response with all edge cases?

samliok (approved):

simplex/epoch_test.go
yacovm: 🤷

PR #573, opened 2026-09-01
Ensure Non-Validators and Validators Bootstrap Before Indexing and Verifying Blocks

### LINKED ISSUE #530: Validate sealing blocks only via backward hash chain validation on bootstrap instead of also via QC
When we bootstrap a node, it continues replicating from its latest epoch stored in its ledger. 

This means that the epoch changes it performs rely on the fact that the blocks that cause the epoch change are properly signed by a quorum of nodes from the latest epoch.

However, this can be problematic in a setting where the node was offline for a long time or is a new node, because if the keys of the validator set have been leaked in the past, a malicious node can falsify an entire chain of blocks. 

What we should do instead is when we bootstrap the node, we replicate only the sealing blocks backwards, until we reach the latest committed sealing block, and then we keep the chain of sealing blocks and use them and never replicate them by verifying quorum certificates.

samliok:
Addresses https://github.com/ava-labs/Simplex/issues/530 but for non-validator and validators.

Nodes will first need to complete bootstrapping before being able to index/verify blocks.


nonvalidator/epochs.go
```diff
@@ -135,12 +135,10 @@ func (e epochs) canValidate(block common.Block) bool {
...
 // latestValidatorSetRetriever is an allows the epoch replicator to get the latest validator set.
 // This is used to calculate the threshold of votes needed to validate an epoch.
-type latestValidatorSetRetriever interface {
-	Validators() common.Nodes
-}
+type latestValidatorSetRetriever func() common.Nodes
 
 // epochDigestCounter counts sealing block responses from validators for each epoch.
-// It uses latestValidatorSetRetriever to determine when the required response threshold
+// It uses LatestValidatorSetRetriever to determine when the required response threshold
```
yacovm: why capital L?

yacovm:
Looks like this test demonstrates we have a liveness problem in case we miss the first broadcast we send via `broadcastLatestEpoch`: 

```
// TestNonValidatorBootstrapsWhenStartupBroadcastIsLost asserts a non-validator still bootstraps when
// nobody receives the replication request it broadcasts on start. The non-validator joins an empty
// network, so that broadcast reaches no one, and the validator only comes online afterwards. Until it
// bootstraps, the non-validator drops every block and finalization the validator sends it, so the only
// way for it to recover is to ask again.
func TestNonValidatorBootstrapsWhenStartupBroadcastIsLost(t *testing.T) {
	validator := newNodeMapping(1)
	pChain := newTestPChain([]metadata.NodeBLSMapping{validator})
	network := newNetwork(t, pChain)

	// The non-validator starts alone, so the replication request it broadcasts on start is lost.
	nonValidator := newNodeMapping(2)
	node := network.addNode(nonValidator.NodeID[:])
	isValidator, bootstrapped := node.role()
	require.False(t, isValidator)
	require.False(t, bootstrapped)

	// The validator then comes online and commits the first block on its own,
	// broadcasting its finalization to the non-validator, which drops it.
	network.addNode(validator.NodeID[:]).sync()

	// The validator is there to answer, so the non-validator must eventually ask again and bootstrap.
	require.Eventually(t, func() bool {
		_, bootstrapped := node.role()
		return bootstrapped
	}, 20*time.Second, 100*time.Millisecond, "the non-validator never bootstrapped after its startup broadcast was lost")

	node.sync()
}
```

yacovm:
In the issue https://github.com/ava-labs/Simplex/issues/530 I wrote: 

> What we should do instead is when we bootstrap the node, we replicate only the sealing blocks backwards, until we reach the latest committed sealing block, and then we keep the chain of sealing blocks and use them and never replicate them by verifying quorum certificates.

The below test demonstrates that that's not what we do:

```
// TestNonValidatorRejectsSealingBlockOffTheHashChain asserts a bootstrapping non-validator only accepts
// sealing blocks that lie on the backward hash chain from the sealing block it bootstrapped on, however
// well they are signed. See https://github.com/ava-labs/Simplex/issues/530.
//
// The node has committed epoch 1's defining block. The current validator set tells it the latest sealing
// block is at seq 3, whose PrevSealingBlockHash names an honest sealing block at seq 2. A peer then serves
// a forged sealing block at seq 2 that carries a valid quorum certificate of epoch 1's validators but is
// not the block seq 3 points to. Epoch 1's keys may have leaked long ago, so the forged block must be
// rejected. Today the node validates it by its quorum certificate alone and indexes it.
func TestNonValidatorRejectsSealingBlockOffTheHashChain(t *testing.T) {
	validator := newNodeMapping(1)
	validators := metadata.NodeBLSMappings{validator}
	signers := []common.NodeID{validator.NodeID[:]}
	sigAggregator := &testutil.TestSignatureAggregator{N: len(validators)}

	pChain := newTestPChain(validators)
	storage, epochBlock := newChainStorage(t, validators)

	// A node outside the validator set whose ledger ends at epoch 1's defining block.
	nonValidator := newNodeMapping(2)
	node := newNetwork(t, pChain).addNodeWithConfig(nonValidator.NodeID[:], nodeConfig{storage: storage})
	isValidator, bootstrapped := node.role()
	require.False(t, isValidator)
	require.False(t, bootstrapped)

	// sealingBlock builds a sealing block at seq in the given epoch, keeping the same validator set.
	sealingBlock := func(seq, epoch uint64, prev, prevSealingBlockHash [32]byte, payload string) *ParsedBlock {
		timestamp := epochBlockTime.Add(time.Duration(seq) * time.Millisecond)
		return &ParsedBlock{StateMachineBlock: metadata.StateMachineBlock{
			InnerBlock: &testInnerBlock{Height_: seq, TS: timestamp, Payload: []byte(payload)},
			Metadata: metadata.StateMachineMetadata{
				Timestamp:               uint64(timestamp.UnixMilli()),
				SimplexProtocolMetadata: common.ProtocolMetadata{Epoch: epoch, Round: seq, Seq: seq, Prev: common.Digest(prev)},
				SimplexEpochInfo: metadata.SimplexEpochInfo{
					EpochNumber:          epoch,
					PrevSealingBlockHash: prevSealingBlockHash,
					BlockValidationDescriptor: &metadata.BlockValidationDescriptor{
						AggregatedMembership: metadata.AggregatedMembership{Members: validators},
					},
				},
			},
		}}
	}

	// The honest chain: seq 1 (in storage) <- honest seq 2 <- seq 3. The forged seq 2 is signed by the
	// same epoch 1 validators but is not what seq 3 points to.
	honestSealing2 := sealingBlock(2, 1, epochBlock.Digest(), epochBlock.Digest(), "honest sealing block 2")
	forgedSealing2 := sealingBlock(2, 1, epochBlock.Digest(), epochBlock.Digest(), "forged sealing block 2")
	require.NotEqual(t, honestSealing2.Digest(), forgedSealing2.Digest())
	sealing3 := sealingBlock(3, 2, honestSealing2.Digest(), honestSealing2.Digest(), "sealing block 3")

	honestFinalization2, _ := testutil.NewFinalizationRecord(t, sigAggregator, honestSealing2, signers)
	forgedFinalization2, _ := testutil.NewFinalizationRecord(t, sigAggregator, forgedSealing2, signers)
	finalization3, _ := testutil.NewFinalizationRecord(t, sigAggregator, sealing3, signers)

	// The current validator set reports seq 3 as the latest sealing block, which bootstraps the node.
	require.NoError(t, node.inst.HandleMessage(&common.Message{
		ReplicationResponse: &common.ReplicationResponse{
			LatestSeq: &common.QuorumRound{Block: sealing3, Finalization: &finalization3},
		},
	}, validator.NodeID[:]))
	_, bootstrapped = node.role()
	require.True(t, bootstrapped)

	// A peer serves the forged seq 2. Its quorum certificate checks out against epoch 1, but its digest
	// is not the PrevSealingBlockHash of seq 3, so it is off the hash chain and must not be indexed.
	require.NoError(t, node.inst.HandleMessage(&common.Message{
		ReplicationResponse: &common.ReplicationResponse{
			Data: []common.QuorumRound{{Block: forgedSealing2, Finalization: &forgedFinalization2}},
		},
	}, validator.NodeID[:]))
		_, _, err := node.storage.Retrieve(2)
		return err == nil
	}, 2*time.Second, 50*time.Millisecond, "indexed a sealing block that is not on the hash chain from the sealing block we bootstrapped on")

	// The honest seq 2 is what seq 3 points to, so it is accepted.
	require.NoError(t, node.inst.HandleMessage(&common.Message{
		ReplicationResponse: &common.ReplicationResponse{
			Data: []common.QuorumRound{{Block: honestSealing2, Finalization: &honestFinalization2}},
		},
	}, validator.NodeID[:]))
	committed := node.storage.WaitForBlockCommit(2)
	require.Equal(t, honestSealing2.Bytes(), committed.Bytes())
}
```

nonvalidator/non_validator.go
```diff
@@ -165,7 +176,68 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+	if qr.Block.SealingBlockInfo() == nil {
+		n.sendRequest(qr.Block.BlockHeader().Epoch, from)
+		return nil
+	}
+
+	if !n.highestEpochCollector.collectedSealingBlockInfo(qr.Block.SealingBlockInfo(), qr.Block.BlockHeader(), from) {
+		return nil
+	}
+
+	n.Logger.Info("Bootstrapped, received a threshold of sealing block info for an epoch", zap.Stringer("Info", qr.Block.SealingBlockInfo()))
```
yacovm: We should only consider ourselves as bootstrapped when we have replicated and committed all blocks from the last block in the ledger to the last known tip.

samliok: i think we can consider ourselves bootstrapped once we have validated the hash chain of sealing blocks, then we can start as normal syncing all the blocks in between

samliok: as to your test, i made a few non-validator tests to ensure we do the backwards hash validation first

`TestNonValidator_BootstrapIgnoresSealingBlockOffChain` && `TestNonValidator_BootstrapWalksHashChain`

common/api.go
```diff
@@ -203,6 +203,31 @@ func (nws Nodes) Contains(nodeID NodeID) bool {
 	return false
 }
 
+// Equal returns whether both hold the same nodes, ignoring order.
+func (nws Nodes) Equal(other Nodes) bool {
```
yacovm, line 207: can we add a UT for this?

common/timeout_handler.go
```diff
@@ -133,6 +133,14 @@ func (t *TimeoutHandler[T]) RemoveTask(ID T) {
 	delete(t.tasks, ID)
 }
 
+// HasTasks reports whether any task is still outstanding.
+func (t *TimeoutHandler[T]) HasTasks() bool {
```
yacovm: Maybe call it something shorter like `Empty()` and just use `! Empty()` ?

instance_test.go
```diff
@@ -368,6 +379,19 @@ func TestNonValidatorSkipsMSMVerification(t *testing.T) {
 	parent, _, err := nonValidatorNode.storage.GetBlock(1)
 	require.NoError(t, err)
 
+	// The non-validator drops every message until it bootstraps. One peer reporting the block
```
yacovm: I don't understand this statement. 

If it drops every message how does it bootstrap? 🤡

nonvalidator/non_validator.go
```diff
@@ -58,6 +59,10 @@ type Config struct {
 	// TransitionToValidator is called when our non-validator indexes the highest known epoch
 	// and it is in the validator set
 	TransitionToValidator func(epoch uint64, validators common.Nodes)
+
+	// Bootstrapped is set once every epoch from our tip up to the one a threshold of the latest
```
yacovm, line 63: Why would we call that bootstrapped? Bootstrapped should just mean that we finished bootstrapping, exactly like we do in snowman. Which is that we have replicated all blocks we know are missing from the latest discovered tip down to the tip before bootstrapping.

nonvalidator/non_validator.go
```diff
@@ -165,7 +197,95 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+		n.Logger.Info("A threshold of validators reported a sealing block", zap.Uint64("Seq", bh.Seq), zap.Stringer("Info", sealingInfo))
+		n.sealingBlockTimeouts.RemoveTask(startBroadcastTask)
+		if !n.isIndexed(bh.Seq) {
+			n.validateSealingBlock(qr, from)
+		}
+	default:
+		return nil
+	}
+
+	// No sealing block is missing, so every epoch from our tip to the highest is validated.
```
yacovm, line 229: but we should still replicate all blocks in the last epoch that we know about before we declare that we have finished bootstrapping.

nonvalidator/non_validator.go
```diff
@@ -154,6 +181,11 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
 		return n.haltedError
 	}
 
+	if !n.Bootstrapped && msg.ReplicationResponse == nil {
```
yacovm, line 184: why don't we answer also `ReplicationRequest`? If we're a validator that has restarted we should answer queries. 

We could be a validator but simply not be aware of it because the condition `latestIndexedEpochValidators.Equal(latestValidatorSet.Nodes())` may not hold due to a single validator change on the P-chain. 

In an event of a system maintenance or power outage where a lot of validators are being shut down and started again  the validators may be then stuck in an endless bootstrapping loop.

instance.go
```diff
@@ -128,6 +123,32 @@ func (i *Instance) Start(ctx context.Context) error {
 	return nil
 }
 
+func (i *Instance) bootstrap() error {
```
yacovm: shouldn't this be called `maybeBootstrap()` ?

instance.go
```diff
@@ -128,6 +123,32 @@ func (i *Instance) Start(ctx context.Context) error {
...
+
+	latestIndexedEpochValidators, err := getLastAcceptedValidatorSet(&i.Config)
+	if err != nil {
+		return err
+	}
+
+	// We have indexed the latest validator set, therefore we can skip bootstrapping and start as a validator.
+	// Note: this may not be the latest epoch, but a future PR will eventually notice we are behind and transition properly.
+	if latestIndexedEpochValidators.Equal(latestValidatorSet.Nodes()) && latestValidatorSet.Nodes().Contains(i.Config.ID) {
+		i.Config.Logger.Debug("Node finished bootstrapping, its latest epoch is up to date with the Platform Chain")
```
yacovm: seems like the node has skipped bootstrapping and never started it.

instance.go
```diff
@@ -128,6 +123,32 @@ func (i *Instance) Start(ctx context.Context) error {
 	return nil
 }
 
+func (i *Instance) bootstrap() error {
+	i.Config.Logger.Debug("Node started bootstrapping")
```
yacovm: we haven't started bootstrapping yet

instance.go
```diff
@@ -162,19 +185,18 @@ func (i *Instance) startNonValidator() error {
...
 
-	height := i.Config.PlatformChain.GetCurrentHeight()
-	mappings, err := i.Config.PlatformChain.GetValidatorSet(height)
+	latestValidatorSet, err := getLatestPlatformChainValidatorSet(i.Config.PlatformChain)
 	if err != nil {
 		return nonvalidator.Config{}, err
 	}
 
-	comm := newCommunication(i.Config.Sender, i.Config.Broadcaster, mappings.Nodes())
+	comm := newCommunication(i.Config.Sender, i.Config.Broadcaster, latestValidatorSet.Nodes())
```
yacovm, line 199: unrelated to this PR, but... what updates the comm's validator set once we move through epochs *after* we bootstrap?

instance.go
```diff
@@ -128,6 +123,32 @@ func (i *Instance) Start(ctx context.Context) error {
...
+	}
+
+	latestIndexedEpochValidators, err := getLastAcceptedValidatorSet(&i.Config)
+	if err != nil {
+		return err
+	}
+
+	// We have indexed the latest validator set, therefore we can skip bootstrapping and start as a validator.
+	// Note: this may not be the latest epoch, but a future PR will eventually notice we are behind and transition properly.
+	if latestIndexedEpochValidators.Equal(latestValidatorSet.Nodes()) && latestValidatorSet.Nodes().Contains(i.Config.ID) {
```
yacovm, line 140: what if we're a validator and behind (old epoch) but somehow the latest on-chain validator set is equal to the P-chain validator set?

We won't know we're behind and will reject all messages because they will have a higher epoch.

Is https://github.com/ava-labs/Simplex/pull/578 supposed to solve this?

nonvalidator/non_validator.go
```diff
@@ -154,6 +181,11 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+		n.Logger.Debug("Dropping message received while bootstrapping, we only accept replication responses", zap.Any("Message", msg), zap.Stringer("From", from))
```
yacovm, line 185: we assume here that the message can only have a single field inside of it. 

But it can have at least a single field, not up to a single field. 

Shouldn't we re-order the switch a few lines below to have `msg.ReplicationResponse != nil:` first?

nonvalidator/epochs.go
yacovm, line 132: what if the block is already indexed?

nonvalidator/non_validator.go
```diff
@@ -165,7 +197,95 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+
+// processBootstrapQuorumRound handles quorum rounds until bootstrapping finishes.
+// Once a threshold validates an epoch, only sealing blocks are validated
+// and stored(in a backwards manner).
+func (n *NonValidator) processBootstrapQuorumRound(qr *common.QuorumRound, from common.NodeID) error {
+	block := qr.Block
+	bh := block.BlockHeader()
+	sealingInfo := block.SealingBlockInfo()
+
+	if sealingInfo == nil {
```
yacovm, line 210: what if we have the epoch in the ledger?

yacovm (commented):
These are the comments I have so far, I'm not nearly done with the review.

instance.go
samliok: yea we need to change the non-validator comm to not hardcode its `Validators()` method. everytime it calls `Validators` the pchain should be queried or something

instance.go
samliok: yep

nonvalidator/epochs.go
samliok: We initialize `epochs` with the latest indexed sealing block so we should be good.

instance_test.go
samliok: ya i meant to say drops every message that doesn't help it bootstrap.

nonvalidator/non_validator.go
samliok: so i think we still need this in the case for these scenarios

`TestNonValidatorBootstrapLatestKnownEpoch` and `TestNonValidatorBootstrapRequestsSealingBlock`

nonvalidator/non_validator.go
```diff
@@ -109,18 +119,24 @@ func NewNonValidator(config Config) (*NonValidator, error) {
...
 		lock:                  lock,
-		highestEpochCollector: newEpochReplicator(config.Logger, config.Comm),
+		highestEpochCollector: newEpochReplicator(config.Logger, config.Comm.Validators),
 		oneTimeVerifier:       simplex.NewOneTimeVerifier(config.Logger),
 		sequenceReplicator:    replicator,
-	}, nil
+	}
+	nv.sealingBlockTimeouts = common.NewTimeoutHandler(config.Logger, "sealing block replication", config.StartTime, simplex.DefaultReplicationRequestTimeout, nv.requestMissingSealingBlocks)
+	if !config.Bootstrapped {
+		nv.sealingBlockTimeouts.AddTask(startBroadcastTask)
```
yacovm: `nv.sealingBlockTimeouts.AddTask(0)` - I don't understand what we're trying to do here.

nonvalidator/non_validator.go
```diff
@@ -165,7 +197,95 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+		n.maybeTransitionToValidator(highestEpoch, validators)
+	}
+	return nil
+}
+
+// validateSealingBlock validates the epoch a sealing block opens and stores its quorum round.
+// The finalization has not been verified yet. Storing tells the replicator a valid sequence exists
+// and its validity is checked when the round is processed.
+func (n *NonValidator) validateSealingBlock(qr *common.QuorumRound, from common.NodeID) {
+	n.maybeValidateNextEpoch(qr.Block, from)
```
yacovm, line 244: It looks like we maybe validate and then anyway store the QR?

but `maybeValidateNextEpoch` can return early in many cases. Is that intentional? 

I guess the purpose of `validateSealingBlock` is to kickstart validation of the previous sealing block? If so, is this the right name for the method?

nonvalidator/epochs.go
yacovm: Not really relevant to this PR, but - if `validators` changes between two invocations of this function, then we have a problem. 

Consider two invocations `collectedSealingBlockInfo()` in `t1` and `collectedSealingBlockInfo()` in `t2`, and the first one sampled `validators = [v1, v2, v3, v4]` and the second once sampled `v2, v3, v4, v5]`'. 

The first one got a vote from `v1` and the second one got a vote from `v2` but `v1` is not in the second. 

We reached the threshold `f+1` but with an illegal count.

nonvalidator/non_validator.go
```diff
@@ -165,7 +197,95 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+	case n.epochs.canValidate(block):
+		// The sealing block in the backwards hash chain
+		n.validateSealingBlock(qr, from)
+	case n.highestEpochCollector.collectedSealingBlockInfo(sealingInfo, bh, from):
+		n.Logger.Info("A threshold of validators reported a sealing block", zap.Uint64("Seq", bh.Seq), zap.Stringer("Info", sealingInfo))
+		n.sealingBlockTimeouts.RemoveTask(startBroadcastTask)
+		if !n.isIndexed(bh.Seq) {
+			n.validateSealingBlock(qr, from)
+		}
+	default:
```
yacovm, line 229: if we have a default case that returns `nil` then how do we proceed below?

We don't, which means this is some kind of catch-all case in which we don't want to proceed.

Can we add some comments to each if case that explain better what we're trying to do?

nonvalidator/non_validator.go
```diff
@@ -165,7 +197,95 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+		return nil
+	}
+
+	// No sealing block is missing, so every epoch from our tip to the highest is validated.
+	if !n.sealingBlockTimeouts.Empty() {
+		n.finishBootstrap()
+		// If the highest epoch is already indexed, nothing more gets indexed to trigger the transition.
```
yacovm: Don't we need to replicate the blocks between the sealing blocks?

nonvalidator/non_validator.go
```diff
@@ -165,7 +197,95 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+		return nil
+	}
+
+	switch {
+	case n.epochs.canValidate(block):
+		// The sealing block in the backwards hash chain
+		n.validateSealingBlock(qr, from)
+	case n.highestEpochCollector.collectedSealingBlockInfo(sealingInfo, bh, from):
+		n.Logger.Info("A threshold of validators reported a sealing block", zap.Uint64("Seq", bh.Seq), zap.Stringer("Info", sealingInfo))
+		n.sealingBlockTimeouts.RemoveTask(startBroadcastTask)
```
yacovm: i get what you're trying to do here:

1. Can we comment to explain it so it will be more obvious
2. What if we broadcast this special 0 request and we get a sealing block with seq 100 with epoch 50, and then we ask for epoch 50 and we get the sealing block for epoch 50 from two nodes, in that case we will define the last epoch to replicate from to be epoch 50 and not epoch 100. Isn't that a problem? Should we somehow "block" retrieving the sealing block until task 0 is fulfilled?

nonvalidator/non_validator.go
yacovm: Maybe not in this PR, but can we open an issue for a follow-up PR that lets us answer replication responses? Otherwise we'll be in trouble as explained above.

nonvalidator/non_validator.go
```diff
@@ -165,7 +197,95 @@ func (n *NonValidator) HandleMessage(msg *common.Message, from common.NodeID) er
...
+func (n *NonValidator) validateSealingBlock(qr *common.QuorumRound, from common.NodeID) {
+	n.maybeValidateNextEpoch(qr.Block, from)
+	n.sequenceReplicator.StoreQuorumRound(qr)
+}
+
+// finishBootstrap marks bootstrapping done. Every epoch from our tip to the highest one a
+// threshold of validators reported is validated, so replication and live messages can be handled.
+func (n *NonValidator) finishBootstrap() {
+	n.Bootstrapped = true
+	highestEpoch, _ := n.epochs.highestEpoch()
```
yacovm: don't we need to kick-start the replication process here again to replicate the blocks we need from the beginning of the epoch to the tip?

nonvalidator/non_validator.go
yacovm: I think this comment is still relevant

nonvalidator/non_validator.go
yacovm: If we're bootstrapping but replication response is *not* nil, we can still process block messages. Can we re-order so we have `msg.ReplicationResponse != nil:` first?

nonvalidator/non_validator.go
yacovm: I am not sure i understand. If we have the epoch in the ledger why do we need to fetch it again?

common/timeout_handler.go
```diff
@@ -133,6 +133,13 @@ func (t *TimeoutHandler[T]) RemoveTask(ID T) {
 	delete(t.tasks, ID)
 }
 
+func (t *TimeoutHandler[T]) Empty() bool {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	return len(t.tasks) > 0
```
yacovm: Shouldn't this be == instead of > ?

nonvalidator/epochs.go
yacovm:
```
	switch {
	case n.epochs.canValidate(block):
		// The sealing block in the backwards hash chain
		n.validateSealingBlock(qr, from)
	case n.highestEpochCollector.collectedSealingBlockInfo(sealingInfo, bh, from):
		n.Logger.Info("A threshold of validators reported a sealing block", zap.Uint64("Seq", bh.Seq), zap.Stringer("Info", sealingInfo))
		n.sealingBlockTimeouts.RemoveTask(startBroadcastTask)
		if !n.isIndexed(bh.Seq) {
			n.validateSealingBlock(qr, from)
		}
	default:
		return nil
	}
```

If we get a block we've already indexed we may think we can validate it and then we will get into `validateSealingBlock` even though we don't need to.

We can just add a check here:

```
	switch {
	case n.epochs.canValidate(block):
		// The sealing block in the backwards hash chain
		if !n.isIndexed(bh.Seq) {
			n.validateSealingBlock(qr, from)
		}
	case n.highestEpochCollector.collectedSealingBlockInfo(sealingInfo, bh, from):
		n.Logger.Info("A threshold of validators reported a sealing block", zap.Uint64("Seq", bh.Seq), zap.Stringer("Info", sealingInfo))
		n.sealingBlockTimeouts.RemoveTask(startBroadcastTask)
		if !n.isIndexed(bh.Seq) {
			n.validateSealingBlock(qr, from)
		}
	default:
		return nil
	}
```

no?

yacovm (commented):
Made another pass

nonvalidator/non_validator.go
samliok: created https://github.com/ava-labs/Simplex/issues/606

nonvalidator/non_validator.go
samliok: i dont think we should process block messages or finalization messages until bootstrapped. Doesn't that run into the same issue as before where we can end up indexing blocks that were forged by validators from past epochs? 

HandleBlock and handleFinalization only process blocks from epochs we know anyways, and if we are behind we probably will not know of the epoch the block contains.

nonvalidator/non_validator.go
samliok: we need to confirm it is the highest epoch. otherwise how can we tell that our epoch is actually the highest epoch vs our epoch is behind?

nonvalidator/non_validator.go
yacovm: I guess it depends how we define being bootstrapped. 

I am referring to "being bootstrapped" as replicating all blocks possible until the latest known tip of the chain. 

So we should first figure out the sealing blocks of all epochs and afterwards replicate all blocks between them and only then declare that we finished bootstrapping.

nonvalidator/non_validator.go
yacovm: I'm saying that if we happen to already have `bh.Epoch` locally, we do we need to request it?

nonvalidator/non_validator.go
samliok: i updated it to 
```go
// initialBootstrapTask is the sealingBlockTimeouts task that continuously asks for sealing blocks until a
// threshold of responses validates an epoch. We use Seq 1, since requests with Seq 0 are dropped.
const initialBootstrapTask uint64 = 1
```

This timeout task only gets removed when we validate an epoch, otherwise it will keep sending replication requests. This avoids the problem where we send a request, but that request gets dropped and never delivered. 

I also moved adding the task to `Start`.

nonvalidator/non_validator.go
samliok: yea we only call `validateSealingBlock` when the sealing block can be validated. so we should store the quorum round. 

also `maybeValidateNextEpoch` is called in case the block is sealing block, not necessarily from replication.

What would u suggest the name be?

nonvalidator/non_validator.go
samliok: 1. added comments.

2. thats a good point, but i dont think we should ever request epoch `50` when bootstrapping. In the normal non-bootstrapped code, if we receive an unknown epoch we will request it's sequence. However, `processBootstrapQuorumRound` should only be requesting the highest epoch we haven't validated.

In `processBootstrapQuorumRound` we do have 

```go
if sealingInfo == nil {
	n.sendRequest(bh.Epoch, from)
	return
}
```

but this should only send requests to the highest epoch since when we started we broadcasted the special `0` request. So nodes should respond with their highest sequences. If an adversary sends us a lower epoch, we will just send a request only to that adversary and it won't be able to make the threshold of `f + 1`.

nonvalidator/non_validator.go
samliok: when we finish bootstrapping we will call `processReplicationState` which will update the replicator

```go
func (n *NonValidator) handleReplicationResponse(resp *common.ReplicationResponse, from common.NodeID) error {
	.....

	return n.processReplicationState()
}

func (n *NonValidator) processReplicationState() error {
	if !n.Bootstrapped {
		return nil
	}

	nextSeqToCommit := n.nextSeqToCommit()
	n.sequenceReplicator.MaybeAdvanceState(nextSeqToCommit, 0, 0)
```

because we have stored the sealing blocks in replication state, this will kick off replication since the replicator now knows we are behind

nonvalidator/non_validator.go
samliok: updated from bootstrapped terminology to EpochsReplicated https://github.com/ava-labs/Simplex/pull/573/commits/57c282e0cf69b8f8f6b8632c5471c179869540bc

nonvalidator/non_validator.go
samliok: updated from bootstrapped terminology to EpochsReplicated https://github.com/ava-labs/Simplex/pull/573/commits/57c282e0cf69b8f8f6b8632c5471c179869540bc

nonvalidator/non_validator.go
samliok: no, when the epochs are replicated the replicator will do this

nonvalidator/epochs.go
samliok: https://github.com/ava-labs/Simplex/issues/607

nonvalidator/non_validator.go
samliok: added

nonvalidator/non_validator.go
samliok: ya i think this is cleared up with `EpochsReplicated` so im going to resolve this

nonvalidator/epochs.go
```diff
@@ -111,6 +111,8 @@ func (e epochs) removeOldEpochs(minEpochToKeep uint64) {
 	}
 }
 
+// canValidate returns true if `block` is valid sealing block in the chain. It can
+// be valid if `block` is a backwards pointer to any sealing block already validated (backwards hash chain validation).
```
yacovm: > It can be valid if `block` is a backwards pointer to any sealing block already validated (backwards hash chain validation).

This reads weird to me. Block isn't a backwards pointer, it's a block.

Let's say instead:

`block` is valid if it's a sealing block that isn't in `e` and for which there exists a sealing block in `e` that has a backwards hash pointer to `block`.

PR #575, opened 2026-09-02
Remove Epoch From the EpochConfig

### LINKED ISSUE #536: Remove the Epoch value from EpochConfig
now that we have sealing blocks, we can properly set the epoch on initialization/start. This will make it easier on the creator of the epoch as they no longer need to override the `Epoch` number after initializing/starting.

samliok:
this simplifies the instance, and is no longer necessary with the recent addition of `block.SealingBlockInfo`

Another main benefit is we don't need to artificially set the Epoch value after initializing it. Since the epoch's config already has all the information it needs to know about its own epoch, it should be responsible for setting this value. 

We also weren't properly setting the epoch if the only blocks in storage were non-simplex blocks, so this PR adds a fix in `setMetadataFromStorage` and an instance test


simplex/epoch.go
```diff
@@ -711,6 +719,9 @@ func (e *Epoch) setMetadataFromRecords(records [][]byte) error {
 			}
 			if finalization.Finalization.Round >= highestRound {
 				highestRound = finalization.Finalization.Round
+				// We do not need to check if this finalization is associated for a sealing block, because it is impossible for a finalization
+				// for another epoch to exist in our wal, yet the original epoch remains un-sealed.
+				// This is because we issue te-locks after sealing blocks, and never move onto the next epoch until the original sealing block was indexed.
```
yacovm: it's Telocks, not te-locks.

simplex/epoch.go
```diff
@@ -711,6 +719,9 @@ func (e *Epoch) setMetadataFromRecords(records [][]byte) error {
 			}
 			if finalization.Finalization.Round >= highestRound {
 				highestRound = finalization.Finalization.Round
+				// We do not need to check if this finalization is associated for a sealing block, because it is impossible for a finalization
```
yacovm: > // We do not need to check if this finalization is associated for a sealing block, because it is impossible for a finalization
> // for another epoch to exist in our wal, yet the original epoch remains un-sealed.

That's true, but why are we writing this?

simplex/epoch_test.go
```diff
@@ -933,58 +933,60 @@ func TestEpochSimpleFlow(t *testing.T) {
...
 func TestEpochResizesBlacklistOnEpochChange(t *testing.T) {
-	epoch1Block := testutil.NewTestBlock(ProtocolMetadata{Epoch: 1, Round: 0, Seq: 0}, NewBlacklist(1))
 	nodes := []NodeID{{1}, {2}}
 	bb := testutil.NewTestBlockBuilder()
-	conf, _, _ := testutil.DefaultTestNodeEpochConfig(t, NodeID{2}, testutil.NewNoopComm(nodes), bb)
-	conf.Epoch = 2
-	require.NoError(t, conf.Storage.Index(context.Background(), epoch1Block, Finalization{}))
-	require.Equal(t, uint16(1), epoch1Block.Blacklist().NodeCount,
+
+	// The epoch number is the sequence of the last indexed sealing block, so both nodes
```
yacovm, line 1026: why is this change needed?

simplex/epoch.go
samliok: i added because at first i thought we needed to also set the epoch here, so i wrote it here since it wasn't initially obvious. Maybe a mythos scan will note this as a bug if not documented in the future

simplex/epoch_test.go
samliok: because we can no longer artificially set the epoch in the config, so we need to go through with actual messages/storage

simplex/epoch.go
yacovm: I understand that it might make sense to make this comment in the context _of this PR_.

However after the PR is merged - I think it'll just be confusing. 

There is actually a way for a finalization of a higher epoch be present in our WAL - if we have a sealing block in seq 100 but we don't have a finalization for seq 99, we will write the finalization of the sealing block in the WAL. 

I suggest we just remove this text.

samliok: sounds good, updated

samliok:
getting a determinist failure on the instance test, will look closer tomorrow. I think its because im not setting the epoch correctly when we build off a non-simplex block

simplex/util_test.go
```diff
@@ -17,12 +17,9 @@ import (
 )
 
 func TestRetrieveFromStorage(t *testing.T) {
+	nodes := []NodeID{{1}, {2}, {3}, {4}}
 	block := testutil.NewTestBlock(ProtocolMetadata{Seq: 0}, emptyBlacklist)
-	finalization := Finalization{
-		Finalization: ToBeSignedFinalization{
```
samliok, line 22: i had to update the test finalizations to contain QCs

simplex/epoch.go
```diff
@@ -699,8 +700,24 @@ func (e *Epoch) setMetadataFromStorage() error {
...
+	if e.lastBlock.VerifiedBlock.SealingBlockInfo() != nil {
+		e.Epoch = bh.Seq
+		return nil
+	}
+
+	// An indexed block without a finalization predates Simplex, so no Simplex block has
+	// been indexed and the first Simplex epoch is the sequence the first one will occupy.
+	if e.lastBlock.Finalization.QC == nil {
+		e.Epoch = e.Storage.NumBlocks()
+		e.round = e.Storage.NumBlocks()
```
yacovm: `e.round = e.Storage.NumBlocks()` - I don't think we should artificially set this

yacovm (approved):

PR #576, opened 2026-09-03
Don't verify a future proposal a second time when it is already scheduled

yacovm:
When a verification task consumes a parked finalization, the commit calls startRound, which schedules the next round's proposal from within the task. The outer maybeLoadFutureMessages loop then re-iterates for the new round and dispatches the same proposal again, because the task that would remove it from the future messages map has not run yet.

Skip proposals already marked as being processed, and clear that mark even when verification fails so a failed proposal can still be re-dispatched.

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -3131,7 +3131,7 @@ func (e *Epoch) maybeLoadFutureMessages() error {
 
 		for from, messagesFromNode := range e.futureMessages {
 			if msgs, exists := messagesFromNode[round]; exists {
-				if msgs.proposal != nil {
+				if msgs.proposal != nil && !msgs.proposalBeingProcessed {
 					if err := e.handleBlockMessage(msgs.proposal, common.NodeID(from)); err != nil {
 						return err
 					}
--- a/simplex/epoch_test.go
+++ b/simplex/epoch_test.go
@@ -2794,3 +2794,120 @@ func TestEpochVoteSentTwiceKeepsBufferedVote(t *testing.T) {
 		}
 	}
 }
+
+func TestFutureProposalDispatchedOnceAfterReentrantCommit(t *testing.T) {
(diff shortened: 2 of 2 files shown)


samliok:
> consumes a parked finalization

very claudish word

simplex/epoch.go
```diff
@@ -3109,7 +3109,7 @@ func (e *Epoch) maybeLoadFutureMessages() error {
 
 		for from, messagesFromNode := range e.futureMessages {
 			if msgs, exists := messagesFromNode[round]; exists {
-				if msgs.proposal != nil {
+				if msgs.proposal != nil && !msgs.proposalBeingProcessed {
```
samliok, line 3134: in your other pr we have something similar https://github.com/ava-labs/Simplex/pull/570/changes. where we have a pending rounds for blocks being processed.  i'm wondering if we can simplify these two bugs into one solution? 

also why can't we just delete the message from future messages when we send it to `handleBlockMessage`?

yacovm: These two bugs are unrelated, but I did use this idea 👍 : take a look [here](https://github.com/ava-labs/Simplex/pull/570).

samliok (approved):

yacovm:
```
[09-08|22:15:10.518] WARN testutil/logger.go:87 Too many blocks being verified to ingest another one {"pendingBlocks": 500}
    replication_backpressure_test.go:121: handling the replication response with a full verification queue returned: too many blocks being verified to ingest another one: 500 pending verifications (max 500)
    replication_backpressure_test.go:129: 
        	Error Trace:	/home/runner/work/Simplex/Simplex/simplex/replication_backpressure_test.go:129
        	Error:      	Condition never satisfied
        	Test:       	TestReplicationSurvivesFullVerificationQueue
        	Messages:   	node never committed the replicated sequences after the verification queue drained
```

PR #579, opened 2026-09-06
Skip parsing an empty VM block

### LINKED ISSUE #540: blockDeserializer passes nil InnerBlockBytes to VM.ParseBlock, breaking WAL restore for inner-less blocks
## Details
The library's canonical block encoding (msm.RawBlock, canoto) omits the InnerBlockBytes field for blocks that have no inner block - the zero block, all Telocks, and any block the state machine builds when the VM times out (buildBlockImpatiently), all of which are legal, protocol-mandated block types. blockDeserializer.DeserializeBlock (adapters.go:250-270) does not special-case this: it forwards the resulting nil slice straight to the embedder's VM.ParseBlock, whose contract (avalanchego VMs) is to parse a serialized block and error on garbage/empty input. Every validator writes every verified proposal to its WAL (storeProposal), and Epoch.Start() -> restoreFromWal -> restoreBlockRecord deserializes every WAL block record - before and regardless of whether the block was already indexed - and aborts startup on any deserialization error. Consequences: (1) if VM.ParseBlock(nil) errors (the typical case), any node restarting while an inner-less block record is in its WAL fails to start, and keeps failing on every retry because the WAL is persistent - a crash loop requiring manual WAL deletion; the poisoned state exists on all validators during the zero block's epoch, during every epoch transition (Telock phase), and after any VM block-building timeout, and a malicious leader can create it on demand in any round by proposing a valid inner-less normal block (accepted by verifyNormalBlock since verifyAgainstExpected uses the proposal's own nil inner block). (2) If instead the VM parses nil into a block object, the deserialized block's InnerBlock is non-nil, so its digest (StateMachineBlock.Digest hashes InnerBlock.Digest() instead of the zero hash) diverges from the identity the proposer signed, breaking notarization matching during restore (self-detected 'equivocation', round deletion, replication churn). Either way, deserialized inner-less blocks are not identity-stable with locally built ones. The correct behavior - returning a StateMachineBlock with InnerBlock == nil when InnerBlockBytes is empty - is implemented for the reverse direction (Bytes) and for verification (Verify with OnlyVM), but missing here.

## Evidence
1. [adapters.go:250–270](https://github.com/ava-labs/Simplex/blob/main/adapters.go#L250-L270)
   DeserializeBlock unconditionally calls bd.vm.ParseBlock(ctx, rawBlock.InnerBlockBytes). For zero blocks, Telocks, and any block built without an inner block, InnerBlockBytes is nil after unmarshaling, so a nil byte slice is pushed onto the embedder's VM. There is no guard mirroring ParsedBlock.Verify (external.go:59-61) or StateMachineBlock.Bytes (msm/block.go:137-139), both of which special-case a nil inner block.
2. [msm/block.canoto.go:206–217](https://github.com/ava-labs/Simplex/blob/main/msm/block.canoto.go#L206-L217)
   MarshalCanotoInto omits field 2 entirely when len(InnerBlockBytes)==0, and UnmarshalCanotoFrom (lines 122-132) rejects a present-but-empty field with ErrZeroValue. Therefore the canonical encoding of an inner-less block has no InnerBlockBytes field and deserialization always yields nil for it - the nil argument to ParseBlock is unavoidable for these blocks.
3. [simplex/epoch.go:320–337](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L320-L337)
   Reachability: restoreBlockRecord deserializes every WAL block record via BlockFromRecord -> DeserializeBlock BEFORE the already-indexed check, and any error aborts restoreFromWal, which aborts Epoch.Start() (line 300-304). If VM.ParseBlock(nil) returns an error - the normal behavior for a VM asked to parse empty bytes - the node cannot start; since the WAL persists on disk, every restart fails until the WAL is manually deleted.
4. [msm/msm.go:1399–1412](https://github.com/ava-labs/Simplex/blob/main/msm/msm.go#L1399-L1412)
   Telocks are built with wrapBlock(nil, ...) - no inner block - and are a mandatory phase of every epoch transition. Every validator that verifies a Telock proposal WALs it via storeProposal (simplex/epoch.go:3136-3179), planting an inner-less record in every validator's WAL during every epoch transition.
5. [msm/msm.go:1272–1296](https://github.com/ava-labs/Simplex/blob/main/msm/msm.go#L1272-L1296)
   buildBlockImpatiently legitimately builds blocks with a nil inner block whenever the VM times out, and verifyNormalBlock/verifyAgainstExpected accept such blocks (innerBlock nil skips inner verification). Hence a malicious leader can deliberately propose an inner-less block in ANY round of normal operation; all honest validators verify it, WAL it, and are then unable to restart until the record is garbage-collected at the next epoch transition.
6. [msm/block.go:43–56](https://github.com/ava-labs/Simplex/blob/main/msm/block.go#L43-L56)
   Identity divergence in the alternative outcome: Digest() uses the zero hash when InnerBlock is nil. If the embedder's ParseBlock(nil) instead returns a non-nil block object, the deserialized block's digest differs from the digest the proposer computed and signed, so restored blocks mismatch their WAL'd notarizations (storeNotarization then treats the node's own state as equivocation and deletes the round).
7. [msm/msm.go:828–879](https://github.com/ava-labs/Simplex/blob/main/msm/msm.go#L828-L879)
   The zero block - the first ever Simplex block - is also built with wrapBlock(nil, ...), so its WAL record is inner-less too; any restart during the first epoch (before the first epoch-transition GC removes the record) replays it through DeserializeBlock.

## Impact
A validator with an inner-less block record in its WAL cannot complete Epoch.Start after a restart: startup fails deterministically on every attempt until the WAL is manually removed, taking the validator offline (and with a malicious leader planting the record network-wide, every validator that restarts is bricked). In the non-erroring VM variant, impact degrades to state divergence and replication churn on restore. No confidentiality impact; persisted chain data is not modified.

## Reproduction steps
1. The poisoned WAL state arises without any attacker during epoch transitions, first-epoch operation, and VM timeouts; a malicious validator leader can additionally plant it on all honest nodes on demand by proposing a valid inner-less block (PR LOW for that path). The impact fires when a node restarts, and depends on the embedder VM erroring on nil bytes - both deployment/runtime conditions (attack_requirements PRESENT). Delivery is via ordinary consensus messages over the network.

## Recommended fix
1. DeserializeBlock violates the encoding's own convention that an absent InnerBlockBytes field means 'no inner block', instead delegating nil bytes to VM.ParseBlock whose contract does not cover empty input. Fix criteria: Deserializing the canonical encoding of an inner-less block (zero block, Telock, VM-timeout block) must yield a block with InnerBlock == nil, never invoke VM.ParseBlock with empty input, and produce a digest identical to the locally built block's digest. Verify by round-tripping a Telock and a zero block through Bytes()/DeserializeBlock and asserting digest equality and successful WAL restore.
2. restoreFromWal aborts node startup on any block-record deserialization failure, even for records of blocks already safely indexed, converting a single bad record into a persistent startup failure. Fix criteria: WAL restore should be resilient: records for already-indexed sequences should not require successful inner-block parsing to boot, and an isolated undecodable record should degrade to replication-based recovery rather than preventing startup. Verify a node restarts successfully with a WAL containing a record its VM cannot parse.

---
**Severity:** HIGH
**Status:** Open
**Category:** Improper input handling
**CWE:** [CWE-754](https://cwe.mitre.org/data/definitions/754.html)
**Branch:** main
**Date created:** 2026-08-21

---

yacovm:
This commit makes DeserializeBlock skip parsing an empty inner block when recovering from the WAL

### DIFF
--- a/instance_helpers_test.go
+++ b/instance_helpers_test.go
@@ -549,6 +549,7 @@ func noopICMTransition(_ metadata.ICMEpochInput) metadata.ICMEpochInfo {
 
 type node struct {
 	t       *testing.T
+	stopped bool
 	net     *network
 	id      common.NodeID
 	vm      *blockBuilderVM
@@ -560,8 +561,36 @@ type node struct {
 
 // stop stops the instance, then drains the messages the node was sending.
 func (n *node) stop() {
+	if n.stopped {
+		return
+	}
 	n.inst.Stop()
 	n.comm.stop()
+	n.stopped = true
+}
+
+func (n *node) restart() *node {
+	n.stop()
+
+	var records [][]byte
+	n.wals.lock.Lock()
+	for _, w := range n.wals.wals {
+		walRecords, err := w.ReadAll()
+		require.NoError(n.t, err)
+		records = append(records, walRecords...)
+	}
+	n.wals.lock.Unlock()
+
+	// Come back up over the same storage, restoring a WAL rebuilt from the exported records.
+	restoredWAL := testutil.NewTestWAL(n.t)
+	for _, record := range records {
+		require.NoError(n.t, restoredWAL.Append(record))
+	}
+
+	newNode := n.net.addNodeWithConfig(n.id, nodeConfig{
+		wals: []wal.DeletableWAL{restoredWAL}, storage: n.storage, existingNode: true,
+	})
+	return newNode
 }
 
 // sync syncs a node by waiting for the commit of the latest sequence.
@@ -625,6 +654,8 @@ type nodeConfig struct {
 	storage *testStorage
 	// wals are pre-existing WALs the instance restores on start.
 	wals []wal.DeletableWAL
+	// existingNode indicates whether the node is being added to the network for the first time (false) or is a restart of an existing node (true).
+	existingNode bool
 }
 
 // addNode creates and starts a node in the network.
@@ -637,11 +668,6 @@ func (n *network) addNodeWithConfig(id common.NodeID, cfg nodeConfig) *node {
 	storage := cfg.storage
 	if storage == nil {
 		storage = newTestStorageWithGenesis(n.t)
-	}
-
-	// ensure a unique id; snapshot because nodes may be added concurrently
-	for _, node := range n.nodesSnapshot() {
-		require.NotEqual(n.t, node.id, id)
 	}
 
 	comm := newInstanceComm(n, id)
@@ -677,7 +703,20 @@ func (n *network) addNodeWithConfig(id common.NodeID, cfg nodeConfig) *node {
 	}
 
 	n.lock.Lock()
-	n.nodes = append(n.nodes, node)
+	if cfg.existingNode {
+		for i, existingNode := range n.nodes {
+			if existingNode.id.Equals(id) {
+				n.nodes[i] = node
+				break
+			}
+		}
+	} else {
+		// ensure a unique id; snapshot because nodes may be added concurrently
+		for _, node := range n.nodes {
+			require.NotEqual(n.t, node.id, id)
+		}
+		n.nodes = append(n.nodes, node)
+	}
 	n.lock.Unlock()
 
 	instance.Config.Logger.Debug("Created a node in the test network", zap.Uint64("Seq", n.seq), zap.Uint64("num block", node.storage.NumBlocks()))
--- a/instance_test.go
+++ b/instance_test.go
@@ -11,6 +11,7 @@ import (
 	metadata "github.com/ava-labs/simplex/msm"
 	"github.com/ava-labs/simplex/simplex"
 	"github.com/ava-labs/simplex/testutil"
+	"github.com/ava-labs/simplex/wal"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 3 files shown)


instance_test.go
```diff
@@ -323,6 +324,61 @@ func TestInstanceValidatorSkipsAnEpoch(t *testing.T) {
...
+		WalCreator:               wc.createWAL,
+		Storage:                  node.storage,
+		Logger:                   testutil.MakeLogger(t, 1),
+		WALs:                     []wal.DeletableWAL{restoredWAL},
+		VM:                       node.vm,
+		ICMETransition:           noopICMTransition,
+		ID:                       validator.NodeID[:],
+	})
+	t.Cleanup(restarted.Stop)
+	require.NoError(t, restarted.Start(t.Context()), "a node that built the zero block cannot restart")
```
samliok: i think we should make restarting a node a helper function, so that we can easily re-use it across tests

yacovm: https://github.com/ava-labs/Simplex/pull/579/commits/56650467ecdca48f8fa558d552e7e02dc53201ac

instance_helpers_test.go
```diff
@@ -677,7 +703,20 @@ func (n *network) addNodeWithConfig(id common.NodeID, cfg nodeConfig) *node {
...
-	n.nodes = append(n.nodes, node)
+	if cfg.existingNode {
+		for i, node := range n.nodes {
+			if node.id.Equals(id) {
+				n.nodes[i] = node
+				break
+			}
+		}
+	} else {
+		// ensure a unique id; snapshot because nodes may be added concurrently
```
samliok, line 722: i think i removed the fact that nodes may be added concurrently

instance_test.go
```diff
@@ -323,6 +323,26 @@ func TestInstanceValidatorSkipsAnEpoch(t *testing.T) {
...
+
+	zeroBlock, _, err := node.storage.GetBlock(1)
+	require.NoError(t, err)
+	require.Equal(t, metadata.BlockTypeZero, zeroBlock.Type())
+	require.Nil(t, zeroBlock.InnerBlock)
+
+	node = node.restart()
+
+	// The restarted node keeps the chain going.
+	network.pending.addPendingBlock()
```
samliok: why don't we just do network.acceptNewBlock?

instance_helpers_test.go
yacovm: This is inside the lock...

instance_test.go
yacovm: fixed

samliok (approved):

PR #580, opened 2026-09-06
Standardize VerifySignature parameters

### LINKED ISSUE #436: SignatureVerifier Consolidation
so we have a lot of different declarations of verify signature.

```go
// api.go
type SignatureVerifier interface {
	VerifySignature(message []byte, signature []byte, publicKey []byte) error
}
```

```go
// config.Go <- crypto ops interface
VerifySignature(message []byte, signature []byte, publicKey []byte) error
```

```go
//msm.go
type SignatureVerifier interface {
	VerifySignature(signature []byte, message []byte, publicKey []byte) error
}
```

does it make sense to consolidate this into one? i think there may be a current bug in our test code since we use the same verifier to implement all 3 interfaces, even though the function signatures are different. 

_Originally posted by @yacovm in https://github.com/ava-labs/Simplex/pull/427#discussion_r3574074928_

### LINKED ISSUE #548: VerifySignature message/signature arguments transposed between common and msm interfaces
## Details
The repository defines two structurally identical but semantically contradictory signature-verification interfaces: common.SignatureVerifier declares VerifySignature(message, signature, publicKey) (common/api.go:78), while metadata (msm).SignatureVerifier declares VerifySignature(signature, message, publicKey) (msm/msm.go:121). Since all parameters are []byte, Go's structural typing allows a single implementation to satisfy both.

Instance.createEpochConfig wires the one Config.CryptoOps object into both consumers: instance.go:447 passes it as the msm state machine's SignatureVerifier and instance.go:509 passes it as the Epoch's common.SignatureVerifier. The CryptoOps contract (config.go:77) documents the message-first order, matching the common interface.

Consequently, every msm verification call passes its arguments in an order the implementation does not expect:
- msm/approvals.go:156 verifies validator epoch-transition approvals as VerifySignature(approval.Signature, toBeSigned, pk);
- msm/msm.go:1103 verifies the aggregated next-epoch certificate carried by sealing blocks as VerifySignature(next.NextEpochApprovals.Signature, toBeSigned, aggPK).

With any real scheme (BLS12-381 or ECDSA), interpreting the 40-byte approval payload (8-byte P-chain height plus 32-byte aux-info digest) as a signature fails parsing/verification deterministically, so the checks fail closed: all legitimate approvals are dropped by ApprovalStore.HandleApproval and all sealing blocks fail verifyNextEpochCertificate. Epoch transitions can therefore never be approved or verified, halting the consensus network at its first epoch transition. Conversely, an implementation written to the msm order would break the common-path vote verification (common/msg.go:152, used by epoch.go vote/finalize-vote/empty-vote handling), halting consensus entirely. No single implementation can satisfy both call sites, so one security check family is guaranteed to be computed on transposed inputs.

The defect is invisible to the existing test suite: the only cryptographically real test verifier (msm/util_test.go:131) uses the msm-local order, and the cross-package integration doubles (instance_test.go:1230 testCryptoOps, testutil/util.go:197 TestVerifier) ignore their message/signature arguments altogether.

Because verification fails closed for standard schemes, this is an availability/liveness defect of the security function rather than a signature-forgery bypass; no scheme accepted here treats a 40-byte payload as a valid signature.

## Evidence
1. [msm/msm.go:118–122](https://github.com/ava-labs/Simplex/blob/main/msm/msm.go#L118-L122)
   Defines metadata.SignatureVerifier with the convention VerifySignature(signature, message, publicKey) - signature first. This is the outlier: both the embedder-facing CryptoOps contract (config.go:77) and common.SignatureVerifier (common/api.go:78) use message-first. Because all three parameters are []byte, Go's structural typing lets one concrete implementation satisfy both interfaces even though their semantics contradict each other, so the transposition compiles silently. This interface (and its two call sites) is the code a fix must change.
2. [common/api.go:77–79](https://github.com/ava-labs/Simplex/blob/main/common/api.go#L77-L79)
   Defines common.SignatureVerifier with the convention VerifySignature(message, signature, publicKey) - message first. This is the convention consumed by the consensus vote/finalization verification path (common/msg.go verifyContext, used at simplex/epoch.go:986, 1146, 1211, 2371 via EpochConfig.Verifier).
3. [instance.go:442–447](https://github.com/ava-labs/Simplex/blob/main/instance.go#L442-L447)
   Instance.createEpochConfig passes the single Config.CryptoOps object verbatim as the msm state machine's SignatureVerifier (line 447), whose call sites are signature-first. No adapter swaps the arguments anywhere in the repository.
4. [instance.go:509](https://github.com/ava-labs/Simplex/blob/main/instance.go#L509)
   The same Config.CryptoOps object is simultaneously wired as the Epoch's Verifier, typed common.SignatureVerifier (simplex/epoch.go:78), whose call sites are message-first. Because the identical function value serves both conventions with no reordering adapter, exactly one of the two verification families always invokes the implementation with message and signature transposed.
5. [config.go:73–79](https://github.com/ava-labs/Simplex/blob/main/config.go#L73-L79)
   The embedder-facing CryptoOps interface documents the message-first order (VerifySignature(message, signature, publicKey)), matching common.SignatureVerifier. Any implementation written against this documented contract receives transposed arguments on every msm call.
6. [msm/approvals.go:149–156](https://github.com/ava-labs/Simplex/blob/main/msm/approvals.go#L149-L156)
   checkApprovalSignature calls VerifySignature(approval.Signature, toBeSigned, pk) - signature first. With a contract-conforming message-first CryptoOps implementation, this asks whether the deterministic ASN.1-marshaled approval payload is a valid signature over the peer-supplied signature bytes, which deterministically fails (the payload cannot parse as a BLS point or ECDSA SEQUENCE{INTEGER,INTEGER}), so ApprovalStore.HandleApproval silently drops every legitimate validator epoch-transition approval.
7. [msm/msm.go:1098–1106](https://github.com/ava-labs/Simplex/blob/main/msm/msm.go#L1098-L1106)
   verifyNextEpochApprovalsSignature calls VerifySignature(next.NextEpochApprovals.Signature, toBeSigned, aggPK) - signature first. Reached in production via ParsedBlock.Verify (external.go:65) -> msm.VerifyBlock -> verifyCollectingApprovalsBlock (msm.go:1055). With a message-first implementation this always fails, and since blocks in the approval-collection phase must carry approvals once auxiliary info is ready (errEmptyNextEpochApprovals, msm.go:1031-1034), every proposable block is rejected by honest verifiers, halting consensus at every epoch transition.
8. [common/msg.go:146–153](https://github.com/ava-labs/Simplex/blob/main/common/msg.go#L146-L153)
   verifyContext calls verifier.VerifySignature(toBeSigned, signature, pk) - message first, per the common interface. This is the vote/finalize-vote/empty-vote verification path (epoch.go:986, 1146, 1211, 2371). If the embedder instead implements the msm convention to make approvals work, this entire path breaks and consensus halts immediately, so no single implementation can be correct for both call families.
9. [msm/util\_test.go:127–149](https://github.com/ava-labs/Simplex/blob/main/msm/util_test.go#L127-L149)
   The only test double that actually performs cryptographic verification (msm's) implements the signature-first order, while the integration-level testCryptoOps (instance_test.go:1230) and testutil.TestVerifier (testutil/util.go:197) ignore their message/signature arguments entirely. The test suite therefore cannot detect the transposition, leaving the defect latent until production wiring with a real verifier.

## Impact
The affected checks fail closed for all realistic schemes (a 40-byte payload cannot parse as a BLS/ECDSA signature), so there is no forgery or data exposure. The impact is that epoch-transition approvals and sealing-block certificates can never verify: epoch transitions stall permanently, a network-wide consensus liveness halt (or, with the opposite implementation convention, all vote verification fails and consensus halts immediately). That is a serious, deterministic loss of the service's availability.

## Reproduction steps
1. No attacker action is strictly required: the transposed verification executes whenever the network processes epoch-transition approvals or sealing blocks (normal protocol operation over the network). Any remote peer or validator sending an approval or sealing block exercises the broken check. The condition depends on the deployment's CryptoOps implementation following the documented message-first contract and on the network reaching an epoch transition, so a deployment condition is present. No privileges or user interaction are needed.

## Recommended fix
1. Two same-named interfaces (common.SignatureVerifier and msm.SignatureVerifier) declare VerifySignature with transposed message/signature parameter order, and instance.go wires one CryptoOps implementation into both, so one family of signature checks always receives swapped arguments. Fix criteria: A single, consistent parameter order for VerifySignature is used across common, msm, config.CryptoOps, and all call sites (or an explicit adapter reorders arguments at the boundary). Verified when a real (argument-sensitive) verifier implementation passes both the msm approval/next-epoch-certificate verification path and the common vote-verification path against the same signing convention.
2. Test doubles mask the defect: the integration-level testCryptoOps and testutil.TestVerifier ignore the message and signature arguments, so the transposition cannot be caught by the existing suite. Fix criteria: Integration tests use an argument-sensitive verifier (one that actually checks the signature over the message) wired through Instance's CryptoOps, and they exercise both an epoch transition (approvals plus sealing-block certificate) and normal vote verification; the suite fails if either path receives transposed arguments.

---
**Severity:** HIGH
**Status:** Open
**Category:** Incorrect argument order
**CWE:** [CWE-683](https://cwe.mitre.org/data/definitions/683.html)
**Branch:** main
**Date created:** 2026-08-21

---

yacovm:


samliok:
i think I also noted this issue a while back https://github.com/ava-labs/Simplex/issues/436

instance_helpers_test.go
```diff
@@ -189,23 +189,33 @@ func (pc *testPlatformChain) LastNonSimplexBlockPChainHeight() uint64 {
 	return pc.genesisHeight
 }
 
+// testCryptoOps signs a message with its digest, and verifies a signature by recomputing that digest,
+// so a call that transposes the message and the signature fails verification instead of passing silently.
```
samliok: idk if this comment is necessary

samliok (approved):

PR #583, opened 2026-09-08
Keep the first buffered finalize vote per node and round

samliok:
Store the first finalized vote in future messages


simplex/epoch_test.go
```diff
@@ -3116,3 +3116,48 @@ func TestReplicationRequestsKeepHighestObservedRound(t *testing.T) {
...
+	require.NoError(t, err)
+	t.Cleanup(e.Stop)
+	require.NoError(t, e.Start())
+
+	b, ok := bb.BuildBlock(context.Background(), e.Metadata(), emptyBlacklist)
+	require.True(t, ok)
+	block := b.(Block)
+
+	// We have no round for these votes yet, so they are held without being verified.
+	testutil.InjectTestFinalizeVote(t, e, b, nodes[2])
```
yacovm: can we make a table test in which we flip the order and once we should expect to commit, once we should expect not to commit?

yacovm (approved):

PR #586, opened 2026-09-08
Bound replication state memory and delete old sequences

### LINKED ISSUE #558: Non-validator replication path lacks MaxSequenceWindow bound, allowing unbounded block buffering
## Details
The non-validator's replication-response path verifies each QuorumRound's finalization and then stores the full block and finalization in ReplicationState.seqs (StoreQuorumRound -> storeSequence) keyed by peer-chosen sequence number. Neither handleReplicationResponse nor processQuorumRound enforces the MaxSequenceWindow look-ahead bound that (a) the same file enforces for direct block messages (non_validator.go:180) and finalizations (non_validator.go:340), and (b) the validator's handleReplicationResponse enforces for the identical message type (simplex/epoch.go:3436-3440).

Because the finalizations must verify, an attacker cannot fabricate content, but every finalized block of the real chain beyond the victim's commit height is valid ammunition and publicly replayable by any peer. A non-validator that is far behind (bootstrap or extended downtime on a long chain) can be streamed the entire remaining chain: each QuorumRound (blocks up to ~MB) is retained in the unbounded seqs map far ahead of the commit pointer, while committing is throttled by sequential VM verification and indexing. Memory grows with the full span between the victim's height and the chain tip (potentially millions of blocks) instead of the configured window (default 10), leading to memory exhaustion and an OOM crash. Entries below nextSeqToCommit that are skipped (e.g., digest-superseded) are also never garbage-collected except by exact-match DeleteSeq.

This is a missing-bound logic flaw (the documented window is simply not applied on this path), not generic volumetric flooding: a single message burst is retained indefinitely rather than processed and released.

## Evidence
1. [nonvalidator/non\_validator.go:419–433](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/non_validator.go#L419-L433)
   handleReplicationResponse iterates all QuorumRounds in resp.Data and also processes resp.LatestSeq, without any check that a sequence is within MaxSequenceWindow of nextSeqToCommit, and without matching responses to outstanding requests (unsolicited responses are fully processed).
2. [nonvalidator/non\_validator.go:469–509](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/non_validator.go#L469-L509)
   processQuorumRound checks only staleness (isAccepted, a lower bound) and QC validity before StoreQuorumRound; there is no upper-bound window check on the block's sequence, unlike handleBlock (line 180) and handleFinalization (line 340) which enforce MaxSequenceWindow for the same node.
3. [simplex/replication\_state.go:122–136](https://github.com/ava-labs/Simplex/blob/main/simplex/replication_state.go#L122-L136)
   storeSequence inserts a full block plus finalization into the seqs map keyed by sequence with no cap on map size; maxRoundWindow is only used by the requestor for issuing requests, not for bounding stored data. The only eviction is exact-match DeleteSeq(nextSeqToCommit) as commits advance.
4. [simplex/epoch.go:3435–3447](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L3435-L3447)
   The validator's equivalent handler explicitly drops quorum rounds whose sequence exceeds nextSeqToCommit+MaxRoundWindow, demonstrating the intended bound that the non-validator path omits.
5. [nonvalidator/non\_validator.go:138–165](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/non_validator.go#L138-L165)
   NonValidator.HandleMessage dispatches ReplicationResponse from any sender with no validator-membership check, unlike the validator's Epoch.HandleMessage (epoch.go:168-178) which restricts non-validator senders to ReplicationRequest only. This makes the sink reachable by any network peer (privileges_required NONE).
6. [nonvalidator/non\_validator.go:46–47](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/non_validator.go#L46-L47)
   Config documents MaxSequenceWindow as 'How many sequences we allow to look past our next sequence to commit', confirming the replication path violates the intended memory bound.

## Impact
Unbounded retained allocations lead to memory exhaustion and process crash (OOM) of the non-validator during catch-up; recoverable by restart but re-triggerable while it remains behind. No data disclosure or corruption.

## Reproduction steps
1. Any network peer can send unsolicited ReplicationResponse messages to a non-validator: NonValidator.HandleMessage has no sender-membership gate (unlike the validator path, which only accepts ReplicationRequest from non-validators). The attacker replays real finalized quorum rounds — publicly available chain data — for sequences far beyond the victim's commit height; each passes QC verification and is retained in full. Meaningful impact requires the victim to be substantially behind the chain tip (bootstrap/extended downtime on a long chain), and the attacker must stream data volume comparable to the memory consumed.

## Recommended fix
processQuorumRound/handleReplicationResponse store verified quorum rounds at any future sequence, ignoring the MaxSequenceWindow bound that all other ingestion paths enforce, so replication memory is bounded by chain length instead of the configured window. Fix criteria: Quorum rounds with sequences beyond nextSeqToCommit+MaxSequenceWindow must be rejected (or only their metadata recorded) before storage, mirroring the validator's handler, and stale entries must be pruned as the commit pointer advances. Verify that streaming far-future finalized rounds leaves the seqs map bounded by the configured window.

---
**Severity:** MEDIUM
**Status:** Open
**Category:** Allocation without limits
**Branch:** main
**Date created:** 2026-08-21

---

samliok:
Addresses https://github.com/ava-labs/Simplex/issues/558

### DIFF
--- a/nonvalidator/non_validator.go
+++ b/nonvalidator/non_validator.go
@@ -522,8 +522,22 @@ func (n *NonValidator) processQuorumRound(qr *common.QuorumRound, from common.No
 
 	// This block could be a sealing block, validate the next epoch if so.
 	n.maybeValidateNextEpoch(block)
-	n.sequenceReplicator.StoreQuorumRound(qr)
+	n.storeQuorumRound(qr)
 	return nil
+}
+
+// storeQuorumRound updates replication state, and stores qr if its within MaxSequenceWindow.
+func (n *NonValidator) storeQuorumRound(qr *common.QuorumRound) {
+	seq := qr.Block.BlockHeader().Seq
+	nextSeqToCommit := n.nextSeqToCommit()
+
+	if seq > n.MaxSequenceWindow+nextSeqToCommit {
+		n.Logger.Debug("Received a quorum round from a sequence too far ahead", zap.Uint64("Next Seq To Commit", nextSeqToCommit), zap.Uint64("Block Sequence", seq))
+		n.sequenceReplicator.ReceivedFutureFinalization(qr.Finalization, nextSeqToCommit)
+		return
+	}
+
+	n.sequenceReplicator.StoreQuorumRound(qr)
 }
 
 func (n *NonValidator) handleQrFromUnknownEpoch(qr *common.QuorumRound, from common.NodeID) {
@@ -539,14 +553,9 @@ func (n *NonValidator) handleQrFromUnknownEpoch(qr *common.QuorumRound, from com
 	if n.epochs.canValidate(block) {
 		n.Logger.Debug("We can validate an epoch block as we have validated the one after it.", zap.Stringer("Info", block.SealingBlockInfo()))
 		n.maybeValidateNextEpoch(block)
-
-		// We are storing a quorum round with a finalization we have not yet verified.
-		// We do this to tell the replicator a valid sequence exists and to begin replication if necessary.
-		// We will check the validity when we process this round.
-		n.sequenceReplicator.StoreQuorumRound(qr)
+		n.storeQuorumRound(qr)
 		return
 	}
-
 	if n.highestEpochCollector.collectedSealingBlockInfo(qr.Block.SealingBlockInfo(), qr.Block.BlockHeader(), from) {
 		n.Logger.Debug("We can validate an epoch because we have received a threshold of messages of it.", zap.Stringer("Info", block.SealingBlockInfo()))
 		n.maybeValidateNextEpoch(block)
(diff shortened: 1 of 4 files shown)


nonvalidator/non_validator.go
```diff
@@ -536,26 +550,21 @@ func (n *NonValidator) handleQrFromUnknownEpoch(qr *common.QuorumRound, from com
...
+	switch {
+	case n.epochs.canValidate(block):
 		n.Logger.Debug("We can validate an epoch block as we have validated the one after it.", zap.Stringer("Info", block.SealingBlockInfo()))
-		n.maybeValidateNextEpoch(block)
-
-		// We are storing a quorum round with a finalization we have not yet verified.
-		// We do this to tell the replicator a valid sequence exists and to begin replication if necessary.
-		// We will check the validity when we process this round.
-		n.sequenceReplicator.StoreQuorumRound(qr)
+	case n.highestEpochCollector.collectedSealingBlockInfo(qr.Block.SealingBlockInfo(), qr.Block.BlockHeader(), from):
```
yacovm: It seems to me that the left side is equivalent to the right side, it's just a refactor isn't it? 

If so, can we please revert to the previous version? 

If we want to refactor stuff we can do that in separate PRs.

samliok: removed

yacovm (approved):

PR #587, opened 2026-09-09
Remove the Epoch From Simplex Epoch Info

samliok:
The value should always be the same as the value in the protocol metadata, so there is no point in having a duplicate field


msm/encoding.go
```diff
@@ -83,7 +83,7 @@ func (ei *ICMEpochInfo) Equal(other *ICMEpochInfo) bool {
 		return other == nil
 	}
 	if other == nil {
-		return ei == nil
+		return false
```
yacovm, line 86: why?

msm/encoding.go
yacovm, line 100: we should decrement all field numbers by 1

yacovm (dismissed):

yacovm:
LGTM otherwise

msm/encoding.go
samliok: because ei can never be nil here

yacovm (approved):

PR #588, opened 2026-09-09
Don't propagate a full verification scheduler out of HandleMessage

samliok:
Claude noted this as different then how we handle block messages and I agreed


yacovm:
what about the non-validator?

samliok:
> what about the non-validator?

I could add but i dont think its likely to happen for a non-validator. We only schedule blocks that are nextSeqToCommit, so if our block scheduler is full something else is probably wrong

yacovm (approved):

PR #589, opened 2026-09-10
Remove early return in verify block message

### LINKED ISSUE #546: Vote overwrite skips signature verification, poisoning notarization QC aggregation
## Details
Simplex builds notarization quorum certificates in common.NewNotarization by aggregating the Signature of every stored vote whose header matches the proposal. NewNotarization performs no signature verification (common/notarization.go:48-75); it relies on the invariant that every entry of Round.votes was verified on insertion.

handleVoteMessage (simplex/epoch.go:1136-1155) breaks that invariant. To avoid re-verifying duplicates it verifies the signature only when no vote from that signer is stored (line 1145), then unconditionally overwrites the stored vote with the newly received message (line 1152) without checking that the new message equals the previously verified one. The from==Signature.Signer gate (line 1094) restricts the attack to the attacker's own slot: a malicious validator M first sends a valid vote for the round's proposal, then immediately a second message carrying the same ToBeSignedVote but arbitrary Signature.Value. The second message skips verification and replaces the verified vote. (The finalize-vote path shows the intended invariant: isFinalizationVoteValid verifies every incoming message unconditionally.)

From then on, maybeCollectNotarization (epoch.go:1613-1648) counts M's signer ID toward quorum (IsQuorum over signer IDs) and NewNotarization iterates the whole vote map, so the garbage signature is included in the aggregate even when quorum-many honest votes exist. The SignatureAggregator contract (common/api.go:152-165) does not require verification during aggregation (BLS aggregation is point addition), so either:
- Aggregate fails to parse the bytes and returns an error, which propagates out of handleVoteMessage to the embedding application via HandleMessage; the poisoned vote stays in the map, so the round can never be vote-notarized at the victim and falls back to timeout/empty notarization; or
- Aggregate succeeds (e.g., a well-formed BLS point that is not a signature over the vote), and the victim persists an invalid notarization QC to its WAL, broadcasts it (all peers reject it in VerifyQC, epoch.go:1783), advances its round on it, credits M in redeemedRounds without a valid signature (epoch.go:1675-1686), and later serves the unverifiable QC to replicating peers.

A related instance exists in VerifyBlockMessageVote (epoch.go:2347-2354): the leader's proposal vote is not re-verified when a buffered proposal with an equal block header exists, even though the re-sent message may carry different signature bytes; that unverified vote is later stored into round.votes by the block verification task (epoch.go:2188).

By performing the two-message sequence against every honest node each round, a single malicious validator — well below the f Byzantine nodes the protocol is designed to tolerate — prevents vote-based notarization network-wide, forcing every round through timeout/empty notarization and halting block production while remaining an apparently honest voter, and pollutes WALs and replication state with quorum certificates that fail verification.

## Evidence
1. [simplex/epoch.go:1136–1155](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L1136-L1155)
   handleVoteMessage verifies a vote's signature only when no vote from that signer is stored yet (line 1145: `if _, exists := round.votes[string(signature.Signer)]; !exists { Verify }`) and then unconditionally overwrites the stored vote with the incoming message at line 1152, with no check that the new message equals the previously verified one. A validator can therefore replace its own verified vote with one carrying the same ToBeSignedVote but arbitrary Signature.Value. The from==Signature.Signer gate at line 1094 limits replacement to the signer's own slot, so the attacker must be a validator poisoning its own entry.
2. [common/notarization.go:42–77](https://github.com/ava-labs/Simplex/blob/main/common/notarization.go#L42-L77)
   NewNotarization is the QC assembly sink: it iterates the entire votesForCurrentRound map, keeps every vote whose header matches the proposal, and passes each vote's Signature directly to signatureAggregator.Aggregate. It performs no signature verification itself, relying entirely on the (broken) invariant that everything in round.votes was verified. The poisoned signature is therefore aggregated into the quorum certificate even when quorum-many honest votes exist.
3. [simplex/epoch.go:1613–1648](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L1613-L1648)
   maybeCollectNotarization counts the poisoned vote toward the quorum check (IsQuorum over signer IDs at line 1629, not verified signatures) and then calls common.NewNotarization with the full votes map, so the garbage signature both helps reach quorum and is included in the aggregate even when enough honest votes exist.
4. [simplex/epoch.go:1666–1710](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L1666-L1710)
   persistNotarization and persistAndBroadcastNotarization show the consequence when Aggregate succeeds on the garbage signature: the self-assembled QC is written to the WAL, every QC signer (including the attacker) receives redemption credit (lines 1675-1686), the round is advanced, and the notarization is broadcast to all peers — without VerifyQC ever being run on it (contrast handleNotarizationMessage at line 1783, which verifies QCs received from peers and rejects invalid ones). When Aggregate instead errors, the error propagates out of handleVoteMessage/HandleMessage and the round can never be vote-notarized because the poisoned vote stays in the map.
5. [simplex/epoch.go:2344–2354](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L2344-L2354)
   Same weakness class in the proposal path: VerifyBlockMessageVote skips verifying the leader's vote whenever a buffered proposal for the round has an equal block header, without binding the vote's signature bytes to the previously verified message. A leader re-sending its proposal with different, unverified vote signature bytes can have that vote stored into round.votes by the block verification task (round.votes[string(vote.Signature.Signer)] = &vote at line 2188).

## Impact
Availability: every notarization the victim assembles includes the garbage signature — Aggregate either errors (the round can never be vote-notarized; the error surfaces out of HandleMessage) or yields an invalid QC. Poisoning all honest nodes each round forces every round through timeout/empty notarization, halting block production indefinitely. Integrity: in the invalid-QC case the victim persists the unverifiable notarization to its WAL, broadcasts it (peers reject it via VerifyQC), advances its round on it, grants the attacker redemption credit without a valid signature, and serves it to replicating peers; finalize votes are independently verified, so finalization safety is preserved and integrity loss is bounded.

## Reproduction steps
1. Attacker is one validator in the set, sending ordinary network consensus messages. After the victim stores the round's proposal, the attacker sends a correctly signed vote (verified and stored), then immediately a second vote with identical ToBeSignedVote but arbitrary signature bytes; because a vote from that signer already exists, verification is skipped and the stored vote is overwritten. The only timing constraint — landing before the round is notarized — is trivially met by sending back-to-back upon seeing the proposal (honest voters are still verifying the block), and the attack is repeatable every round against every node.

## Recommended fix
1. handleVoteMessage treats the existence of a stored vote from a signer as proof that the incoming message was verified, then overwrites the stored vote with the unverified message, violating the invariant that Round.votes contains only verified signatures. Fix criteria: A vote message whose signature has not been verified must never replace or be stored alongside a verified vote. Either verify every incoming vote message before storing, or drop incoming votes for signers that already have a stored vote (or only skip verification when the new message is byte-identical to the stored one). Verify that sending a valid vote followed by a same-signer vote with corrupted signature bytes leaves the originally verified vote in place and still yields a valid, verifiable notarization QC.
2. common.NewNotarization aggregates caller-supplied signatures into a quorum certificate without any verification and without the assembled QC ever being self-checked before it is persisted to the WAL, broadcast, and used to advance the round. Fix criteria: The consensus engine must not persist, broadcast, or act on a self-assembled QC that would fail its own VerifyQC; defense in depth such as verifying the aggregated QC (or the individual signatures) before storeNotarization/persistAndBroadcastNotarization. Verify that a QC assembled from a vote set containing one invalid signature is rejected locally and never reaches the WAL or the network.
3. VerifyBlockMessageVote skips verification of a re-sent proposal's vote based only on block-header equality, without binding the vote's signature bytes to the previously verified message. Fix criteria: Re-sent proposals must have their accompanying vote verified unless the entire message (including signature bytes) matches the previously verified copy. Verify that a leader re-sending its proposal with altered vote signature bytes cannot get the altered vote stored in Round.votes.

---
**Severity:** HIGH
**Status:** Open
**Category:** Improper signature verification
**CWE:** [CWE-347](https://cwe.mitre.org/data/definitions/347.html)
**Branch:** main
**Date created:** 2026-08-21

---

samliok:
the previous check was a slight optimization that said "if a proposal was already sent and is in future messages, don't verify the vote". 

This was broken as noted in #546, because an adversary could just send us a invalid vote and override the valid one. 

IMO the check for future messages doesn't save us much, and is not worth the additional complexity. Best case it saves us one signature verification if a malicious node sends us two blocks for the same round.

Closes #546

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2413,8 +2413,11 @@ func (e *Epoch) verifyBlockMessageVote(from common.NodeID, md common.BlockHeader
 		return fmt.Errorf("received a finalization from an unknown node %s", from)
 	}
 
-	// Ensure the block was voted on by its block producer:
+	if !md.Equals(&vote.Vote.BlockHeader) {
+		return errors.New("vote block header does not match block header")
+	}
 
+	// Ensure the block was voted on by its block producer:
 	// 1) Verify block digest corresponds to the digest voted on
 	if !bytes.Equal(vote.Vote.Digest[:], md.Digest[:]) {
 		e.Logger.Debug("ToBeSignedVote digest mismatches block digest", zap.Stringer("voteDigest", vote.Vote.Digest),
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -1933,7 +1933,7 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
 	}
 
 	// Check if we have verified this message in the past:
-	if err := e.VerifyBlockMessageVote(from, md, vote); err != nil {
+	if err := e.verifyBlockMessageVote(from, md, vote); err != nil {
 		return nil
 	}
 
@@ -2405,18 +2405,8 @@ func (e *Epoch) createNotarizedBlockVerificationTask(block common.Block, notariz
 	}
 }
 
-// VerifyBlockMessageVote checks if we have the block in the future messages map.
-// If so, it means we have already verified the vote associated with this proposal.
-// If not, it verifies that the vote corresponds to the block proposed, and that the vote is properly signed.
-func (e *Epoch) VerifyBlockMessageVote(from common.NodeID, md common.BlockHeader, vote common.Vote) error {
-	msgsForRound, exists := e.futureMessages[string(from)][md.Round]
-	if exists && msgsForRound.proposal != nil {
-		bh := msgsForRound.proposal.Block.BlockHeader()
-		if bh.Equals(&md) {
-			return nil
-		}
-	}
-
+// verifyBlockMessageVote verifies that the vote corresponds to the block proposed, and that the vote is properly signed.
+func (e *Epoch) verifyBlockMessageVote(from common.NodeID, md common.BlockHeader, vote common.Vote) error {
 	pk, exists := e.validatorsToPKs[string(vote.Signature.Signer)]
 	if !exists {
 		e.Logger.Debug("Received a finalization from an unknown node", zap.Stringer("NodeID", from))
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2413,17 +2413,12 @@ func (e *Epoch) verifyBlockMessageVote(from common.NodeID, md common.BlockHeader
 		return fmt.Errorf("received a finalization from an unknown node %s", from)
 	}
 
+	// Ensure the block was voted on by its block producer:
+	// 1) Verify block header corresponds to the block header voted on
 	if !md.Equals(&vote.Vote.BlockHeader) {
 		return errors.New("vote block header does not match block header")
 	}
 
-	// Ensure the block was voted on by its block producer:
-	// 1) Verify block digest corresponds to the digest voted on
-	if !bytes.Equal(vote.Vote.Digest[:], md.Digest[:]) {
-		e.Logger.Debug("ToBeSignedVote digest mismatches block digest", zap.Stringer("voteDigest", vote.Vote.Digest),
-			zap.Stringer("blockDigest", md.Digest))
-		return errors.New("vote digest mismatches block digest")
-	}
 	// 2) Verify the vote is properly signed
 	if err := vote.Vote.Verify(vote.Signature.Value, e.Verifier, pk); err != nil {
 		e.Logger.Debug("ToBeSignedVote verification failed", zap.Stringer("NodeID", vote.Signature.Signer), zap.Error(err))
--- a/simplex/epoch_test.go
+++ b/simplex/epoch_test.go
@@ -2307,7 +2307,7 @@ func TestEpochBlockVoteHeaderMismatch(t *testing.T) {
 	require.NoError(t, err)
 
 	mismatched := *vote
-	mismatched.Vote.BlockHeader.Round++
+	mismatched.Vote.Round++
 	require.NoError(t, e.HandleMessage(&Message{
 		BlockMessage: &BlockMessage{Vote: mismatched, Block: block},
 	}, nodes[2]))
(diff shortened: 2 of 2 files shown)


simplex/epoch.go
```diff
@@ -1933,7 +1933,7 @@ func (e *Epoch) handleBlockMessage(message *common.BlockMessage, from common.Nod
 	}
 
 	// Check if we have verified this message in the past:
-	if err := e.VerifyBlockMessageVote(from, md, vote); err != nil {
+	if err := e.verifyBlockMessageVote(from, md, vote); err != nil {
```
samliok, line 1936: im not sure why this function was exported before

simplex/epoch.go
```diff
@@ -2405,26 +2405,19 @@ func (e *Epoch) createNotarizedBlockVerificationTask(block common.Block, notariz
...
+// verifyBlockMessageVote verifies that the vote corresponds to the block proposed, and that the vote is properly signed.
+func (e *Epoch) verifyBlockMessageVote(from common.NodeID, md common.BlockHeader, vote common.Vote) error {
 	pk, exists := e.validatorsToPKs[string(vote.Signature.Signer)]
 	if !exists {
 		e.Logger.Debug("Received a finalization from an unknown node", zap.Stringer("NodeID", from))
 		return fmt.Errorf("received a finalization from an unknown node %s", from)
 	}
 
-	// Ensure the block was voted on by its block producer:
+	if !md.Equals(&vote.Vote.BlockHeader) {
```
yacovm, line 2418: can we add a test?

samliok: added

yacovm (approved):

PR #590, opened 2026-09-10
Re-request rounds/sequences if buffer is full

### LINKED ISSUE #542: Replication state deleted before verification scheduling; queue-full error causes permanent stall or halt
## Details
Replicated blocks are removed from the ReplicationState before their verification is scheduled, and a scheduling failure is treated as either an ignorable or a fatal error - never as a retryable one. Two concrete consequences:

1) Permanent replication stall. When a replication response delivers the finalized block for nextSeqToCommit = S, storeSequence stores it and immediately cancels the re-request timeout for S (replication_state.go:123-136). processReplicationState later pops it and calls DeleteSeq(S) before processFinalizedBlock (epoch.go:3546-3549). If the shared verification queue is full at that moment, ScheduleTaskWithDependencies returns ErrTooManyPendingVerifications (common/block_scheduler.go:151-155) and processFinalizedBlock returns it (epoch.go:2062). The error propagates out of HandleMessage, but S is now gone from r.seqs, has no timeout task, and the requestor never re-requests sequences below highestRequested (requestor.go:159-173: start = max(current, highestRequested); during batched catch-up highestRequested > S). Since S is the next sequence to commit, the node can never commit S or any later sequence: peers keep answering requests for S+1..S+9, which sit unused in r.seqs while GetFinalizedBlockForSequence(S) stays empty forever. Recovery requires a process restart. The analogous pattern exists for round replication (DeleteRound before processNotarizedBlock, epoch.go:3556-3563).

2) Permanent node halt. When the same queue-full scheduling failure occurs inside a running verification task's processReplicationState call (createFinalizedBlockVerificationTask epoch.go:2253-2258, createNotarizedBlockVerificationTask epoch.go:2325-2330), the error is written to e.haltedError. haltedError is never cleared and HandleMessage returns it unconditionally thereafter (epoch.go:159-161), so the node stops processing all consensus messages until restarted (and the error is surfaced to the embedder, which for avalanchego-style integrations is treated as fatal).

The queue-full precondition is attacker-controllable: a malicious validator can fill the 500-slot verification queue with equivocating proposals during its leadership round (reported separately as leader-equivocation-floods-verification-queue-dropping-genuine-proposals), and nodes that fell behind during that starvation window are exactly the ones that then replicate finalized sequences - the attacker times its refill while victims fetch, so the pop of seq S (or the in-task processReplicationState) observes a full queue. What should be transient backpressure is thereby converted into an unrecoverable per-node liveness failure.

## Evidence
1. [simplex/epoch.go:3546–3550](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L3546-L3550)
   processReplicationState pops the finalized block for nextSeqToCommit and calls DeleteSeq(nextSeqToCommit) BEFORE processFinalizedBlock. If processFinalizedBlock fails (including the transient ErrTooManyPendingVerifications from a full verification queue), the block and finalization are already erased from replication state and are never restored.
2. [simplex/epoch.go:2060–2062](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L2060-L2062)
   processFinalizedBlock returns the error from ScheduleTaskWithDependencies directly. When the shared verification queue holds >= 500 pending tasks (attacker-fillable via leader equivocation flood), this returns ErrTooManyPendingVerifications, so the already-deleted sequence is never scheduled for verification and never committed.
3. [common/block\_scheduler.go:151–155](https://github.com/ava-labs/Simplex/blob/main/common/block_scheduler.go#L151-L155)
   Origin of the failure: when len(dependencies)+scheduler.Size() >= maxDeps (DefaultProcessingBlocks = 500, epoch.go:31), ScheduleTaskWithDependencies returns ErrTooManyPendingVerifications without enqueueing the task. The caller has already destroyed the replication-state entry, so this drop-without-refill error path loses the only copy of the data needed to retry.
4. [simplex/replication\_state.go:123–136](https://github.com/ava-labs/Simplex/blob/main/simplex/replication_state.go#L123-L136)
   storeSequence removes the re-request timeout task for the sequence (finalizationRequestor.removeTask and digestTimeouts.RemoveTask) as soon as the response is received. After the subsequent DeleteSeq + failed scheduling, no timeout remains that would ever re-request this sequence.
5. [simplex/requestor.go:159–173](https://github.com/ava-labs/Simplex/blob/main/simplex/requestor.go#L159-L173)
   sendMoreReplicationRequests computes start = max(currentSeqOrRound, highestRequested). During batched catch-up highestRequested is typically above nextSeqToCommit, so a dropped sequence S < highestRequested is permanently excluded from all future requests; observedSignedQuorum (lines 140-155) additionally short-circuits when highestRequested >= observed. Replication can then never fetch S again, and since S = nextSeqToCommit, no later sequence can be committed either: the node is stuck until restart.
6. [simplex/epoch.go:2253–2259](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L2253-L2259)
   Escalation path: createFinalizedBlockVerificationTask calls processReplicationState from the scheduler goroutine after indexing a block. If that call tries to schedule the next finalized sequence while the queue is full, the returned ErrTooManyPendingVerifications is assigned to e.haltedError - a transient backpressure condition is converted into a permanent halt.
7. [simplex/epoch.go:2325–2330](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L2325-L2330)
   Same escalation in createNotarizedBlockVerificationTask: processReplicationState errors (including queue-full scheduling failures of processFinalizedBlock/processNotarizedBlock) are stored in e.haltedError.
8. [simplex/epoch.go:159–161](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L159-L161)
   Once haltedError is set it is never cleared; every subsequent HandleMessage invocation returns the error immediately, permanently disabling all consensus and replication message processing for the node until process restart.
9. [simplex/epoch.go:3556–3563](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L3556-L3563)
   The same consume-before-process pattern exists for round replication: DeleteRound(lowestRound) is called before processNotarizedBlock, whose ScheduleTaskWithDependencies error (epoch.go:2120) likewise discards the round data; the roundRequestor timeout was already removed on receipt (requestor.receivedSignedQuorum), so the round is not re-requested.

## Impact
The affected node permanently stops committing blocks (stuck at nextSeqToCommit with no re-request path) or stops processing all messages entirely via haltedError; both persist until an operator restarts the process. Since the triggering flood is broadcast, multiple honest nodes can be wedged simultaneously, threatening network liveness if enough validators are affected. No confidentiality impact; no safety violation (no incorrect finalization), so integrity NONE.

## Reproduction steps
1. Attacker is a validator. It first fills the victim's 500-slot verification queue via an equivocation flood during its leadership round; victims that fell behind then fetch finalized sequences via replication. The attacker sustains/refreshes the full queue so that when the victim pops the next sequence to commit (or a running replication task calls processReplicationState), scheduling fails. This requires the victim to be in replication catch-up with a full queue at the right moment - a runtime-state condition the attacker fosters but does not fully control, hence attack_requirements PRESENT.

## Recommended fix
1. processReplicationState removes the block/finalization (DeleteSeq) and round data (DeleteRound) from the replication state before verification scheduling has succeeded, so a scheduling failure discards data that has no remaining re-request mechanism. Fix criteria: Replicated data must remain retrievable (in replication state or via an active re-request timeout) until its verification task has been successfully admitted to the scheduler. Verify that when ScheduleTaskWithDependencies returns ErrTooManyPendingVerifications for nextSeqToCommit, the sequence is retried or re-requested and the node eventually commits it once the queue drains.
2. ErrTooManyPendingVerifications, a transient backpressure signal, propagates from processReplicationState into e.haltedError inside verification tasks, permanently halting message processing for a recoverable condition. Fix criteria: Transient scheduling/backpressure errors must be distinguished from fatal errors (storage/WAL corruption) and must never set haltedError or propagate as fatal out of HandleMessage. Verify that a full verification queue during replication leaves the node degraded but able to resume normal operation once capacity frees, without restart.
3. The requestor's monotonic highestRequested cursor (start = max(current, highestRequested)) assumes every requested sequence is eventually processed; it cannot re-request a sequence that was received but subsequently dropped, leaving nextSeqToCommit permanently unfetchable. Fix criteria: The replication requestor must be able to re-request any uncommitted sequence at or above nextSeqToCommit regardless of highestRequested, e.g. driven by lack of commit progress. Verify that deleting a stored sequence below highestRequested results in it being requested again and committed.

---
**Severity:** HIGH
**Status:** Open
**Category:** Improper handling of exceptional conditions
**CWE:** [CWE-755](https://cwe.mitre.org/data/definitions/755.html)
**Branch:** main
**Date created:** 2026-08-21

---

samliok:
Shouldn't happen for finalized blocks which is why I used a warning log, but more probable for notarized blocks. 

Closes #542

### DIFF
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2188,18 +2188,16 @@ func (e *Epoch) processNotarizedBlock(block common.Block, notarization *common.N
 	task := e.createNotarizedBlockVerificationTask(e.oneTimeVerifier.Wrap(block), *notarization)
 	blockDependency, missingRounds := e.blockDependencies(md)
 
+	e.replicationState.CreateDependencyTasks(blockDependency, md.Seq-1, missingRounds)
+
 	err := e.blockVerificationScheduler.ScheduleTaskWithDependencies(task, md.Seq, blockDependency, missingRounds)
 	if errors.Is(err, common.ErrTooManyPendingVerifications) {
 		e.Logger.Debug("Verification queue is full, re-requesting notarized block", zap.Uint64("round", md.Round))
 		e.replicationState.ResendRoundRequest(md.Round, notarization.QC.Signers())
 		return nil
 	}
-	if err != nil {
-		return err
-	}
 
-	e.replicationState.CreateDependencyTasks(blockDependency, md.Seq-1, missingRounds)
-	return nil
+	return err
 }
 
 func (e *Epoch) createBlockVerificationTask(block common.Block, from common.NodeID, vote common.Vote) func() common.Digest {
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2125,10 +2125,9 @@ func (e *Epoch) processFinalizedBlock(block common.Block, finalization *common.F
 	task := e.createFinalizedBlockVerificationTask(e.oneTimeVerifier.Wrap(block), finalization)
 	err := e.blockVerificationScheduler.ScheduleTaskWithDependencies(task, block.BlockHeader().Seq, blockDependency, []uint64{})
 	if errors.Is(err, common.ErrTooManyPendingVerifications) {
-		// The sequence was already removed from the replication state, so re-request it.
-		// This shouldn't happen since processing a finalized block means its the next sequence to commit
-		// so we should be able to schedule it.
-		e.Logger.Warn("Verification queue is full, re-requesting finalized block", zap.Uint64("seq", block.BlockHeader().Seq))
+		// A full scheduler shouldn't fatal
+		e.Logger.Debug("Dropping finalized block, too many pending verifications",
+			zap.Uint64("seq", block.BlockHeader().Seq), zap.Error(err))
 		e.replicationState.ResendFinalizationRequest(block.BlockHeader().Seq, finalization.QC.Signers())
 		return nil
 	}
--- a/simplex/epoch.go
+++ b/simplex/epoch.go
@@ -2123,7 +2123,16 @@ func (e *Epoch) processFinalizedBlock(block common.Block, finalization *common.F
 
 	// Create a task that will verify the block in the future, after its predecessors have also been verified.
 	task := e.createFinalizedBlockVerificationTask(e.oneTimeVerifier.Wrap(block), finalization)
-	return e.blockVerificationScheduler.ScheduleTaskWithDependencies(task, block.BlockHeader().Seq, blockDependency, []uint64{})
+	err := e.blockVerificationScheduler.ScheduleTaskWithDependencies(task, block.BlockHeader().Seq, blockDependency, []uint64{})
+	if errors.Is(err, common.ErrTooManyPendingVerifications) {
+		// The sequence was already removed from the replication state, so re-request it.
+		// This shouldn't happen since processing a finalized block means its the next sequence to commit
+		// so we should be able to schedule it.
+		e.Logger.Warn("Verification queue is full, re-requesting finalized block", zap.Uint64("seq", block.BlockHeader().Seq))
+		e.replicationState.ResendFinalizationRequest(block.BlockHeader().Seq, finalization.QC.Signers())
+		return nil
+	}
+	return err
 }
 
 // processNotarizedBlock processes a block that has a notarization.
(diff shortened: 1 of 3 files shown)


yacovm (approved):

yacovm:
```
2026-09-11T15:12:16.7061298Z [09-11|15:09:16.188] WARN testutil/logger.go:87 Too many blocks being verified to ingest another one {"pendingBlocks": 500}
```

This fails the CI job

PR #595, opened 2026-09-15
Simplify Epoch Collector to Store One Vote Per Validator

### LINKED ISSUE #563: epochDigestCounter grows unbounded under attacker-chosen sequence keys that are never pruned
## Details
The epochDigestCounter records sealing-block attestations in sealingBlockResponses, a map keyed by the sequence number read from the sender's UNVERIFIED block header (bh.Seq). A malicious validator can send a stream of ReplicationResponses, each containing a fabricated quorum round (block with a non-nil SealingBlockInfo, an unknown epoch number, and a digest-consistent fake finalization so it passes VerifyQCConsistentWithBlock) with a distinct, arbitrarily large Seq. Each message allocates a new outer map entry plus inner map (sender -> digest). Cleanup (removeOldEpochs) deletes only keys strictly below the epoch of the last indexed block, so keys chosen near 2^64 are never removed for the process lifetime — the memory is a permanent leak rather than transient buffering, growing linearly and without bound in the number of attacker messages. The counter is unreachable for non-validator senders (validators.Contains(from) gate), so validator privileges are required.

## Evidence
1. [nonvalidator/epochs.go:182–190](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/epochs.go#L182-L190)
   collectedSealingBlockInfo creates a new sealingBlockResponses[newEpoch] map entry for every distinct bh.Seq taken from the sender's unverified block header; a validator can send quorum rounds with arbitrary distinct Seq values, each allocating a persistent entry.
2. [nonvalidator/epochs.go:208–213](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/epochs.go#L208-L213)
   removeOldEpochs only deletes keys strictly below minEpochToKeep (the epoch of the last indexed block). Keys chosen near MaxUint64 are never pruned for the lifetime of the process.
3. [nonvalidator/non\_validator.go:532–540](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/non_validator.go#L532-L540)
   Reachable from handleQrFromUnknownEpoch for any quorum round whose block carries a SealingBlockInfo and an unknown epoch; the only gate is that the sender is in the current validator set.

## Impact
Persistent, unprunable memory growth on the non-validator, eventually degrading or crashing the process; each entry is small so exhaustion requires a large number of messages, and a restart clears the state. No confidentiality or integrity impact.

## Reproduction steps
1. Attacker must be a current validator (membership check against Comm.Validators()). It streams replication responses with fabricated sealing blocks carrying distinct huge sequence numbers; every message permanently allocates map state on the non-validator. Impact accumulates with sustained traffic since per-entry size is small.

## Recommended fix
sealingBlockResponses is keyed by an unvalidated, attacker-chosen sequence number, has no cap on the number of tracked epochs, and its pruning predicate (strictly below the current epoch) never removes keys above the chain's progress. Fix criteria: Bound the number of tracked prospective epochs (e.g., only track sequences within a window above the current tip, or cap entries per sender) and prune keys that cannot correspond to real epochs. Verify a validator streaming distinct huge sequence numbers cannot grow the map beyond a fixed bound.

---
**Severity:** LOW
**Status:** Open
**Category:** Allocation without limits
**Branch:** main
**Date created:** 2026-08-21

---

samliok:
Closes #563 

Updates the epoch collector to map nodeIDs -> responses rather than key by epoch. We only care about the highest sealing block so no reason to store responses with lower epochs.


nonvalidator/epochs.go
```diff
@@ -179,36 +183,36 @@ func (e *epochDigestCounter) collectedSealingBlockInfo(sealingBlockInfo *common.
...
+	for _, other := range e.sealingBlockResponses {
+		if other != response {
+			continue
+		}
 		count++
-		counts[digest] = count
-
-		if counts[digest] >= uint64(threshold) {
-			e.logger.Info("We received enough messages to validate a higher epoch", zap.Stringer("EpochInfo", sealingBlockInfo), zap.Int("Threshold", threshold), zap.Uint64("Responses", counts[digest]))
+		if count >= threshold {
```
yacovm: I am not sure I agree to this kind of counting. 

This counts responses from *any* epoch for the `F` of the latest validator set.

So for epoch `i` with `n=10` we can have `2` votes that are valid for an epoch `j>i` with `n=7`. 

Shouldn't we verify that all responses we found in `e.sealingBlockResponses` are within `e.validators`?

yacovm (approved):

PR #598, opened 2026-09-16
Rebroadcast finalize votes before proposing an empty block

yacovm:
When the tip is notarized but not finalized, two timers race: NotarizationTime rebroadcasts the finalize vote for the tip so peers can send back the missing finalization, and the empty-block builder gives up waiting for transactions and proposes an empty block on top of the tip.

Align them so the rebroadcast always comes first. FinalizeRebroadcastTimeout is now three network delays, so the rebroadcast fires within six network delays of entering the round, and the empty-block timeout is four times MaxProposalWait, eight network delays, which leaves peers a round trip to answer.

Add an instance-level test that partitions finalization traffic between two validators and asserts the leader rebroadcasts its finalize vote before it proposes the empty block.


simplex/epoch.go
```diff
@@ -248,7 +248,7 @@ func (e *Epoch) init() error {
 	}
 	e.blockBuilder = &EmptyBlockBuilder{
 		ShouldBuildEmptyBlock: e.haveUnFinalizedButNotarizedSuffix,
-		Timeout:               e.MaxProposalWait,
+		Timeout:               e.MaxProposalWait * 4,
```
samliok, line 251: x3 and x4 seem like they are magic numbers that we will forget in a couple weeks. Is there a better way to set these toggles so they dont become footguns?

instance_test.go
```diff
@@ -556,3 +556,87 @@ func TestValidatorSetsMetadataFromSnowman(t *testing.T) {
...
+
+// TestFinalizeVoteRebroadcastPrecedesEmptyBlockProposal asserts that a leader whose tip is notarized
+// but not finalized rebroadcasts its finalize vote for the tip (using NotarizationTime) before it proposes an
+// empty block on top of it (empty-block timeout), so peers get a chance to send the finalization first.
+//
+// Node A leads round 2, node B leads round 3. Finalize votes and finalizations from A to B are dropped,
+// so A finalizes block 2 while B only notarizes it. B then enters round 3 as leader with no pending
+// transactions. Among the messages B sends after entering round 3, the finalize vote for round 2 must
+// appear before the (empty) block proposal for round 3.
+func TestFinalizeVoteRebroadcastPrecedesEmptyBlockProposal(t *testing.T) {
```
samliok, line 568: i feel like this test is kind of long and hard to parse, i tried to make a simplified version of it [here](https://github.com/ava-labs/Simplex/pull/602). lemme know what you think

instance_helpers_test.go
```diff
@@ -639,6 +650,22 @@ func (n *network) nodesSnapshot() []node {
...
 
+// setInterceptor installs the delivery interceptor; pass nil to remove it.
+func (n *network) setInterceptor(intercept func(from, to common.NodeID, msg *common.Message) bool) {
+	n.lock.Lock()
+	defer n.lock.Unlock()
+	n.intercept = intercept
+}
+
+// deliverable reports whether a message from one node to another should be delivered.
+func (n *network) deliverable(from, to common.NodeID, msg *common.Message) bool {
```
samliok, line 661: we have the `messageFilter` term and types in testutil, maybe we can re-use that?

PR #603, opened 2026-09-18
Remove NotarizationTime, rely on empty blocks instead

yacovm:
The node may periodically re-broadcast a finalize vote on a round that it has notarized but not finalized. This is to re-broadcast finalize votes lost in the network.
    
If the node has a notarization for the round but has not stored the finalize vote, it will create the finalize vote on its own before sending it.
    
However, a node may also replicate a notarization from another node.
    
A node should not send a finalize vote if it has timed out on a round, otherwise there is a safety violation.
    
This commit removes the re-broadcast of finalize votes altogether, because it is not redundant thanks to the mechanism which makes nodes build empty blocks once a tip of the chain is notarized but not finalized.
    
We just change the mechanism to also take into account blocks that we haven't timed out on in the middle of the chain are notarized but not finalized, and not only in the tip of the chain.


simplex/epoch_test.go
```diff
@@ -3414,3 +3421,66 @@ func TestEpochLeaderRecordsTimeoutDespiteStaleBlacklistOnEpochChange(t *testing.
 	require.True(t, expected.Equals(&actual),
 		"the timeout of nodes[2] was not recorded: expected %s, got %s", expected.String(), actual.String())
 }
+
+// TestRebroadcastDoesNotFinalizeVoteOnTimedOutRound asserts that a node never casts a finalize vote for a
+// round it has cast an empty vote for. Doing both lets an empty notarization and a finalization form for
+// the same round, since the two quorums may then overlap only in nodes that voted both ways.
+func TestRebroadcastDoesNotFinalizeVoteOnTimedOutRound(t *testing.T) {
```
samliok, line 3428: nice test 👍

simplex/epoch.go
```diff
@@ -329,15 +304,20 @@ func (e *Epoch) Start() error {
 	return nil
 }
 
-func (e *Epoch) haveUnFinalizedButNotarizedSuffix(ctx context.Context) bool {
+func (e *Epoch) haveUnFinalizedButNotarizedRound(ctx context.Context) bool {
```
samliok: this is only called from the block builder, and it is no longer just checking if we have a not finalized & notarized round. Maybe we can rename it to 

`shouldBuildEmptyBlock`

simplex/epoch.go
```diff
@@ -261,28 +258,6 @@ func (e *Epoch) init() error {
...
-
-func (e *Epoch) getRound() uint64 {
-	e.lock.Lock()
-	defer e.lock.Unlock()
-
-	return e.round
-}
-
 func (e *Epoch) maybeAssignDefaultConfig() error {
 	if e.FinalizeRebroadcastTimeout == 0 {
```
samliok: we can remove this config value

simplex/epoch.go
```diff
@@ -3113,6 +3102,34 @@ func (e *Epoch) increaseRound() {
...
+		return nil
+	}
+
+	if e.haveWeAlreadyTimedOutOnThisRound(r) {
+		e.Logger.Debug("Not finalize voting for a replicated notarization of a round we timed out on", zap.Uint64("round", r))
+		return nil
+	}
+
+	md := round.notarization.Vote.BlockHeader
+	finalizeVote, finalizeVoteMsg, err := e.constructFinalizeVoteMessage(md)
```
samliok: i think it's weird that we have 3 finalize places where we construct the finalized votes, i feel like we should be able to create a single `finalizeVote` method that handles constructing, broadcasting and checking if we have timed out.

samliok: thoughts on something like this? https://github.com/ava-labs/Simplex/compare/finalizevote?expand=1

yacovm: 1. We don't want to call `e.handleFinalizeVoteMessage(&vote, e.ID)` each time we cast a finalize vote. We will verify it needlessly. BLS signatures are much more expensive to verify than to create, because of the pairing check. 
2. This is a fragile piece of the code so I think we should do refactoring in a separate PR and not in this one.

yacovm: I did add a defensive sanity check:

```
func (e *Epoch) maybeConstructFinalizeVoteMessage(md common.BlockHeader) (common.FinalizeVote, *common.Message, error) {
	if e.haveWeAlreadyTimedOutOnThisRound(md.Round) {
		e.Logger.Error("Will not cast a finalize vote for a round we timed out on)", zap.Uint64("round", md.Round))
		return common.FinalizeVote{}, nil, errAlreadyTimedOutOnRound
	}
```

for now, that should prevent us from having such bugs.

samliok (approved):

PR #604, opened 2026-09-18
Non-validator: Store blocks from many leaders

### LINKED ISSUE #565: Byzantine validator can wedge per-sequence block slot with self-chosen leader round
## Details
handleBlock authenticates a proposed block only by checking that the sender equals LeaderForRound(epoch.nodes, bh.Round), where bh.Round is read from the unverified block itself. Any validator can therefore craft a junk block for any sequence within the MaxSequenceWindow, choose a Round value for which it is the leader (round mod n selects it), and have the block stored in incompleteSequences[seq]. The slot is first-writer-wins: once occupied, the genuine leader's real block for the same sequence is discarded at line 208 (incomplete.block != nil), and when the genuine finalization arrives the digest mismatch path (lines 385-395) neither evicts the junk block nor completes the pairing. The sequence can then only be committed via the replication fallback (ReceivedFutureFinalization -> request -> response -> processQuorumRound), adding request/response round-trips per poisoned sequence.

A single Byzantine validator can do this continuously for every upcoming sequence in the window, permanently degrading the non-validator's fast path so that it trails the network by replication latency for every block. Impact is limited to added latency/traffic (self-healing via replication; no halt, no incorrect data), hence LOW.

## Evidence
1. [nonvalidator/non\_validator.go:185–209](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/non_validator.go#L185-L209)
   handleBlock accepts an unverified block if the sender is the leader of the round THE BLOCK ITSELF declares (bh.Round is attacker-chosen and unconstrained by the node's round view, so any validator can pick a round making itself leader), stores it first-writer-wins in incompleteSequences[bh.Seq], and line 208 then rejects any later (real) block for that sequence while the entry exists.
2. [nonvalidator/non\_validator.go:378–395](https://github.com/ava-labs/Simplex/blob/main/nonvalidator/non_validator.go#L378-L395)
   When the genuine QC-verified finalization arrives, it is stored (line 378) but the digest mismatch with the junk block only logs and returns; the junk block is not evicted, so the direct block+finalization fast path for that sequence can never complete and the node must fall back to replication round-trips.
3. [simplex/epoch.go:3684–3687](https://github.com/ava-labs/Simplex/blob/main/simplex/epoch.go#L3684-L3687)
   LeaderForRound selects nodes[r % n] over the sorted epoch validator set; because the non-validator evaluates it on the attacker-chosen bh.Round, any validator can choose a round congruent to its own index and pass the leader check. The validator-side handler, by contrast, constrains the round to its current view and verifies a signed vote — protections the non-validator path lacks.

## Impact
Only the direct commit fast path is disabled per sequence; the replication fallback recovers each sequence with extra round-trips, so the effect is sustained latency degradation and extra traffic, not loss of the service.

## Reproduction steps
1. Attacker must be a validator of a registered epoch (the leader check compares the sender to the leader derived from the attacker-chosen round). It sends one junk block message per target sequence within the window before the real proposal arrives; no other conditions are needed.

## Recommended fix
The leader check trusts the round number declared by the unverified block, letting any validator impersonate 'the leader', and the per-sequence block slot is first-writer-wins with no eviction when the verified finalization's digest contradicts the stored block. Fix criteria: When a verified finalization's digest mismatches the stored block, the stored block must be replaced/evicted so the genuine block can still pair directly; and/or blocks should only be accepted for rounds consistent with the node's view. Verify that after a junk block is stored, delivery of the real block and finalization still commits via the direct path.

---
**Severity:** LOW
**Status:** Open
**Category:** Insufficient verification of data authenticity
**CWE:** [CWE-345](https://cwe.mitre.org/data/definitions/345.html)
**Branch:** main
**Date created:** 2026-08-21

---

samliok:
Adversaries can send us many blocks for different sequences, and we can't tell which block is legit until we receive the finalization.

Closes #565


nonvalidator/non_validator.go
```diff
@@ -17,23 +17,24 @@ import (
 )
 
 type finalizedSeq struct {
-	block        common.Block
+	// blocks holds one block per leader that proposed for this sequence.
+	// Rounds that share a sequence can have different leaders, so several blocks may be stored.
+	blocks       map[string]common.Block
```
yacovm: we know that we hold blocks. Can we make the field name more descriptive? 

Maybe `proposerIDToBlock`?

nonvalidator/non_validator.go
```diff
@@ -402,35 +389,27 @@ func (n *NonValidator) handleFinalization(finalization *common.Finalization, fro
...
 			// because stored.Epoch < bh.Epoch
-			n.Logger.Debug("Dropping stored Telock sequence", zap.Stringer("Sequence", incomplete))
-			incomplete.block = nil
-			n.sequenceReplicator.ReceivedFutureFinalization(finalization, n.nextSeqToCommit())
+			n.Logger.Debug("Dropping stored Telock finalization", zap.Stringer("Sequence", incomplete))
 		}
 	}
 
 	incomplete.finalization = finalization
+	// Let the replicator know this sequence is finalized in case no stored block matches it.
```
yacovm: why is that "in case" ? shouldn't we do it even if a stored block matches it?

nonvalidator/non_validator.go
```diff
@@ -402,35 +389,27 @@ func (n *NonValidator) handleFinalization(finalization *common.Finalization, fro
...
-			zap.Stringer("Finalization Digest", bh.Digest),
-			zap.Stringer("Block digest", digest),
-			zap.Stringer("From", from),
-		)
-
-		incomplete.block = nil
-		n.sequenceReplicator.ReceivedFutureFinalization(finalization, n.nextSeqToCommit())
-		return nil
+	for _, block := range blocks {
+		if block.BlockHeader().Digest == bh.Digest {
```
yacovm, line 406: if it's not equal shouldn't we delete the block from memory?

nonvalidator/non_validator_test.go
```diff
@@ -175,6 +175,66 @@ func TestHandleMessages(t *testing.T) {
 			},
 			expectedHeight: 3,
 		},
+		{
```
yacovm, line 178: we have `3` items, so the number of permutations is 6, not 3. 

For example, we're missing finalization then the blocks.

nonvalidator/non_validator.go
```diff
@@ -223,8 +221,6 @@ func (n *NonValidator) handleBlock(block common.Block, from common.NodeID) error
 		return nil
 	}
 
-	incomplete.block = block
-
 	n.maybeValidateNextEpoch(block)
 	return n.scheduleNewFinalizedBlockTask(block, incomplete.finalization)
```
yacovm, line 225: can't an adversary schedule as many blocks as it wants now? Shouldn't we mark that "we have scheduled the block" ?

nonvalidator/non_validator.go
yacovm, line 376: Wondering how does this play with receiving finalizations for:

1. A Telock that is built on a sealing block
2. The first block of the next epoch

?

nonvalidator/non_validator.go
samliok: I believe this relates to this issue and fix https://github.com/ava-labs/Simplex/pull/596

nonvalidator/non_validator.go
samliok: wouldn't the finalizations have different epochs?

nonvalidator/non_validator.go
samliok: maybe the finalization we received is for a telock, but the blocks we have in memory are for the next epoch.

nonvalidator/non_validator_test.go
samliok: added

